Issue #11, opened 2022-08-16
Multiple credentials of the same type


OIDF-automation:
### Original Reporter: dwc8

Some issuers may issue multiple credentials of the same type. Examples include: two almost identical passports, once used by the holder to visit Israel and the other to visit Iran; or a current credit card that expires next month and a second one that starts the following month.

The current issuing API does not cater for this as the wallet can only specify the credential type and the issuer can only return a single credential. One solution would be to allow the issuer to return an array of credentials of the same type and format.

Sakurann:
PR merged!

PR #65, opened 2023-09-01
add identifiers for issuing credential of the same type, different content

Sakurann:
addresses Issue #30 and #11.
Migrated from Bitbucket: https://bitbucket.org/openid/connect/pull-requests/612

The problem addressed is: lack of ability to communicate issuance of credentials that are same format/type, but different content.

there were few iterations, but the latest ver of a PR:
- Token response: Issuer can communicate a specific credential for which the Access token is valid for using a new `identifiers` parameter (array)
- Credential request: wallet must use identifiers received in token response using a new `identifiers` parameter (array)


Sakurann:
@tlodderstedt made a comment in Bitbucket regarding token response:
> This list might be ok for scope-based requests. However, if authorization was requested using authorization details, the identifiers should be added to the authorization details objects.
>
> For scope-based requests: How is the wallet supposed to relate the identifiers to the (potentially) different credential types it had requested?

I didn't fully understand why.. needs discussion.

sloops77:
To clarify the PR - what is being proposed is that a human-friendly identifier will be required for each different combination of credential data that could be considered unique to a person (or their wallet).

For example, an Issuer issuing employee of the month badges would be required to continually use (& manage) new `identifiers` to differentiate to the wallet that `EmployeeOfTheMonth-May2023` is different from `EmployeeOfTheMonth-Dec2023`, where the `type: OpenBadgeCredential`. In other cases the year may be sufficient for differentiating, in others it may not be time but another variable.

Is anyone else nervous about having Issuers figure out what is unique about a credential? 

Going with a straight generated `id` and describing in the `/token` response which generated `id` corresponds to which type would have the benefit of not requiring this effort from issuers (or their software)

Sakurann:
@sloops77 this PR is not about the human-readable `id` per type, it is about multiple `id`s per type, for the use-cases when there are multiple credentials being issued are of the same type but different content. and this identifier does not have to be human-readable, can be a UUID IMO... wallet just needs it as a signal that despite requesting only one type, multiple credentials are being issued.

Sakurann:
@tlodderstedt added `authorization_details.identifiers` parameter for the token response.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -602,8 +607,11 @@ In addition to the response parameters defined in [@!RFC6749], the AS MAY return
 
 * `c_nonce`: OPTIONAL. JSON string containing a nonce to be used to create a proof of possession of key material when requesting a Credential (see (#credential_request)). When received, the Wallet MUST use this nonce value for its subsequent Credential Requests until the Credential Issuer provides a fresh nonce.
 * `c_nonce_expires_in`: OPTIONAL. JSON integer denoting the lifetime in seconds of the `c_nonce`.
+* `identifiers`: OPTIONAL. JSON array of JSON strings that each identify a credential that can be issued using Access Token returned in the same response. MUST NOT be used when `authorization_details` parameter was used to request issuance of a certain Credential type as defined in (#authorization-details).
```
tlodderstedt, line 610: A wallet could use `scope` and `authorization_details` in the same request. So the language should be along the lines of "this array contains the credentials that can be requested based on the scope values passed to the authorization request". This means `identifiers` and `authorization_details/identifiers` can turn up in the same token response.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -689,6 +727,7 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
 * `format`: REQUIRED. Format of the Credential to be issued. This Credential format identifier determines further parameters required to determine the type and (optionally) the content of the credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles).
 * `proof`: OPTIONAL. JSON object containing proof of possession of the key material the issued Credential shall be bound to.  The `proof` object MUST contain a following claim:
     * `proof_type`: REQUIRED. JSON string denoting the key proof type. The value of this claim determines other claims in the key proof object and its respective processing rules. Key proof types defined in this specification can be found in (#proof_types).
+* `identifier`: OPTIONAL. JSON string that identifies a credential that is being requested to be issued. It MUST be present if Credential Issuer returned `identifiers` parameter the Token Response.
```
tlodderstedt, line 730: Not sure this parameter should be mutual exclusive with all additional parameters used to determine the type of credentials as this data was already passed in the authorization request.

tlodderstedt (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: I rephrase my feedback: I suggest to change the text for 'format' so it MUST NOT be specified if it can already be determined from the identifier.

Sakurann:
@pmhsfelix

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -461,7 +466,7 @@ Credential Issuers MUST ignore unknown scope values in a request.
 
 If the Credential Issuer metadata contains an `authorization_server` property, it is RECOMMENDED to use a `resource` parameter [@!RFC8707] whose value is the Credential Issuer's identifier value to allow the AS to differentiate Credential Issuers.  
 
-Below is a non-normative example of an Authorization Request using the scope `com.example.healthCardCredential` that would be sent by the User Agent to the Authorization Server in response to an HTTP 302 redirect response by the Wallet (with line wraps within values for display purposes only):
+Below is a non-normative example of an Authorization Request using the scope `UniversityDegree_JWT` that would be sent by the User Agent to the Authorization Server in response to an HTTP 302 redirect response by the Wallet (with line wraps within values for display purposes only):
```
jogu, line 469: I think we only have this one example using scopes now - it would be great to have examples for the other options too (so 3 in total I think? one with scope, one with authorization_details, one with identifiers)

tlodderstedt:
Should we require the AS to always return identifiers for all credentials? That might improve and simplify implementations.

Sakurann:
Kristina to make changes making identifier mandatory in the token response and credential request to solve another issue too - WG to review next week and make a decision.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -314,7 +314,9 @@ For security considerations, see (#credential-offer-security).
...
+* `credentials`: REQUIRED. A JSON array, where every entry is a JSON string or a JSON object describing a credential the Wallet MAY request. If the entry is a string, the string value MUST be one of the `scope` values included in a `credentials_supported` Credential Issuer metadata parameter as defined in (#credential-metadata-object). When processing, the Wallet MUST resolve this string value to the respective object. If the entry is an object, the object contains the data related to a certain credential type. Each object contains the following parameters: 
+  * `format`: REQUIRED. JSON string determining the format of the credential.
+  * Parameters characterizing the type of the credential to be requested: REQUIRED. These parameters are specific to the credential format profile, some of which are defined in (#format_profiles).
```
danielfett, line 319: The current definition does not include the `display` and other properties defined in (#credential-metadata-object) — is that intentional?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -444,7 +447,9 @@ Note: Applications MAY combine authorization details of type `openid_credential`
...
 
-When the Wallet does not know which scope value to use to request issuance of a certain credential, it can discover it using the `scope` Credential Issuer metadata parameter defined in (#credential-metadata-object). When the flow starts with a Credential Offer, the Wallet can use the information in the `credentials` Credential Offer parameter. When the entry of the `credentials` parameter is a string, it can be used as a scope value. When the entry of the `credentials` parameter is an object, it can be used to discover the desired scope value from the Credential Issuer metadata that can be obtained using `credential_issuer` Credential Offer parameter as defined in (#credential-issuer-wellknown).
+When the Wallet does not know which scope value to use to request issuance of a certain credential, it can discover it using the `scope` Credential Issuer metadata parameter defined in (#credential-metadata-object). When the flow starts with a Credential Offer, the Wallet can use the information in it as following:
```
danielfett, line 450: `scope` is not a credential issuer metadata parameter.
```suggestion
When the Wallet does not know which scope value to use to request issuance of a certain credential, it can discover the available `scope` values in the `credentials_supported` Credential Issuer metadata parameter defined in (#credential-metadata-object). When the flow starts with a Credential Offer, the Wallet can use the information in it as following:
```

openid-4-verifiable-credential-issuance-1_0.md
danielfett, line 610: The whole `identifiers` thing is not communicated well right now. 

What does an identifier identify? What is it used for? Why are there two types of identifiers (for creds requested with scopes and authorization details)? Are they conflict-free? Why do identifiers show up in the credential response? 

I can guess answers to all of these questions, but these things must be explicit in the text.

I suppose this should be integrated into an explanation of the various ways of communicating credential contents (offered or requested). (I.e., explaining the mental model behind the credential offer, the metadata, the authorization request, the token response, the credential request and credential response in a very explicit way.) Some of this is done in (#credential-authz-request), but it dives into details really fast and lacks the high-level explanation.

openid-4-verifiable-credential-issuance-1_0.md
pmhsfelix, line 319: I wonder if an identifier should also be allowed in the offer credentials object, as a way to precisely specify the credential to issue.
Use case:
- Training company has a web site where users can see the completed training courses.
- For each course (e.g. `AERO-123-FluidMechanics-II-2023`), user can require a credential, which requires some sort of payment.
- After payment transaction is completed, web site presents an offer as a QR code. This offer exactly identifies the credentials that the user can request (e.g. `AERO-123-FluidMechanics-II-2023`).
- User scans QR code with wallet, authenticates, consents on a screen that only shows `AERO-123-FluidMechanics-II-2023` credential information (no need for the user to select anything), and finally gets an issued credential.

The use of an identifier would allow an offer to be much more precise in the credential that is being offered.

The identifier could be carried in the authorization request's `authorization_details` object.

tlodderstedt: That proposal would require to send the identifiers with the authorization and/or token request, too. It looks complex on first sight but could further simplify the live of implementers as the wallet could _just_ use the identifiers to request the authz code or the access token. This requires us to redefine the authorization details type to allow to carry identifier only (very much like what I proposed for the credential endpoint). I'm not sure how it would work with scopes, I guess we would need to allow authorization request with a list of identifiers w/o scope parameter at all. 
Note: I think it makes sense for pre authz code and authorization code with issuer state, I don't see how it could be implemented with authz code w/o issuer state. 
PS: I would trade that feature for dropping scopes at all ;-)

Sakurann: Credential Offer is a mere offer as explained in the spec already and elaborated in PR #73. 

@danielfett @pmhsfelix `display`, `cryptographic_binding_methods_supported` and `cryptographic_suites_supported` should be discovered using credential issuer metadata instead of being passed in the credential offer.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -444,7 +447,9 @@ Note: Applications MAY combine authorization details of type `openid_credential`
...
+  * When the entry of the `credentials` Credential Offer parameter is a string, it can be directly used as a scope value.
+  * When the entry of the `credentials` Credential Offer parameter is an object, it can be used to discover the desired `scope` value from the Credential Issuer metadata that can be obtained using `credential_issuer` Credential Offer parameter as defined in (#credential-issuer-wellknown).
```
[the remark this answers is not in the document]
Sakurann: agreed that a key is needed when doing a look up in `credentials_supported` entry that matches `credentials` entry. also based on your comment above, I think an optional `identifier` in the offer might be better suited for this purpose, because the scope value will not always be unique to a credential being offered.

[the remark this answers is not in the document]
Sakurann: > an issuer may want to offer credentials that don't appear in the metadata, namely because they are very context specific.

How does the wallet know that it is being offered such a credential? I think this requires a separate issue.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* `identifiers`: OPTIONAL. JSON array of JSON strings that each identify a credential that can be issued using Access Token returned in this response. 
```
I am sorry, i cannot understand a proposed sentence. I think removing `MUST NOT...` sentence should be sufficient.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -602,8 +607,11 @@ In addition to the response parameters defined in [@!RFC6749], the AS MAY return
...
 * `c_nonce_expires_in`: OPTIONAL. JSON integer denoting the lifetime in seconds of the `c_nonce`.
+* `identifiers`: OPTIONAL. JSON array of JSON strings that each identify a credential that can be issued using Access Token returned in the same response. MUST NOT be used when `authorization_details` parameter was used to request issuance of a certain Credential type as defined in (#authorization-details). 
+* `authorization_details`: REQUIRED when `authorization_details` parameter was used to request issuance of a certain Credential type as defined in (#authorization-details). A JSON array of objects as defined in Section 7 of [@!RFC9396]. MUST NOT be used when `scope` parameter was used to request issuance of a certain Credential type as defined in (#credential-request-using-type-specific-scope). In addition to the parameters received from the Wallet, the AS MAY return the following parameter:
```
Sakurann, line 611:
```suggestion
* `authorization_details`: REQUIRED when `authorization_details` parameter was used to request issuance of a certain Credential type as defined in (#authorization-details). A JSON array of objects as defined in Section 7 of [@!RFC9396]. MAY be used when `scope` parameter was used to request issuance of a certain Credential type as defined in (#credential-request-using-type-specific-scope). In addition to the parameters received from the Wallet, the AS MAY return the following parameter:
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: > When using identifier on a credential request, none of these parameters should be present in the request.

+1

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
Sakurann: should it be `c_identifier` in all occurences and not only in token response?

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: I agree it is beneficial to add identifiers in the credential offer. it will also allow the wallet to look up a specific credential's metadata in the issuer metadata. 

@tlodderstedt when scopes are used, I would expect the following:
- wallet uses an identifier to look up metadata about an offered credential. not scope because there is no guarantee scope value is unique to a credential. (there is a related comment from Pedro below)
- wallet uses the scope value discovered in the previous step in authz/token requests.
- the rest is as already defined.

when authorization details is used, identifier should probably be included inside the `authorization_details` parameter passed in authz/token requests (token req would be optional)?

pmhsfelix: Does that mean that an offer should always be associated to exactly one entry in the metadata's `credentials_supported` array?
Namely, can't we use offers for credentials _kinds_ that are not statically described in the server's metadata?

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
pmhsfelix: Following what was done for nonces, the `c_` prefix could only be used on the token response. However using the same identifier everywhere will make implementation and debugging much easier, IMO.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: > can't we use offers for credentials kinds that are not statically described in the server's metadata?

in my mental model issuer was offering concrete credentials using the offer. what does offering `credential kinds` mean? how does a wallet know what to request?

pmhsfelix: I used the term "kind" to avoid the term "type", but I'm also thinking in the offer describing a concrete credential.
In my view, the wallet needs:
- display info to show the offered credential to the end user. When using W3C data model, this can be a `credentialSubject`.
- cryptographic parameters needed for the credential request proof (e.g. `cryptographic_binding_methods_supported`).
- The *identifier*, to be sent on the authorization request or token request, when using pre-authorized code.
This seems enough for the wallet to:
1) Present a UI to the user.
2) Know how to perform the complete flow.

The offer should be self-sufficient for the wallet to perform a full credential request flow, without needing to use issuer's metadata `credentials_supported`.
The authorization server ID and credential endpoint URL could still come from the issuer metadata.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* `identifier`: OPTIONAL. JSON string that identifies a credential that is being requested to be issued. It MUST be present if Credential Issuer returned `identifiers` parameter the Token Response. When this parameter is used, `format` parameter and any other credential format specific set of parameters such as those defined in (#format_profiles) MUST NOT be present.
```

`format` should not be present, too, right?

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: I am less sure about the need to pass self-containing credential description in the credential offer - I think that would require signing credential offer because credential offer is fundamentally not trusted, which I think should be avoided. 
You seem to have concern around "needing to use issuer's metadata credentials_supported", but I am not sure I fully understant what is it? 
would enabling a new option, where instead of fetching entire issuer metadata and searching within it itself, the wallet can send a request to the issuer using the identifier it has obtained from the credential offer so that the issuer can return only the part of the credentials_supported object that the wallet needs? @pmhsfelix

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -689,6 +727,7 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
 * `format`: REQUIRED. Format of the Credential to be issued. This Credential format identifier determines further parameters required to determine the type and (optionally) the content of the credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles).
```
Sakurann, line 727:
```suggestion
* `format`: REQUIRED if `identifier` was not returned from the Token Response. Format of the Credential to be issued. This Credential format identifier determines further parameters required to determine the type and (optionally) the content of the credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles).
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: I think no need for a new issue. discussion is happening in issue #77

Sakurann: I think no need for a new issue. discussion is happening in issue #77

Sakurann: also I think issue #42 might be related.

Sakurann: this topic is out of scope for this PR, I will revert this change. I was just trying to clarify the text, so my changes here did not introduce any normative changes.

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
Sakurann: this topic is out of scope for this PR, I will revert this change. I was just trying to clarify the text, so my changes here did not introduce any normative changes.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: +1, but this topic is out of scope for this PR, opened a new issue #81 to track this

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
Sakurann: changes applied

Sakurann: great questions!
I started by clarifying the definition of the c_identifiers, which is in scope for this PR.
I think higher level explanation should be done in the PR to issue #77.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -602,8 +603,11 @@ In addition to the response parameters defined in [@!RFC6749], the AS MAY return
...
 * `c_nonce_expires_in`: OPTIONAL. JSON integer denoting the lifetime in seconds of the `c_nonce`.
+* `c_identifiers`: OPTIONAL. JSON array of JSON strings that each identify a credential that can be issued using Access Token returned in this response. This parameter is used to uniquely identify a credential beyond the combination of credential format and type. For example, when the Credential Issuer is issuing credentials of the same type, but with different subset of claims. This parameter MUST be used when `scope` parameter was used in the Authorization Request to request credential type of a credential that is being uniquely identified. When received, the Wallet MUST use these values together with an Access Token in the subsequent Credential Request(s).
+* `authorization_details`: REQUIRED when `authorization_details` parameter was used to request issuance of a certain Credential type as defined in (#authorization-details). A JSON array of objects as defined in Section 7 of [@!RFC9396]. In addition to the parameters received from the Wallet, the AS MAY return the following parameter:
```
peppelinux, line 607:
```suggestion
* `authorization_details`: REQUIRED when `authorization_details` parameter is used for requesting the issuance of a certain Credential type, as defined in (#authorization-details). A JSON array of objects as defined in Section 7 of [@!RFC9396]. In addition to the parameters received from the Wallet, the AS MAY return the following parameter:
```

peppelinux (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -602,8 +603,11 @@ In addition to the response parameters defined in [@!RFC6749], the AS MAY return
 
 * `c_nonce`: OPTIONAL. JSON string containing a nonce to be used to create a proof of possession of key material when requesting a Credential (see (#credential_request)). When received, the Wallet MUST use this nonce value for its subsequent Credential Requests until the Credential Issuer provides a fresh nonce.
 * `c_nonce_expires_in`: OPTIONAL. JSON integer denoting the lifetime in seconds of the `c_nonce`.
+* `c_identifiers`: OPTIONAL. JSON array of JSON strings that each identify a credential that can be issued using Access Token returned in this response. This parameter is used to uniquely identify a credential beyond the combination of credential format and type. For example, when the Credential Issuer is issuing credentials of the same type, but with different subset of claims. This parameter MUST be used when `scope` parameter was used in the Authorization Request to request credential type of a credential that is being uniquely identified. When received, the Wallet MUST use these values together with an Access Token in the subsequent Credential Request(s).
```
peppelinux, line 606:
```suggestion
* `c_identifiers`: OPTIONAL. JSON array of JSON strings that each identify a Credential that can be issued using Access Token returned in this response. This parameter is used to uniquely identify a Credential beyond the combination of Credential format and type. For example, when the Credential Issuer is issuing Credentials of the same type, but with different subset of claims. This parameter MUST be used when the `scope` parameter is used in the Authorization Request to request the type of a Credential that is being uniquely identified. When received, the Wallet MUST use these values together with an Access Token in the subsequent Credential Request(s).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -602,8 +603,11 @@ In addition to the response parameters defined in [@!RFC6749], the AS MAY return
...
+* `authorization_details`: REQUIRED when `authorization_details` parameter was used to request issuance of a certain Credential type as defined in (#authorization-details). A JSON array of objects as defined in Section 7 of [@!RFC9396]. In addition to the parameters received from the Wallet, the AS MAY return the following parameter:
+  * `c_identifiers`: OPTIONAL. JSON array of JSON strings that each identify a credential that can be issued using Access Token returned in this response. This parameter is used to uniquely identify a credential beyond the combination of credential format and type. For example, when Credential Issuer is issuing credentials of the same type, but with different subset of claims. This parameter MUST be used when `authorization_details` parameter was used in the Authorization Request to request credential type of a credential that is being uniquely identified. When received, the Wallet MUST use these values together with an Access Token in the subsequent Credential Request(s).
```
peppelinux, line 608:
```suggestion
  * `c_identifiers`: OPTIONAL. JSON array of JSON strings that each identify a Credential that can be issued using Access Token returned in this response. This parameter is used to uniquely identify a Credential beyond the combination of Credential format and type. For example, when Credential Issuer is issuing Credentials of the same type, but with different subset of claims. This parameter MUST be used when `authorization_details` parameter is used in the Authorization Request to request Credential type of a Credential that is being uniquely identified. When received, the Wallet MUST use these values together with an Access Token in the subsequent Credential Request(s).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -602,8 +603,11 @@ In addition to the response parameters defined in [@!RFC6749], the AS MAY return
...
 
-Below is a non-normative example of a Token Response:
+Below is a non-normative example of a Token Response when `scope` parameter was used to request issuance of a certain Credential type:
```
peppelinux, line 610:
```suggestion
Below is a non-normative example of a Token Response when the `scope` parameter was used to request the issuance of a certain Credential type:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -686,9 +720,10 @@ For cryptographic binding, the Client has the following options to provide crypt
 
 A Client makes a Credential Request to the Credential Endpoint by sending the following parameters in the entity-body of an HTTP POST request using the `application/json` media type.
 
-* `format`: REQUIRED. Format of the Credential to be issued. This Credential format identifier determines further parameters required to determine the type and (optionally) the content of the credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles).
+* `format`: REQUIRED if `c_identifier` was not returned from the Token Response. Format of the Credential to be issued. This Credential format identifier determines further parameters required to determine the type and (optionally) the content of the credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles). When this parameter is used, `c_identifier` parameter MUST NOT be present.
```
peppelinux, line 723:
```suggestion
* `format`: REQUIRED when the `c_identifier` was not returned from the Token Response. This parameter determines the format of the Credential to be issued, which may determine the type and any other information related to the Credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles). When this parameter is used, `c_identifier` parameter MUST NOT be present.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -686,9 +720,10 @@ For cryptographic binding, the Client has the following options to provide crypt
...
 * `proof`: OPTIONAL. JSON object containing proof of possession of the key material the issued Credential shall be bound to.  The `proof` object MUST contain a following claim:
```
peppelinux, line 724:
```suggestion
* `proof`: OPTIONAL. JSON object containing proof of possession of the cryptographic key material the issued Credential shall be bound to.  The `proof` object MUST contain a following claim:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -686,9 +720,10 @@ For cryptographic binding, the Client has the following options to provide crypt
...
     * `proof_type`: REQUIRED. JSON string denoting the key proof type. The value of this claim determines other claims in the key proof object and its respective processing rules. Key proof types defined in this specification can be found in (#proof_types).
```
peppelinux, line 725:
```suggestion
    * `proof_type`: REQUIRED. JSON string identifying the cryptographic key proof type. The value of this claim determines other claims in the cryptographic key proof object and its respective processing rules. Cryptographic key proof types defined in this specification can be found in (#proof_types).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -686,9 +720,10 @@ For cryptographic binding, the Client has the following options to provide crypt
...
+* `c_identifier`: REQUIRED if `c_identifier` was returned from the Token Response. JSON string that identifies a credential that is being requested to be issued. It MUST be present if Credential Issuer returned `c_identifiers` parameter the Token Response. When this parameter is used, `format` parameter and any other credential format specific set of parameters such as those defined in (#format_profiles) MUST NOT be present.
```
peppelinux, line 726:
```suggestion
* `c_identifier`: REQUIRED if `c_identifier` was returned from the Token Response. JSON string that identifies a Credential that is being requested to be issued. It MUST be present if the Credential Issuer returned `c_identifiers` parameter in the Token Response. When this parameter is used, the `format` parameter and any other Credential format specific set of parameters such as those defined in (#format_profiles) MUST NOT be present.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -686,9 +720,10 @@ For cryptographic binding, the Client has the following options to provide crypt
...
 * `credential_encryption_jwk`: OPTIONAL. A JSON object containing a single public key as a JWK used for encrypting the Credential Response.
 * `credential_response_encryption_alg`: OPTIONAL. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] REQUIRED for encrypting Credential and/or Batch Credential Responses. The default, if omitted, is that no encryption is performed. If `credential_response_encryption_alg` is present, `credential_encryption_jwk` MUST be present.
```
peppelinux, line 728:
```suggestion
* `credential_response_encryption_alg`: OPTIONAL. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] REQUIRED for encrypting Credential and/or Batch Credential Responses. If omitted, no encryption is intended to be performed. When the `credential_response_encryption_alg` is present, the `credential_encryption_jwk` MUST be present.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1189,7 +1225,7 @@ It is dependent on the Credential format where the available claims will appear
 
 The AS MUST be able to determine from the Issuer metadata what claims are disclosed with the requested credentials to be able to render a meaningful user consent.
 
-The following example shows a non-normative example of an object comprising `credentials_supported` parameter for a credential in JWT VC format (JSON encoding).
+The following is a non-normative example of an object comprising `credentials_supported` parameter for a credential in JWT VC format (JSON encoding).
```
peppelinux, line 1228:
```suggestion
The following is a non-normative example of an object comprising `credentials_supported` parameter for a Credential in JWT VC format (JSON encoding).
```

peppelinux (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -461,7 +462,7 @@ Credential Issuers MUST ignore unknown scope values in a request.
 
 If the Credential Issuer metadata contains an `authorization_server` property, it is RECOMMENDED to use a `resource` parameter [@!RFC8707] whose value is the Credential Issuer's identifier value to allow the AS to differentiate Credential Issuers.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
If the Credential Issuer metadata contains an `authorization_server` property, it is RECOMMENDED to use a `resource` parameter [@!RFC8707] whose value is the Credential Issuer's identifier value. This allows the AS to differentiate Credential Issuers.  
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -461,7 +462,7 @@ Credential Issuers MUST ignore unknown scope values in a request.
 
 If the Credential Issuer metadata contains an `authorization_server` property, it is RECOMMENDED to use a `resource` parameter [@!RFC8707] whose value is the Credential Issuer's identifier value to allow the AS to differentiate Credential Issuers.  
 
-Below is a non-normative example of an Authorization Request using the scope `com.example.healthCardCredential` that would be sent by the User Agent to the Authorization Server in response to an HTTP 302 redirect response by the Wallet (with line wraps within values for display purposes only):
+Below is a non-normative example of an Authorization Request using the scope `UniversityDegree_JWT` that would be sent by the User Agent to the Authorization Server in response to an HTTP 302 redirect response by the Wallet (with line wraps within values for display purposes only):
```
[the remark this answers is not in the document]
Sakurann: i am honestly not sure what is wrong with the current text...

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -602,8 +603,11 @@ In addition to the response parameters defined in [@!RFC6749], the AS MAY return
 
 * `c_nonce`: OPTIONAL. JSON string containing a nonce to be used to create a proof of possession of key material when requesting a Credential (see (#credential_request)). When received, the Wallet MUST use this nonce value for its subsequent Credential Requests until the Credential Issuer provides a fresh nonce.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
* `c_nonce`: OPTIONAL. JSON string containing a nonce to be used when creating a proof of possession of the key proof (see (#credential_request)). When received, the Wallet MUST use this nonce value for its subsequent requests until the Credential Issuer provides a fresh nonce.
```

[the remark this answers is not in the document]
Sakurann: ideally, I would prefer to refrain from making changes to the sections not related to the PR..

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -602,8 +603,11 @@ In addition to the response parameters defined in [@!RFC6749], the AS MAY return
...
 * `c_nonce_expires_in`: OPTIONAL. JSON integer denoting the lifetime in seconds of the `c_nonce`.
```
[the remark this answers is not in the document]
Sakurann: I prefer the current text. this is out of scope for this PR - if this is an issue, please raise another PR.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* `authorization_details`: REQUIRED when `authorization_details` parameter is used to request issuance of a certain Credential type as defined in (#authorization-details). A JSON array of objects as defined in Section 7 of [@!RFC9396]. In addition to the parameters received from the Wallet, the AS MAY return the following parameter:
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: really like this suggestion

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -407,9 +407,10 @@ There are two possible ways to request issuance of a specific Credential type in
 The request parameter `authorization_details` defined in Section 2 of [@!RFC9396] MUST be used to convey the details about the Credentials the Wallet wants to obtain. This specification introduces a new authorization details type `openid_credential` and defines the following elements to be used with this authorization details type:
 
 * `type` REQUIRED. JSON string that determines the authorization details type. MUST be set to `openid_credential` for the purpose of this specification.
-* `format`: REQUIRED. JSON string representing the format in which the Credential is requested to be issued. This Credential format identifier determines further claims in the authorization details object specifically used to identify the Credential type to be issued. This specification defines Credential Format Profiles in (#format_profiles).  
+* `format`: REQUIRED. JSON string representing the format in which the Credential is requested to be issued.
+* Further parameters characterizing the type of the credential to be issued: REQUIRED. These parameters are specific to the credential format profile, some of which are defined in (#format_profiles).
```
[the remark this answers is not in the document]
Sakurann: some of those additional parameters are required. so RECOMMENDED is not accurate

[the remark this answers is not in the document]
peppelinux: Ok, a rewording is then required.
The reader needs to know in a explicit way which parameter is required and which not

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
peppelinux: Nothing wrong, mine is a proof reading during the revision with words that came up to my mind. If something sounds useful please keep it otherwise reject

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
peppelinux: It's text that gives more guidance to the implementers that get instructions about the meaning of the parameters and how their implementation should behave. I think that's not something for an issue but something for this text in this place. If you think that's not required or interesting: no worries

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
Sakurann: that's what this bulletpoint tries to say: that `Further parameters characterizing the type of the credential to be issued` are REQUIRED. I will just revert this part because it is not part of this PR.

Sakurann:
Hi, updated the PR based on the direction discussed during the IIW week:

- remove an option to use credential instance identifiers when scopes are used, allow to use it only when authorization_details are used in authz req. Because it is not clear how best to map scopes to the format/type to the instance. But agreed that @pmhsfelix will open an issue discussing whether an option to use c_instance_identifier with scopes is needed and if yes, what is the best mechanism (one option is a top-level `c_instance_identifier` token response parameter which was what I originally put, alternative is using a `scope` token response parameter and use it to express relationship between scope value and the isntance identifier as in `<scope value>:c_instance_identifier>`).
- clarifying that `c_instance_identifier` can be used to simplify the credential request, not only to differentiate credentials of the same type/format, but different subset of claims.
- renamed to `c_instance_identifier` to differentiate with the credential identifier being introduced in the credential offer in another PR
- i introduced a c_instance_identifier_supported credential issuer metadata parameter, because when issuer uses this new parameter, the wallet is required to use it too, so it felt necessary, but not sure.

cc @tlodderstedt @paulbastian @c2bo @jogu @decentralgabe

peppelinux (changes requested):

decentralgabe (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -602,10 +602,14 @@ Token Responses are made as defined in [@!RFC6749].
...
+* `authorization_details`: REQUIRED when `authorization_details` parameter is used to request issuance of a certain Credential type as defined in (#authorization-details). A JSON array of objects as defined in Section 7 of [@!RFC9396]. This specification defines the following parameter to be used with authorization details type `openid_credential` in the Token Response:
+  * `c_instance_identifiers`: OPTIONAL. JSON array of JSON strings that each uniquely identify a Credential instance that can be issued using Access Token returned in this response. Each Credential instance can be a unique Credential of the same format and type, but with different claim values or different subset of claims within the claimset identified by the Credential type. This parameter can also be used to simplify the Credential Request, since as defined in (#credential_request) `c_instance_identifier` parameter replaces `format` and any other Credential format specific parameters in the Credential Request. When received, the Wallet MUST use these values together with an Access Token in the subsequent Credential Request(s).
```
danielfett, line 608: I missed the discussion on this last week and I have a hard time following this description. Can you please attempt to introduce the concept or mental model behind "instances" of credentials before diving into the definition of the parameter?

danielfett (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
pmhsfelix, line 608: Detail: since `c_instance_identifiers` is already inside an `authorization_details` item with type=`openid_credential`, perhaps we can drop the `c_` prefix. The same for the credential request parameter.
The `c_` prefix was important if the parameter was a top-level one in the token response.

pmhsfelix, line 608: Detail: I'm not 100% comfortable with the `instance_identifier` naming, namely because:
- The credential was not yet created (that will only happen after a credential request), so there isn't yet an instance, right?
- Also, this identifier may be confused with the `callback_id` from https://github.com/openid/OpenID4VCI/pull/70/files, which is described as `A JSON string identifying an issued Credential`.

Would `credential_request_identifier`(or an abbreviation of it) be a better name , since it provides an identifier to request a specific thing on the credential request.

pmhsfelix (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -461,9 +461,9 @@ occurrence MUST be interpreted individually.
 
 Credential Issuers MUST ignore unknown scope values in a request.
 
-If the Credential Issuer metadata contains an `authorization_server` property, it is RECOMMENDED to use a `resource` parameter [@!RFC8707] whose value is the Credential Issuer's identifier value to allow the AS to differentiate Credential Issuers.  
+If the Credential Issuer metadata contains an `authorization_server` property, it is RECOMMENDED to use a `resource` parameter [@!RFC8707] whose value is the Credential Issuer's identifier value. This allows the AS to differentiate Credential Issuers.  
 
-Below is a non-normative example of an Authorization Request using the scope `com.example.healthCardCredential` that would be sent by the User Agent to the Authorization Server in response to an HTTP 302 redirect response by the Wallet (with line wraps within values for display purposes only):
+Below is a non-normative example of an Authorization Request provided by the Wallet to the Authorization Server using the scope `UniversityDegree_JWT` and in response to an HTTP 302 redirect. The line wraps within the values are intended for display purposes only:
```
peppelinux, line 466:
```suggestion
Below is a non-normative example of an Authorization Request provided by the Wallet to the Authorization Server using the scope `university-degree-jwt` and in response to an HTTP 302 redirect. The line wraps within the values are intended for display purposes only:
```

peppelinux (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: I don't understand the suggestion. "UniversityDegree_JWT" is what is used in the example below.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -688,11 +705,12 @@ For cryptographic binding, the Client has the following options to provide crypt
...
 
-* `format`: REQUIRED. Format of the Credential to be issued. This Credential format identifier determines further parameters required to determine the type and (optionally) the content of the credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles).
-* `proof`: OPTIONAL. JSON object containing proof of possession of the key material the issued Credential shall be bound to.  The `proof` object MUST contain a following claim:
+* `format`: REQUIRED when the `c_instance_identifier` was not returned from the Token Response. This parameter determines the format of the Credential to be issued, which may determine the type and any other information related to the Credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles). When this parameter is used, `c_instance_identifier` parameter MUST NOT be present.
+* `proof`: OPTIONAL. JSON object containing proof of possession of the cryptographic key material the issued Credential shall be bound to.  The `proof` object MUST contain a following claim:
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
* `proof`: OPTIONAL. JSON object containing proof of possession of the cryptographic key material the issued Credential would be bound to.  The `proof` object MUST contain a following claim:
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: thank you for the feedback - I tried add more description in the beginning of this section, before parameters are being defined. please take a look, and if it is not clear still, would appreciate concrete text suggestions.

Sakurann: during SIOP call on the week of Oct 16th, we discussed that a term instance does not necessarily mean after issuance and can identify credentials before they are being issued.

Sakurann: i don't have a strong opinion on this, but having `c_` is more specific which is usually better....

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -600,12 +600,18 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
 
 Token Responses are made as defined in [@!RFC6749].
 
+The AS might decide to authorize issuance of multiple instances for each Credential requested in the Authorization Request. Each Credential instance is described using the same entry in the `credentials_supported` Credential Issuer metadata, but contains different claim values or different subset of claims within the claimset identified by the Credential type.
```
danielfett, line 603: Maybe it would be good to find a different word instead of "credential type" to describe an entry in the `credentials_supported` Credential Issuer metadata. The terms feels too overloaded.

How about "credential description" or "credential definition"?

openid-4-verifiable-credential-issuance-1_0.md
danielfett: Thanks, I think it is now much clearer!

One clarification question, and sorry if this has been answered elsewhere:

So the relationship currently described is the in any issuance process, multiple "types" of credentials can be issued (*), and for each "type" there can be multiple instances. An instance is described as "contains different claim values or different subset of claims within the claimset identified by the Credential type"
```
(issuance process) --- 1:n --- (credential definition*) --- 1:m --- (credential instance)
```

How is the relationship to batch issuance for unlinkability as suggested in SD-JWT? Is this already covered or should there be a separate mechanism for that? I.e., to have something like this for k copies of the same credential:

```
(issuance process) --- 1:n --- (credential definition*) --- 1:m --- (credential instance) --- 1:k --- (credential instance copy)
```

----

(*) entry in the `credentials_supported` Credential Issuer metadata - I added a comment suggesting to find a good term for this to use instead of credential type.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -406,12 +406,12 @@ There are two possible ways to request issuance of a specific Credential type in
 
 ### Request Issuance of a Certain Credential Type using `authorization_details` Parameter {#authorization-details}
 
-The request parameter `authorization_details` defined in Section 2 of [@!RFC9396] MUST be used to convey the details about the Credentials the Wallet wants to obtain. This specification introduces a new authorization details type `openid_credential` and defines the following elements to be used with this authorization details type:
+The request parameter `authorization_details` defined in Section 2 of [@!RFC9396] MUST be used to convey the details about the Credentials the Wallet wants to obtain. This specification introduces a new authorization details type `openid_credential` and defines the following parameters to be used with this authorization details type:
 
 * `type` REQUIRED. JSON string that determines the authorization details type. MUST be set to `openid_credential` for the purpose of this specification.
```
danielfett, line 411:
```suggestion
* `type` REQUIRED. String that determines the authorization details type. MUST be set to `openid_credential` for the purpose of this specification.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -406,12 +406,12 @@ There are two possible ways to request issuance of a specific Credential type in
...
-* `format`: REQUIRED. JSON string representing the format in which the Credential is requested to be issued. This Credential format identifier determines further claims in the authorization details object specifically used to identify the Credential type to be issued. This specification defines Credential Format Profiles in (#format_profiles).  
+* `format`: REQUIRED. JSON string representing the format in which the Credential is requested to be issued. This Credential format identifier determines further claims in the authorization details object specifically used to identify the Credential type to be issued. This specification defines Credential Format Profiles in (#format_profiles).
```
danielfett, line 412:
```suggestion
* `format`: REQUIRED. String representing the format in which the Credential is requested to be issued. This Credential format identifier determines further claims in the authorization details object specifically used to identify the Credential type to be issued. This specification defines Credential Format Profiles in (#format_profiles).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -600,12 +600,18 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
 
 Token Responses are made as defined in [@!RFC6749].
 
+The AS might decide to authorize issuance of multiple instances for each Credential requested in the Authorization Request. Each Credential instance is described using the same entry in the `credentials_supported` Credential Issuer metadata, but contains different claim values or different subset of claims within the claimset identified by the Credential type.
+
 In addition to the response parameters defined in [@!RFC6749], the AS MAY return the following parameters:
 
-* `c_nonce`: OPTIONAL. JSON string containing a nonce to be used to create a proof of possession of key material when requesting a Credential (see (#credential_request)). When received, the Wallet MUST use this nonce value for its subsequent Credential Requests until the Credential Issuer provides a fresh nonce.
+* `c_nonce`: OPTIONAL. JSON string containing a nonce to be used when creating a proof of possession of the key proof (see (#credential_request)). When received, the Wallet MUST use this nonce value for its subsequent requests until the Credential Issuer provides a fresh nonce.
```
danielfett, line 607:
```suggestion
* `c_nonce`: OPTIONAL. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential_request)). When received, the Wallet MUST use this nonce value for its subsequent requests until the Credential Issuer provides a fresh nonce.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -600,12 +600,18 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
...
+* `authorization_details`: REQUIRED when `authorization_details` parameter is used to request issuance of a certain Credential type as defined in (#authorization-details). A JSON array of objects as defined in Section 7 of [@!RFC9396]. This specification defines the following parameter to be used with authorization details type `openid_credential` in the Token Response:
+  * `c_instance_identifiers`: OPTIONAL. JSON array of JSON strings that each uniquely identify a Credential instance that can be issued using Access Token returned in this response. Each Credential instance is a unique Credential described using the same entry in the `credentials_supported` Credential Issuer metadata, but can contain different claim values or different subset of claims within the claimset identified by the Credential type. This parameter can also be used to simplify the Credential Request, since as defined in (#credential_request) `c_instance_identifier` parameter replaces `format` and any other Credential format specific parameters in the Credential Request. When received, the Wallet MUST use these values together with an Access Token in the subsequent Credential Request(s).
```
danielfett, line 610:
```suggestion
* `authorization_details`: REQUIRED when `authorization_details` parameter is used to request issuance of a certain Credential type as defined in (#authorization-details). An array of objects as defined in Section 7 of [@!RFC9396]. This specification defines the following parameter to be used with authorization details type `openid_credential` in the Token Response:
  * `c_instance_identifiers`: OPTIONAL. Array of strings that each uniquely identify a Credential instance that can be issued using Access Token returned in this response. Each Credential instance is a unique Credential described using the same entry in the `credentials_supported` Credential Issuer metadata, but can contain different claim values or different subset of claims within the claimset identified by the Credential type. This parameter can also be used to simplify the Credential Request, since as defined in (#credential_request) `c_instance_identifier` parameter replaces `format` and any other Credential format specific parameters in the Credential Request. When received, the Wallet MUST use these values together with an Access Token in the subsequent Credential Request(s).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -688,11 +707,12 @@ For cryptographic binding, the Client has the following options to provide crypt
 
 A Client makes a Credential Request to the Credential Endpoint by sending the following parameters in the entity-body of an HTTP POST request using the `application/json` media type.
 
-* `format`: REQUIRED. Format of the Credential to be issued. This Credential format identifier determines further parameters required to determine the type and (optionally) the content of the credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles).
-* `proof`: OPTIONAL. JSON object containing proof of possession of the key material the issued Credential shall be bound to.  The `proof` object MUST contain a following claim:
+* `format`: REQUIRED when the `c_instance_identifier` was not returned from the Token Response. This parameter determines the format of the Credential to be issued, which may determine the type and any other information related to the Credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles). When this parameter is used, `c_instance_identifier` parameter MUST NOT be present.
```
danielfett, line 710: Can `format` be used if the condition does not apply? Please add a "MUST NOT be used otherwise" or "MAY be used otherwise" somewhere.

danielfett, line 710: Data type is missing.
```suggestion
* `format`: REQUIRED when the `c_instance_identifier` was not returned from the Token Response. String that determines the format of the Credential to be issued, which may determine the type and any other information related to the Credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles). When this parameter is used, `c_instance_identifier` parameter MUST NOT be present.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -688,11 +707,12 @@ For cryptographic binding, the Client has the following options to provide crypt
...
+* `proof`: OPTIONAL. JSON object containing proof of possession of the cryptographic key material the issued Credential would be bound to.  The `proof` object MUST contain a following claim:
```
danielfett, line 711:
```suggestion
* `proof`: OPTIONAL. Object containing the proof of possession of the cryptographic key material the issued Credential would be bound to.  The `proof` object MUST contain a following claim:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -688,11 +707,12 @@ For cryptographic binding, the Client has the following options to provide crypt
...
     * `proof_type`: REQUIRED. JSON string denoting the key proof type. The value of this claim determines other claims in the key proof object and its respective processing rules. Key proof types defined in this specification can be found in (#proof_types).
+* `c_instance_identifier`: REQUIRED when `c_instance_identifier` was returned from the Token Response. JSON string that identifies a Credential that is being requested to be issued. When this parameter is used, the `format` parameter and any other Credential format specific set of parameters such as those defined in (#format_profiles) MUST NOT be present.
```
danielfett, line 713:
```suggestion
* `c_instance_identifier`: REQUIRED when `c_instance_identifier` was returned from the Token Response. String that identifies a Credential that is being requested to be issued. When this parameter is used, the `format` parameter and any other Credential format specific set of parameters such as those defined in (#format_profiles) MUST NOT be present.
```

danielfett: As above, please specify the "otherwise" case.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -688,11 +707,12 @@ For cryptographic binding, the Client has the following options to provide crypt
...
 * `credential_encryption_jwk`: OPTIONAL. A JSON object containing a single public key as a JWK used for encrypting the Credential Response.
```
danielfett, line 714:
```suggestion
* `credential_encryption_jwk`: OPTIONAL. An object containing a single public key as a JWK used for encrypting the Credential Response.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1158,7 +1190,8 @@ This specification defines the following Credential Issuer Metadata:
 * `deferred_credential_endpoint`: OPTIONAL. URL of the Credential Issuer's Deferred Credential Endpoint. This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Credential Issuer does not support the Deferred Credential Endpoint.
 * `credential_response_encryption_alg_values_supported`: OPTIONAL. JSON array containing a list of the JWE [@!RFC7516] encryption algorithms (`alg` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
 * `credential_response_encryption_enc_values_supported`: OPTIONAL. JSON array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
```
danielfett, line 1192:
```suggestion
* `deferred_credential_endpoint`: OPTIONAL. URL of the Credential Issuer's Deferred Credential Endpoint. This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Credential Issuer does not support the Deferred Credential Endpoint.
* `credential_response_encryption_alg_values_supported`: OPTIONAL. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`alg` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
* `credential_response_encryption_enc_values_supported`: OPTIONAL. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
```

danielfett (approved):
I left a bunch of mostly editorial comments and a question, but generally this looks good to me.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: we used to say `string` and i don't remember who, but someone said it should be `JSON string` and we changed... If we change to string here, we should change throughout the spec..

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: great suggestion! thank you, Daniel.

```suggestion
The AS might decide to authorize issuance of multiple instances for each Credential requested in the Authorization Request. Each Credential instance is described using the same entry in the `credentials_supported` Credential Issuer metadata, but contains different claim values or different subset of claims within the claimset identified by the Credential description.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -600,12 +600,18 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
...
+
 In addition to the response parameters defined in [@!RFC6749], the AS MAY return the following parameters:
 
-* `c_nonce`: OPTIONAL. JSON string containing a nonce to be used to create a proof of possession of key material when requesting a Credential (see (#credential_request)). When received, the Wallet MUST use this nonce value for its subsequent Credential Requests until the Credential Issuer provides a fresh nonce.
+* `c_nonce`: OPTIONAL. JSON string containing a nonce to be used when creating a proof of possession of the key proof (see (#credential_request)). When received, the Wallet MUST use this nonce value for its subsequent requests until the Credential Issuer provides a fresh nonce.
 * `c_nonce_expires_in`: OPTIONAL. JSON integer denoting the lifetime in seconds of the `c_nonce`.
+* `authorization_details`: REQUIRED when `authorization_details` parameter is used to request issuance of a certain Credential type as defined in (#authorization-details). A JSON array of objects as defined in Section 7 of [@!RFC9396]. This specification defines the following parameter to be used with authorization details type `openid_credential` in the Token Response:
```
Sakurann, line 609:
```suggestion
* `authorization_details`: REQUIRED when `authorization_details` parameter is used to request issuance of a certain Credential type as defined in (#authorization-details). MUST NOT be used otherwise. A JSON array of objects as defined in Section 7 of [@!RFC9396]. This specification defines the following parameter to be used with authorization details type `openid_credential` in the Token Response:
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: added "MUST NOT be used otherwise". will wait for a WG's opinion on `string` vs `JSON string` for the rest of the suggestions.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* `format`: REQUIRED when the `c_instance_identifier` was not returned from the Token Response. MUST NOT be used otherwise. This parameter determines the format of the Credential to be issued, which may determine the type and any other information related to the Credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles). When this parameter is used, `c_instance_identifier` parameter MUST NOT be present.
```

openid-4-verifiable-credential-issuance-1_0.md
```suggestion
* `c_instance_identifier`: REQUIRED when `c_instance_identifier` was returned from the Token Response. MUST NOT be used otherwise. JSON string that identifies a Credential that is being requested to be issued. When this parameter is used, the `format` parameter and any other Credential format specific set of parameters such as those defined in (#format_profiles) MUST NOT be present.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -617,7 +623,20 @@ Cache-Control: no-store
...
+        "format": "jwt_vc_json",
+        "credential_definition": {
+          "type": [
+              "VerifiableCredential",
+              "UniversityDegreeCredential"
+          ]
+        },
+        "c_instance_identifiers": [ "CivilEngineeringDegree-2023", "ElectricalEngineeringDegree-2023" ]
+      }
+    ]
```
paulbastian, line 639:
```suggestion
    "c_nonce_expires_in": 86400,
    "authorization_details": [
      {
        "c_instance_identifiers": [ "CivilEngineeringDegree-2023", "ElectricalEngineeringDegree-2023" ]
      }
    ]
```

paulbastian: Are the details really relevant? I would see benefit in keeping the Token Response format agnostic

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -600,12 +600,18 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
...
+  * `c_instance_identifiers`: OPTIONAL. JSON array of JSON strings that each uniquely identify a Credential instance that can be issued using Access Token returned in this response. Each Credential instance is a unique Credential described using the same entry in the `credentials_supported` Credential Issuer metadata, but can contain different claim values or different subset of claims within the claimset identified by the Credential type. This parameter can also be used to simplify the Credential Request, since as defined in (#credential_request) `c_instance_identifier` parameter replaces `format` and any other Credential format specific parameters in the Credential Request. When received, the Wallet MUST use these values together with an Access Token in the subsequent Credential Request(s).
```
paulbastian, line 610: Should c_instance_identifiers only be used when there are multiple credentials offered?

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: If we get multiple credentials offered in a Credential Offer, we haven't actually picked one, so therefore we need those details? We could even use `credentials_supported_identifier` from #86 to solve this easier then.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -617,7 +623,20 @@ Cache-Control: no-store
...
+      {
+        "type": "openid_credential",
+        "format": "jwt_vc_json",
+        "credential_definition": {
+          "type": [
+              "VerifiableCredential",
+              "UniversityDegreeCredential"
+          ]
+        },
+        "c_instance_identifiers": [ "CivilEngineeringDegree-2023", "ElectricalEngineeringDegree-2023" ]
```
paulbastian, line 637:
```suggestion
      {
        "type": "openid_credential",
        "credentials_supported_identifier": "UniversityDegreeCredential",
        "c_instance_identifiers": [ "CivilEngineeringDegree-2023", "ElectricalEngineeringDegree-2023" ]
```

paulbastian: see comments below under second suggestion

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -688,11 +707,12 @@ For cryptographic binding, the Client has the following options to provide crypt
...
 
-* `format`: REQUIRED. Format of the Credential to be issued. This Credential format identifier determines further parameters required to determine the type and (optionally) the content of the credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles).
-* `proof`: OPTIONAL. JSON object containing proof of possession of the key material the issued Credential shall be bound to.  The `proof` object MUST contain a following claim:
+* `format`: REQUIRED when the `c_instance_identifier` was not returned from the Token Response. MUST NOT be used otherwise. This parameter determines the format of the Credential to be issued, which may determine the type and any other information related to the Credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles). When this parameter is used, `c_instance_identifier` parameter MUST NOT be present.
```
paulbastian, line 710: When is format actually relevant?
In Authorization Code Flow we already made a choice in the Auth Request.
In Credential Offer we got all the options with  `c_instance_identifier` in the Token Response and we can pick one here with `c_instance_identifier`. What am I missing?

paulbastian (changes requested):
I think this is a great improvement to the spec.
A few questions remain for me:
- Could we use this to solve the questions whether the Wallet needs to call Credential Endpoint or Batch Credential Endpoint?
- Can we keep Token Response credential format agnostic, i.e. only keep c_instance_identifier instead of repeating the whole authorization_details object?
- Could we even remove `format` from Credential Request and make `c_instance_identifier` mandatory?

We don't need to solve everything in this PR however, so I' fine if we know how to proceed with these questions in the future

openid-4-verifiable-credential-issuance-1_0.md
danielfett: It doesn't make sense to say JSON string, as the data structure is a string and the encoding is JSON, but the JSON encoding is applied to the whole object, not just this data field. There is no need to say JSON here and it may even be misinterpreted as implying a JSON encoding for this string, like so:
```json
{
  "type": "\"example\""
}
```

paulbastian (approved):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: good point on potential misinterpretation.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: On your second comment, nothing prevents AS/RS from using the same value for `credentials_upported_identifier` and `c_instance_identifier`, but conceptually, I really think these parameters needs to be separate.

> Are the details really relevant?

I honestly don't know.. RAR spec says "The AS MAY omit values in the authorization_details to the client." in https://datatracker.ietf.org/doc/html/rfc9396#name-token-response. so I think that means that structure below is technically valid, because "type" is the only required parameter if I am correct. but I am not sure if format and type should be omitted here, maybe when there is only one credential being issued, but not when there are multiple, because having type and format here provides a clear binding between an identifier and type/format.

```suggestion
    "c_nonce_expires_in": 86400,
    "authorization_details": [
      {
        "type": "openid_credential",
        "c_instance_identifiers": [ "CivilEngineeringDegree-2023", "ElectricalEngineeringDegree-2023" ]
      }
    ]
```

Sakurann: You seem to want to make most of the requests in VCI format agnostic. why is that..?

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: The current intention is that c_instance_identifiers can be used regardless of how many credentials are being issued. this means that identto simplify the credential request even if only one credential is being issued.

Sakurann:
SIOP/DCP WG call - agreed to change the identifier parameter name to `credential_identifier` and merge. and continue bikeshedding the name and keep discussing the issues Paul is raising in new issues.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: great question. i think a separate issue is needed to discuss this. my current thinking is, `credential_identifier` is used only to issue multiple credentials of the same type, but different content, and there needs to be another switch for issuing copies of the same credential, as using this identifier for both seems really overloaded... cc issue #91

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: solved by renaming to `credential_identifier` for now. cc issue #91

Sakurann: solved by renaming to credential_identifier for now. cc issue https://github.com/openid/OpenID4VCI/issues/91

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: it is relevant when the multiple credentials of different formats are being issued using `scope` parameter mechanism in the authorization request.

c2bo (approved):

jogu (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -600,12 +600,18 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
...
+* `authorization_details`: REQUIRED when `authorization_details` parameter is used to request issuance of a certain Credential type as defined in (#authorization-details). MUST NOT be used otherwise. An array of objects as defined in Section 7 of [@!RFC9396]. This specification defines the following parameter to be used with authorization details type `openid_credential` in the Token Response:
+  * `credential_identifiers`: OPTIONAL. Array of strings that each uniquely identify a Credential instance that can be issued using Access Token returned in this response. Each Credential instance is a unique Credential described using the same entry in the `credentials_supported` Credential Issuer metadata, but can contain different claim values or different subset of claims within the claimset identified by the Credential type. This parameter can also be used to simplify the Credential Request, since as defined in (#credential_request) `credential_identifier` parameter replaces `format` and any other Credential format specific parameters in the Credential Request. When received, the Wallet MUST use these values together with an Access Token in the subsequent Credential Request(s).
```
jogu, line 610: I think I'm slightly confused by this part:

> `credential_identifier` parameter replaces `format` and any other Credential format specific parameters

Doesn't that mean the wallet doesn't know which credential_identifier is associated with what format?

Sakurann:
PR ended up addressing Issue #29 too by simplifying the credential response even when only one credential is being issued.

PR #69, opened 2023-09-07
ed: clarify that credentialSubject is an object and not a dictionary

Sakurann:
issue #57 and #48.


selfissued:
There are multiple instances of using "dictionary" in this way in the spec. We should correct them all.

tlodderstedt (approved):

jogu (approved):

PR #73, opened 2023-09-11
make the guidance to request subset of claims format agnostic

Sakurann:
based on the comment here: https://github.com/vcstuff/oid4vc-haip-sd-jwt-vc/pull/56#issuecomment-1635190840

clarifying that only the subset of claims can be requested using `credentialSubject` and `claims` parameters

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1326,9 +1326,11 @@ To sender-constrain Access Tokens, see the recommendations in Section 4.10.1 in 
 
 # Privacy Considerations
 
-## Wallet Requesting the Subset of the Claims
+## Data Minimization in the Issued Credential
 
-It is on purpose that the Credential Offer does not contain `credentialSubject` nor `claims` properties, while Authorization Details and Credential Request do. This is because this property is meant to be used by the Wallet to specify which claims it is requesting to be issued out of all the claims the Credential Issuer is capable of issuing for this particular Credential (data minimization), while Credential Offer is a mere "invitation" from the Credential Issuer to the Wallet to start the issuance flow.
+It is on purpose that the Credential Offer does not contain `credentialSubject` nor `claims` properties (other Credential Format Profiles may define other values for this property), while Authorization Details and Credential Request do. This is because these properties are meant to be used by the Wallet to specify which claims it is requesting to be issued out of all the claims the Credential Issuer is capable of issuing for this particular credential. Credential Offer is a mere "invitation" from the Credential Issuer to the Wallet to start the issuance flow.
+
+The Wallet MUST NOT use these properties to request claims that are not part of the claims contained in a particular credential type.
 
 {backmatter}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1328,9 +1328,9 @@ To sender-constrain Access Tokens, see the recommendations in Section 4.10.1 in 
 
 ## Data Minimization in the Issued Credential
 
-It is on purpose that the Credential Offer does not contain `credentialSubject` nor `claims` properties (other Credential Format Profiles may define other values for this property), while Authorization Details and Credential Request do. This is because these properties are meant to be used by the Wallet to specify which claims it is requesting to be issued out of all the claims the Credential Issuer is capable of issuing for this particular credential. Credential Offer is a mere "invitation" from the Credential Issuer to the Wallet to start the issuance flow.
+It is on purpose that the Credential Offer does not contain `credentialSubject` nor `claims` properties (other Credential Format Profiles may define other names for this property), while Authorization Details and Credential Request do. This is because these properties are meant to be used by the Wallet to specify which claims it is requesting to be issued out of all the claims the Credential Issuer is capable of issuing for this particular credential. Credential Offer is a mere "invitation" from the Credential Issuer to the Wallet to start the issuance flow.
 
-The Wallet MUST NOT use these properties to request claims that are not part of the claims contained in a particular credential type.
+The Wallet MUST NOT use these properties to request claims that are not part of the claims contained in a particular credential type or claims not authorized by the respective Access Token (in case of a Credential Request).
 
 {backmatter}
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1326,7 +1326,11 @@ To sender-constrain Access Tokens, see the recommendations in Section 4.10.1 in
 
 # Privacy Considerations
 
-TBD
+## Data Minimization in the Issued Credential
+
+It is on purpose that the Credential Offer does not contain `credentialSubject` nor `claims` properties (other Credential Format Profiles may define other values for this property), while Authorization Details and Credential Request do. This is because these properties are meant to be used by the Wallet to specify which claims it is requesting to be issued out of all the claims the Credential Issuer is capable of issuing for this particular credential. Credential Offer is a mere "invitation" from the Credential Issuer to the Wallet to start the issuance flow.
+
+The Wallet MUST NOT use these properties to request claims that are not part of the claims contained in a particular credential type.
```
tlodderstedt, line 1333:
```suggestion
The Wallet MUST NOT use these properties to request claims that are not part of the claims contained in a particular credential type or claims not authorized by the respective access token (in case of a Credential Request).
```

tlodderstedt (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1326,7 +1326,11 @@ To sender-constrain Access Tokens, see the recommendations in Section 4.10.1 in
 
 # Privacy Considerations
 
-TBD
+## Data Minimization in the Issued Credential
+
+It is on purpose that the Credential Offer does not contain `credentialSubject` nor `claims` properties (other Credential Format Profiles may define other values for this property), while Authorization Details and Credential Request do. This is because these properties are meant to be used by the Wallet to specify which claims it is requesting to be issued out of all the claims the Credential Issuer is capable of issuing for this particular credential. Credential Offer is a mere "invitation" from the Credential Issuer to the Wallet to start the issuance flow.
```
Sakurann, line 1331:
```suggestion
It is on purpose that the Credential Offer does not contain `credentialSubject` nor `claims` properties (other Credential Format Profiles may define other names for this property), while Authorization Details and Credential Request do. This is because these properties are meant to be used by the Wallet to specify which claims it is requesting to be issued out of all the claims the Credential Issuer is capable of issuing for this particular credential. Credential Offer is a mere "invitation" from the Credential Issuer to the Wallet to start the issuance flow.
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
The Wallet MUST NOT use these properties to request claims that are not part of the claims contained in a particular credential type or claims not authorized by the respective Access Token (in case of a Credential Request).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1326,7 +1326,11 @@ To sender-constrain Access Tokens, see the recommendations in Section 4.10.1 in
 
 # Privacy Considerations
 
-TBD
+## Data Minimization in the Issued Credential
+
+It is on purpose that the Credential Offer does not contain `credentialSubject` nor `claims` properties (other Credential Format Profiles may define other names for this property), while Authorization Details and Credential Request do. This is because these properties are meant to be used by the Wallet to specify which claims it is requesting to be issued out of all the claims the Credential Issuer is capable of issuing for this particular credential. Credential Offer is a mere "invitation" from the Credential Issuer to the Wallet to start the issuance flow.
```
peppelinux, line 1331: following other/previous chat on this, I really like the snake case in all the oauth2/openid specs, then `credentialSubject` should became `credential_subject`

A study demonstrates that snake case is more readable: https://www.cs.kent.edu/~jmaletic/papers/ICPC2010-CamelCaseUnderScoreClouds.pdf

we already have snake case in the oauth2/jwt ecosystem and I really would like to preserve it :-)

peppelinux (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: `credentialSubject` is what is used in VCDM, it is `claims` for non-VCDM credential formats  such as[sd-jwt vc](https://github.com/vcstuff/oid4vc-haip-sd-jwt-vc/pull/56/files) and mdocs.
from a processing perspective, I think it is easier if wallet/issuer can just take the credentialSubject object in VCI and map it to the VCDM compliant payload.

if you disagree, please open a separate issue on this, it is out of scope for this PR.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1326,7 +1326,11 @@ To sender-constrain Access Tokens, see the recommendations in Section 4.10.1 in
 
 # Privacy Considerations
 
-TBD
+## Data Minimization in the Issued Credential
+
```
pmhsfelix, line 1330: Probably not directly related to this specific PR, however the server metadata for `jwt_vc_*` formats does include `credentialSubject`. I.e. `credentialSubject` is not used only on requests from the wallet to the server but also on information _advertised_ by the server. Therefore, shouldn't it also be optionally included in the offer, namely when the offer does not match any `credentials_supported` entry in the metadata?

pmhsfelix: Question related to https://github.com/openid/OpenID4VCI/issues/82

Sakurann: yes, I think this PR might become redundant based on #77 too

openid-4-verifiable-credential-issuance-1_0.md
peppelinux, line 1331:
```suggestion
It is on purpose that the Credential Offer does not contain `credentialSubject` nor `claims` properties (other Credential Format Profiles may define other names for this property), while Authorization Details and Credential Request do. This is because these properties are meant to be used by the Wallet to specify which claims it is requesting to be issued out of all the claims the Credential Issuer is capable of issuing for this particular Credential. Credential Offer is a mere "invitation" from the Credential Issuer to the Wallet to start the issuance flow.
```

peppelinux (changes requested):

Sakurann:
with PR #86 merged I think this PR can be closed

Issue #72, opened 2023-09-09
Credential Issuer using multiple authorization servers


Sakurann:
In our implementation, Credential Issuer (RS) will use different Authorization Servers depending on the grant type: authorization code flow will use existing AS in production and there will be a new AS spun out for pre-auth code flow. 

We did our best towards using the same AS but in the end it was impossible because we cannot make changes to the AS in production so that token endpoint accepts pre-auth code and using a new AS as an intermediary AS to the existing production-AS for authz code flow broke SSO and all other features (production AS is needed for authz code to authenticate existing customers). There was also a suggestion to use different credential issuer (RS) per AS, but most credentials can be issued using both of the grant types, so that would mean the wallet would have to determine which issuer to use depending on the grant type which adds a lot of complexity (i.e. making `credential_issuer` property in the credential offer an array so that the wallet can choose a right RS/issuer based on the flow?). 
We explored a lot of options and settled on using the same RS with two ASs. 

This means that Credential Issuer will need to communicate in its metadata endpoints of multiple Authorization Servers, while currently issuer metadata only allows to communicate one. Here, we explored the following three options:

1. defining `authorization_code_authorization_server` and `pre-authorization_code_authorization_server` parameters in the credential issuer metadata, the values of which are respective endpoints. (the parameter names already sound horrible, I know)
2. defining `authorization_servers` parameter as an array of strings that can include multiple entries, the wallet will need to determine which AS to use looking into the metadata of each entry.
3. defining  `authorization_servers` parameter as an array of objects where each object entry clarifies which grant type which AS supports. 
 
To the options 3, @selfissued rightly said this "gives the RS the job of supplying some AS metadata.  That could easily become stale or out of sync with what the AS actually does.  It's probably cleaner to let the AS speak for itself." but I think this comment also applies to option 1. 

@jogu has pointed out that "others are likely to have use cases for multiple authorization servers other than an AS per grant type too.", while option 1 and 3 are pretty specific to AS per grant type.

option 2 is also an approach adopted in [resource server metadata](https://datatracker.ietf.org/doc/draft-ietf-oauth-resource-metadata/) draft, which has recently been adopted by IETF OAuth WG. So I think option 2 is the best option. I did a draft PR #78 to see how it would look and I don't think it's too terrible. In most cases, I would expect `authorization_servers` array would only have one entry, but having this parameter as an array would be very helpful.

Feedback would be appreciated.

tlodderstedt:
[...]

PR #78, opened 2023-09-29
make `authorization_servers` credential issuer metadata parameter an array

Sakurann:
problem statement, and various options that have been considered are explained in Issue #72. 

There are use-cases when Credential Issuer is using different authorization servers depending on the grant type (for example, using existing AS for authz code grant and a new one for pre-auth because was not able to make changes to an existing AS in production.) Inspired by a newly adopted resource metadata draft in oauth wg https://datatracker.ietf.org/doc/html/draft-ietf-oauth-resource-metadata-00#section-2-2.4.

When discussed at oauth security workshop 2023, there have been no concerns with the approach, given that credential issuer needs how to differentiate from which AS each access token is coming from.


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1152,14 +1152,16 @@ The path formed following the steps above MUST point to a JSON document complian
...
 * `credential_response_encryption_enc_values_supported`: OPTIONAL. JSON array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
 * `require_credential_response_encryption`: OPTIONAL. If `require_credential_response_encryption` is `true`, the Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request. If `require_credential_response_encryption` is `false` or omitted, the Credential Issuer indicates that no additional Credential Response encryption is required. If `require_credential_response_encryption` is `true`, `credential_response_encryption_alg_values_supported` MUST also be provided.
 
+Note: When using relying on multiple Authorization Servers for authorization, Credential Issuer MUST be able to identify from which Authorization Server each Access Tokens is coming.
```
jogu, line 1163: I don't know if this sentence is really necessary (it seems to overlap with the general "resource endpoints must check the access token is valid for the resource being access", but if it is this might be a better wording:

```suggestion
Note: When the Credential Issuer accepts access tokens from more than one Authorization Servers, the Credential Issuer MUST be able to identify which Authorization Server issued the Access Token.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1152,14 +1152,16 @@ The path formed following the steps above MUST point to a JSON document complian
...
-* `authorization_server`: OPTIONAL. Identifier of the OAuth 2.0 Authorization Server (as defined in [@!RFC8414]) the Credential Issuer relies on for authorization. If this element is omitted, the entity providing the Credential Issuer is also acting as the AS, i.e., the Credential Issuer's identifier is used as the OAuth 2.0 Issuer value to obtain the Authorization Server metadata as per [@!RFC8414].
+* `authorization_servers`: OPTIONAL. A JSON array of strings, where each string is an identifier of the OAuth 2.0 Authorization Server (as defined in [@!RFC8414]) the Credential Issuer relies on for authorization. Since Credential Issuer may be using different authorization server depending on the Grant Type, the Wallet should determine which Authorization Server to use using `grant_types_supported` Authorization Server metadata parameter (If omitted, the default value is [`authorization_code`]). If this parameter is omitted, the entity providing the Credential Issuer is also acting as the AS, i.e., the Credential Issuer's identifier is used as the OAuth 2.0 Issuer value to obtain the Authorization Server metadata as per [@!RFC8414].
```
jogu, line 1155: This wording feels really specific, whereas https://datatracker.ietf.org/doc/html/draft-ietf-oauth-resource-metadata-00#section-2-2.4 leaves the behaviour quite open and I feel like other people are likely to have use cases for multiple authorization servers other than an AS per grant type.

If we do want to say something then perhaps it should be more along the lines:

"When there are multiple entries in the array the wallet may be able to determine which to use by querying the server metadata, for example depending on the grant type the wallet plans to use examining the `grant_types_supported` values might allow the wallet to eliminate all but one server."

jogu (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: When talking to DanielF, Brian, you and others at OSW about this, this sentence seemed to be the key, but maybe I misunderstood.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: good point. 

```suggestion
* `authorization_servers`: OPTIONAL. A JSON array of strings, where each string is an identifier of the OAuth 2.0 Authorization Server (as defined in [@!RFC8414]) the Credential Issuer relies on for authorization. When there are multiple entries in the array, the Wallet may be able to determine which AS to use by querying the metadata, for example, by examining the `grant_types_supported` values might allow the wallet to eliminate all but one server based on the grant type the wallet plans to use. If this parameter is omitted, the entity providing the Credential Issuer is also acting as the AS, i.e., the Credential Issuer's identifier is used as the OAuth 2.0 Issuer value to obtain the Authorization Server metadata as per [@!RFC8414].
```

openid-4-verifiable-credential-issuance-1_0.md
jogu: I do vaguely remember that conversation but I can't remember why anyone felt it important.

Maybe I'm missing something but I think it's also a bit weird because I don't think it's important that credential issuer determines which authorization server issued an access token, I think it's only important that it verify that it verifies that the token was issued by a server that trusts to issue a token for that particular resource etc. (Which is kind a subtle difference, but I think if both ASs issue JWT access tokens signed by the same key then it's not important for the resource server to identify which of the servers issued the access token because it doesn't matter, the RS can trust it because it's signed with the key it expects.)

bc-pi: In the general case of RFC6750 etc there's no defined way for an RS (Credential Issuer here) to determine the AS that issued an AT.  So an RS that accepts ATs from more than one AS will have to do some nonstandard tricks that might not always work to figure out the issuer. That was the trust of the conversation as I recall.

jogu: Thanks Brian. So are we essentially trying to say something like "Implementing a resource server that accepts tokens from multiple authorization servers makes introspecting the token hard and (as of the time of writing) pretty much the only sensible way is to only do this is with authorization servers that use https://datatracker.ietf.org/doc/html/rfc9068 or a proprietary AS/RS collusion that allows the RS to figure out what permissions are granted by the token"?

bc-pi: I'm not sure I want to get into judgements about what's sensible or not :) But basically yes.

jogu: I can live without the word "sensible" there :-)

I'm not sure the current sentence in this PR really gets that nuance across. I'm not sure the suggested sentence I wrote in my previous comment is great either.

peppelinux: @jogu `access tokens` -> `Access Tokens`

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1152,14 +1152,16 @@ The path formed following the steps above MUST point to a JSON document complian
...
-* `authorization_server`: OPTIONAL. Identifier of the OAuth 2.0 Authorization Server (as defined in [@!RFC8414]) the Credential Issuer relies on for authorization. If this element is omitted, the entity providing the Credential Issuer is also acting as the AS, i.e., the Credential Issuer's identifier is used as the OAuth 2.0 Issuer value to obtain the Authorization Server metadata as per [@!RFC8414].
+* `authorization_servers`: OPTIONAL. A JSON array of strings, where each string is an identifier of the OAuth 2.0 Authorization Server (as defined in [@!RFC8414]) the Credential Issuer relies on for authorization. When there are multiple entries in the array, the Wallet may be able to determine which AS to use by querying the metadata, for example, by examining the `grant_types_supported` values might allow the wallet to eliminate all but one server based on the grant type the wallet plans to use. If this parameter is omitted, the entity providing the Credential Issuer is also acting as the AS, i.e., the Credential Issuer's identifier is used as the OAuth 2.0 Issuer value to obtain the Authorization Server metadata as per [@!RFC8414].
```
peppelinux, line 1155:
```suggestion
* `authorization_servers`: OPTIONAL. A JSON array of strings, where each string is an identifier of the OAuth 2.0 Authorization Server (as defined in [@!RFC8414]) the Credential Issuer relies on for authorization. When there are multiple entries in the array, the Wallet may be able to determine which AS to use by querying the metadata; for example, by examining the `grant_types_supported` values the Wallet might filter the server with the supported grant type it plans to use. If this parameter is omitted, the entity providing the Credential Issuer is also acting as the AS, i.e., the Credential Issuer's identifier is used as the OAuth 2.0 Issuer value to obtain the Authorization Server metadata as per [@!RFC8414].
```

peppelinux (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* `authorization_servers`: OPTIONAL. A JSON array of strings, where each string is an identifier of the OAuth 2.0 Authorization Server (as defined in [@!RFC8414]) the Credential Issuer relies on for authorization. When there are multiple entries in the array, the Wallet may be able to determine which AS to use by querying the metadata; for example, by examining the `grant_types_supported` values, the Wallet can filter the server to use based on the grant type it plans to use. If this parameter is omitted, the entity providing the Credential Issuer is also acting as the AS, i.e., the Credential Issuer's identifier is used as the OAuth 2.0 Issuer value to obtain the Authorization Server metadata as per [@!RFC8414].
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: not exactly happy with the last few phrases, but how about this..?

```suggestion
Note: It can be hard for the Credential Issuer that accepts tokens from multiple Authorization Servers to introspect the Access Token to identify its issuer and permissions granted. Few ways to do so are either by relying on the Authorization Servers that use [@!RFC9068] or by Credential Issuer learning to understand proprietary Access Token structure of Authorization Server.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -320,11 +320,13 @@ This specification defines the following parameters for the Credential Offer obj
...
+  * `issuer_state`: OPTIONAL. String value created by the Credential Issuer and opaque to the Wallet that is used to bind the subsequent Authorization Request with the Credential Issuer to a context set up during previous steps. If the Wallet decides to use the Authorization Code Flow and received a value for this parameter, it MUST include it in the subsequent Authorization Request to the Credential Issuer as the `issuer_state` parameter value.
+  * `authotization_server`: OPTIONAL when a Credential Issuer is relying on multiple Authorization Servers. MUST NOT be used otherwise. String that the Wallet can use to identify the Authorization Server to used with this grant type when `authoziation_servers` parameter in the Credential Issuer metadata has multiple entries. MUST match with one of the values in the `authoziation_servers` array obtained from the Credential Issuer metadata.
```
peppelinux, line 324:
```suggestion
  * `authorization_server`: OPTIONAL when a Credential Issuer is relying on multiple Authorization Servers. MUST NOT be used otherwise. String that the Wallet can use to identify the Authorization Server to used with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```

peppelinux (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -320,11 +320,13 @@ This specification defines the following parameters for the Credential Offer obj
...
   * `interval`: OPTIONAL. The minimum amount of time in seconds that the Wallet SHOULD wait between polling requests to the token endpoint (in case the Authorization Server responds with error code `authorization_pending` - see (#token_error_response)). If no value is provided, Wallets MUST use 5 as the default.
+  * `authotization_server`: OPTIONAL when a Credential Issuer is relying on multiple Authorization Servers. MUST NOT be used otherwise. String that the Wallet can use to identify the Authorization Server to used with this grant type when `authoziation_servers` parameter in the Credential Issuer metadata has multiple entries. MUST match with one of the values in the `authoziation_servers` array obtained from the Credential Issuer metadata.
```
c2bo, line 329:
```suggestion
  * `authorization_server`: OPTIONAL when a Credential Issuer is relying on multiple Authorization Servers. MUST NOT be used otherwise. String that the Wallet can use to identify the Authorization Server to used with this grant type when `authoziation_servers` parameter in the Credential Issuer metadata has multiple entries. MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```

c2bo (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -320,11 +320,13 @@ This specification defines the following parameters for the Credential Offer obj
...
   * `interval`: OPTIONAL. The minimum amount of time in seconds that the Wallet SHOULD wait between polling requests to the token endpoint (in case the Authorization Server responds with error code `authorization_pending` - see (#token_error_response)). If no value is provided, Wallets MUST use 5 as the default.
+  * `authorization_server`: OPTIONAL when a Credential Issuer is relying on multiple Authorization Servers. MUST NOT be used otherwise. String that the Wallet can use to identify the Authorization Server to used with this grant type when `authoziation_servers` parameter in the Credential Issuer metadata has multiple entries. MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```
peppelinux, line 329:
```suggestion
  * `authorization_server`: OPTIONAL when a Credential Issuer is relying on multiple Authorization Servers. MUST NOT be used otherwise. String that the Wallet can use to identify the Authorization Server to used with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```

peppelinux (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -320,11 +320,13 @@ This specification defines the following parameters for the Credential Offer obj
...
   * `user_pin_required`: OPTIONAL. Boolean value specifying whether the AS expects presentation of a user PIN along with the Token Request in a Pre-Authorized Code Flow. Default is `false`. This PIN is intended to bind the Pre-Authorized Code to a certain transaction to prevent replay of this code by an attacker that, for example, scanned the QR code while standing behind the legitimate user. It is RECOMMENDED to send a PIN via a separate channel. If the Wallet decides to use the Pre-Authorized Code Flow, a PIN value MUST be sent in the `user_pin` parameter with the respective Token Request.
   * `interval`: OPTIONAL. The minimum amount of time in seconds that the Wallet SHOULD wait between polling requests to the token endpoint (in case the Authorization Server responds with error code `authorization_pending` - see (#token_error_response)). If no value is provided, Wallets MUST use 5 as the default.
```
peppelinux, line 328:
```suggestion
  * `interval`: OPTIONAL. The minimum amount of time in seconds that the Wallet SHOULD wait between polling requests to the token endpoint (in case the Authorization Server responds with error code `authorization_pending` - see (#token_error_response)). If no value is provided, Wallets MUST use `5` as the default.
```

openid-4-verifiable-credential-issuance-1_0.md
peppelinux, line 329:
```suggestion
  * `authorization_server`: OPTIONAL String that the Wallet can use to identify the Authorization Server to used with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. MUST NOT be used otherwise. MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1152,14 +1154,16 @@ The path formed following the steps above MUST point to a JSON document complian
...
 
+Note: When using relying on multiple Authorization Servers for authorization, Credential Issuer MUST be able to identify from which Authorization Server each Access Tokens is coming.
+
 The following parameter MUST be used to communicate the specifics of the Credential that the Credential Issuer supports issuance of:
 
 * `credentials_supported`: REQUIRED. A JSON array containing a list of JSON objects, each of them representing metadata about a separate credential type that the Credential Issuer can issue. The JSON objects in the array MUST conform to the structure defined in (#credential-metadata-object).
```
peppelinux, line 1169:
```suggestion
* `credentials_supported`: REQUIRED. A JSON array containing a list of JSON objects, each of them representing metadata about a separate Credential type that the Credential Issuer can issue. The JSON objects in the array MUST conform to the structure defined in (#credential-metadata-object).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -320,11 +320,13 @@ This specification defines the following parameters for the Credential Offer obj
...
+  * `issuer_state`: OPTIONAL. String value created by the Credential Issuer and opaque to the Wallet that is used to bind the subsequent Authorization Request with the Credential Issuer to a context set up during previous steps. If the Wallet decides to use the Authorization Code Flow and received a value for this parameter, it MUST include it in the subsequent Authorization Request to the Credential Issuer as the `issuer_state` parameter value.
+  * `authorization_server`: OPTIONAL when a Credential Issuer is relying on multiple Authorization Servers. MUST NOT be used otherwise. String that the Wallet can use to identify the Authorization Server to used with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```
selfissued, line 324:
```suggestion
  * `authorization_server`s: OPTIONAL when a Credential Issuer is relying on multiple Authorization Servers. MUST NOT be used otherwise. String that the Wallet can use to identify the Authorization Server to used with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```

selfissued (approved):

c2bo (approved):

peppelinux (approved):

openid-4-verifiable-credential-issuance-1_0.md
jogu: Building upon Kristina's suggestion this might be slightly better:

```suggestion
Note: It can be challenging for a Credential Issuer that accepts tokens from multiple Authorization Servers to introspect an Access Token to check the validity and determine the permissions granted. Some ways to achieve this are relying on Authorization Servers that use [@!RFC9068] or by the Credential Issuer understanding the proprietary Access Token structures of the Authorization Servers.
```

(I've deliberately removed the mention of 'identify the issuer' - I do remember that being a key part of the OSW discussion, but I don't think we fully finished the discussion - the important aspect is that the RS is able to validate the token and get the permissions granted, and there are proprietary ways to do that if the ASes & RS are all run by the same party.)

bc-pi: @jogu's text looks good to me

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -320,11 +320,13 @@ This specification defines the following parameters for the Credential Offer obj
...
+  * `issuer_state`: OPTIONAL. String value created by the Credential Issuer and opaque to the Wallet that is used to bind the subsequent Authorization Request with the Credential Issuer to a context set up during previous steps. If the Wallet decides to use the Authorization Code Flow and received a value for this parameter, it MUST include it in the subsequent Authorization Request to the Credential Issuer as the `issuer_state` parameter value.
+  * `authorization_servers`: OPTIONAL when a Credential Issuer is relying on multiple Authorization Servers. MUST NOT be used otherwise. String that the Wallet can use to identify the Authorization Server to used with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```
Sakurann, line 324:
```suggestion
  * `authorization_server`: OPTIONAL when a Credential Issuer is relying on multiple Authorization Servers. MUST NOT be used otherwise. String that the Wallet can use to identify the Authorization Server to used with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```

bc-pi (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -320,11 +320,13 @@ This specification defines the following parameters for the Credential Offer obj
...
-  * `interval`: OPTIONAL. The minimum amount of time in seconds that the Wallet SHOULD wait between polling requests to the token endpoint (in case the Authorization Server responds with error code `authorization_pending` - see (#token_error_response)). If no value is provided, Wallets MUST use 5 as the default.
+  * `interval`: OPTIONAL. The minimum amount of time in seconds that the Wallet SHOULD wait between polling requests to the token endpoint (in case the Authorization Server responds with error code `authorization_pending` - see (#token_error_response)). If no value is provided, Wallets MUST use `5` as the default.
+  * `authorization_server`: OPTIONAL String that the Wallet can use to identify the Authorization Server to use with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. It MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```
pmhsfelix, line 329: In the `authorization_code` grant type, the equivalent entry to this one has `MUST NOT be used otherwise`. Should we also add this sentence here?

pmhsfelix (approved):
Approved with just one comment.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: yes, let's add that.

```suggestion
  * `authorization_server`: OPTIONAL String that the Wallet can use to identify the Authorization Server to use with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. MUST NOT be used otherwise. The value of this parameter MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann, line 324:
```suggestion
  * `authorization_server`: OPTIONAL string that the Wallet can use to identify the Authorization Server to use with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. MUST NOT be used otherwise. The value of this parameter MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```

Issue #83, opened 2023-10-03
Credential Offer pointing to a specific `credentials_supported` metadata entry.


Sakurann:
Opening this issue based on the discussion in issue #77.

The question is, as @pmhsfelix put it, `If a Credential Offer requires information from a `credentials_supported` metadata entry in order to be usable, then how does the wallet finds out which entry to use`. 

We have changed string entry in the `credentials` array in the Credential Offer to be `scopes` from `identifier`, but I think we are learning that `identifiers` was better suited for this purpose, so suggestion would be to revert this change to `identifiers`. a) because not all implementations use scopes; and b) there are cases where two credentials of the same type, requested using the same scopes value have a different display property and the issuer needs to differentiate those in the Offer so that the Wallet can display the correct one to the user.

Discussion on how wallet can discover part of the issuer metadata using identifier has a separate issue #42.

Open questions (I think) are:
- relationship between this `co_identifier` in the Credential Offer and the credential instance `c_identifier` introduced in PR #65. I think in general they are separate, but I can think of instances where it is beneficial for the wallet to communicate `co_identifier` in the credential request when the AS does not support `c_identifiers`. 
- relationship between identifiers, scope, and type values. I think in general 1:1:1 mapping between them cannot be assumed and it is also up to the Issuer to decide the mapping between scope and identifier/type.

Sakurann:
agreed at pre-IIW DCP WG:

- no inline metadata
- identifier in the offer is an identifier - it is the used to reference either an entry in the issuer metadata, or a separate private file if pre-agreed btw the wallet and the issuer (because there is no obligation in OAuth/OIDC to publish all the credentials supported in the metadata, can be in a private namespace)
- Credential type identifier vs credential instance identifier are different things
- no credential instance identifier in the credential offer; because wallet does not need information on instance identifiers for user consent at that stage of the protocol
- current PR 65 is sufficient for credential instance identifiers - returning it from the token endpoint.

https://docs.google.com/document/d/1q_vcI7PGJLtM3j0jvpLfmiBsj69jCilTzDQCjKRT2xY/edit

Sakurann:
PR merged

Issue #82, opened 2023-10-02
Should it be possible to have offers that are usable by a wallet without resorting to information in the metadata's `credentials_supported`


pmhsfelix:
I'm creating this issue based on a question that we had on the 2023-09-29 DCP meeting (https://lists.openid.net/pipermail/openid-specs-ab/2023-September/010092.html).

Should it be possible to have offers that do *not* require any information from the metadata's `credentials_supported`, i.e., offers that contain *all* the information required for the wallet to do an authorization request + token request + credential request *without* needing to map the offer into one of the metadata's `credential_supported` entry?

The use case would be to create one-of/dynamic offers that don't map into any of the statically defined metadata entries.

This question is similar/identical to the one in https://github.com/openid/OpenID4VCI/issues/77, however the discussion there seems to have gone to a slightly different subject, which is how to map and offer into a `credential_supported` entry.

Notes:
- The issuer metadata would still be needed to discover the credential endpoint and the authorization server ID. However no data from the `credentials_supported` array would be needed by a wallet to obtain a credential from an offer.
- In order for an offer to be usable without anything from the metadata's `supported_credential`, the offer would need to include fields such as `cryptographic_binding_methods_supported`, `cryptographic_suites_supported`, `display`, as well as format specific fields such as `credentialSubject` and `order`. For instance, this would allow for dynamic `credentialSubject` objects tailored to each specific offer.

If the answer is to this issue question is yes, then we probably have other things to discuss however that probably should be the subject of a different issue.

Sakurann:
> The use case would be to create one-of/dynamic offers that don't map into any of the statically defined metadata entries.

Why is this necessary? what are your concerns? is it the ability to dynamically generate metadata entries? Or is it the wallet making an error when trying to obtain a correct `crednetials_supported` object after fetching an entire issuer metadata. 

I think both could be mitigated by for example putting `identifier` in the offer instead of `scope` and enabling an option of the wallet passing `identifier` as a query `?c_identifier=xxx` when fetching the issuer metadata so that the issuer can return only relevant parts.

It might be ok to have `inline metadata` as an option, but it does not make sense to me to pass metadata entry as a value that is also being hosted. Do you agree with the proposal I made in two bullet points in https://github.com/openid/OpenID4VCI/issues/77#issuecomment-1741435566 ? (I do think we should continue this discussion in #77...)

Sakurann:
agreed at pre-IIW DCP WG:

- no inline metadata
[...]

Issue #77, opened 2023-09-25
Inline credential offer lacks required metadata


TimoGlastra:
When using inline credential offers (so using an object directly in the `credentials` array in an offer), the only required general property is the `format`. All other fields are declared by the format. 

The `credentials_supported` from the issuer metadata however has a lot more properties that are supported by default. Most notably: `cryptographic_binding_methods_supported`, `cryptographic_suites_supported` and `display`. 

I think this makes it very hard to use inline credential offers, as you have to guess which binding methods and cryptosuites the issuer supports. 

How I'd first understood the different between referenced offers (with id) and inline is that inline offers may be used for one-off credentials, or credentials you don't want displayed in the public issuer metadata. 

But there's a certain limitation to using the inline format, which I don't necessarily understand.

So my questions are:
- Is there a specific reason why the format of the `credentials_supported` and the inline credential offer format aren't aligned? Also for the formats the credentials_supported has more metadata (such as `credentialSubject`, `order` etc.. This seems to be useful to have with inline offers as well
- Would there be any objection to aligning the properties between credentials_supported and inline offers more? Without at least the `cryptographic_binding_methods_supported` and `cryptographic_suites_supported` properties, the inline offers seem hard to use to me. But even the `display` property would be nice to have in an inline offer. I can understand the display is a bit more risky, as someone could change the display properties somehow, while if you fetch it from a domain you're sure the properties are endorsed by the issuer. But if using a `credential_offer_uri` this is also migitated.

nklomp:
What you see is that in a lot of cases the credential offer types and format are mapped onto the supported metadata (and the examples also imply that). We also do that, but IMO it should indeed be the one or the other. Or you provide an id and map that onto credentials supported (scopes in latest dev draft version), or you provide inline objects, but then allow for everything in there. The problem with current approach is you cannot map for instance two OpenBadge Credentials with the same format. Let's say I want to offer a slightly different version of a credential, but with the same type and format. Obviously I can have multiple definitions in the metadata and assign IDs/scopes to them. That also works with using ID/scopes in the offer. It would fail with the examples and how both we and Mattr have implemented it, as the types and format will never definitively identify which credential we would be talking about.


IMO scopes should just be made mandatory in the definition. Use scope strings in the offer, or have an inline credential definition, but then allow for everything in there (which means you will have to resort to a URI because of the size)

TimoGlastra:
[...]

PR #86, opened 2023-10-16
`credentials` Credentials Offer parameter values identifying `credentials_supported` object(s) in the issuer's metadata.

Sakurann:
As discussed during IIW week, changing the structure of credentials parameter in the credential offer back to the identifier that can be used to obtain specific `credentials_supported` object from the issuer metadata. Also removing an option to pass format/type in the credential offer. Also agreed that inline metadata in the credential offer will not be supported. Addresses issues #83, #82, #77, #53.

During pre-IIW DCP WG mtg, we discussed enabling `credentials_supported` object being hosted not necessarily at `.well-known/openid-credential-issuer` as currently defined in the specification but at the private location that is known privately between the wallet and the Issuer. Current text `A JSON array of JSON strings that each uniquely identify one of the objects in the `credentials_supported` array included in the Credential Issuer metadata as defined in (#credential-issuer-parameters).` should be open enough to allow this.

There are separate issues like #42 on how the wallet could potentially add this identifier as query to the issuer metadata location so that the issuer returns part of the maya data instead of the wallet needing to search potentially large metadata file, but that is probably a separate PR.


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1173,6 +1173,7 @@ The following parameter MUST be used to communicate the specifics of the Credent
 This section defines the structure of the objects that appear in the `credentials_supported` metadata parameter.
 
 * `format`: REQUIRED. A JSON string identifying the format of this credential, i.e., `jwt_vc_json` or `ldp_vc`. Depending on the format value, the object contains further elements defining the type and (optionally) particular claims the credential MAY contain and information about how to display the credential. (#format_profiles) defines Credential Format Profiles introduced by this specification.
+* `credentials_supported_identifier`: REQUIRED when Credential Offer is used. A JSON string uniquely identifying this object within the `credentials_supported` array. It is used in the Credential Offer as defined in (#credential_offer_parameters) to enable the Wallet to discover the information about the Credential being offered.
```
cobward, line 1176: Since this is a unique identifier could we consider renaming this to `id` and moving it above `format` in this definition to signify the importance?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -314,7 +314,7 @@ For security considerations, see (#credential-offer-security).
...
-* `credentials`: REQUIRED. A JSON array, where every entry is a JSON object or a JSON string. If the entry is an object, the object contains the data related to a certain credential type the Wallet MAY request. Each object MUST contain a `format` Claim determining the format of the credential to be requested and further parameters characterizing the type of the credential to be requested as defined in (#format_profiles). If the entry is a string, the string value MUST be one of the `scope` values in one of the objects in the `credentials_supported` Credential Issuer metadata parameter. When processing, the Wallet MUST resolve this string value to the respective object.
+* `credentials`: REQUIRED. A JSON array of JSON strings that each uniquely identify one of the objects in the `credentials_supported` array included in the Credential Issuer metadata as defined in (#credential-issuer-parameters). The Wallet resolves this string value to the respective object to obtain information about the Credential being offered defined in (credential-metadata-object). For example, this string value can be used to obtain `scope` value to be used in the Authorization Request.
```
cobward, line 317: Could this be reworded to make it clear which element in the `credentials_supported` object it is supposed to match against? I also think it is redundant to say `uniquely identify` here, as the definition of the identifier already includes that.

Maybe something like:
```suggestion
* `credentials`: REQUIRED. A JSON array of JSON strings that each can be matched against the `credentials_supported_identifier` parameter of the objects in the `credentials_supported` array included in the Credential Issuer metadata as defined in (#credential-issuer-parameters). The Wallet resolves this string value to the respective object to identify and obtain information about the Credential being offered defined in (credential-metadata-object). For example, this string value can be used to obtain `scope` value to be used in the Authorization Request.
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: Within VCI, there are two different identifiers: 1) identifier that points to a credentials_supported object in the issuer metadata, and 2) identifier that identifies a credential instance during issuance (PR #65), so I was trying to differentiate those.

Having said that, there is only one identifier within issuer metadata, so maybe just "id" is ok...

There are so many "id" claims in the Credential space: in the DIF presentation exchange, in W3C VCDM, that i would prefer differentiating from those, but maybe no need...

On the order, I usually like putting mandatory parameters at the top, but in this case, logically, identifier should probably be at the top.

danielfett:
IMO, this proposal can be greatly simplified by changing the type of `credentials_supported` from an array to an object. The object keys would be what is now `credentials_supported_identifier`. This would also avoid having to name this identifier.

Example for `credentials_supported`:

```json
{
   "UniversityDegreeCredential":{
      "format":"jwt_vc_json",
      "cryptographic_binding_methods_supported":[
         
      ]
   },
   "Example2":{
      "format":"mso_mdoc",
      "doctype":"org.iso.18013.5.1.mDL"
   }
}
```

In the credential offer would then contain `UniversityDegreeCredential` or `Example2` as required.

paulbastian:
@danielfett and I jsut realized the potential of this PR is even bigger. We could simplify Token Request and Credential Request as well and get rid off alomst all credential-format specific things

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* `credentials`: REQUIRED. An array of strings that each can be matched against the `credentials_supported_identifier` parameter of the objects in the `credentials_supported` array included in the Credential Issuer metadata as defined in (#credential-issuer-parameters). The Wallet resolves this string value to the respective object to identify and obtain information about the Credential being offered defined in (credential-metadata-object). For example, this string value can be used to obtain `scope` value to be used in the Authorization Request.
```

Sakurann:
> @danielfett and I jsut realized the potential of this PR is even bigger. We could simplify Token Request and Credential Request as well and get rid off alomst all credential-format specific things

@paulbastian that is already addressed in PR #65. it is out of scope of this PR because two identifiers need to be clearly distinguished: this PR talks only about the identifier used to match against an object within issuer metadata, while PR #65 talks about an identifier of a credential instance that is being issued. 

PR #65 allows to use `c_instance_identifier` instead of format specific set of parameters in the credential request, which could greatly simplify the request when authorization_details are used. there is no consensus whether instance identifiers should be enabled when scopes are used instead of RAR, because the mapping of an instance identifier to a certain scope value can be quite complex.

To prevent further confusion, summarizing current E2E flows we will have in the spec if both this and PR #65 get merged.

1. Credential Offer with scopes
    a. wallet gets credential offer, uses credentials_supported_identifier to obtain a relevant part of a metadata and a scope value associated with it.
    b. wallet uses scope and format specific parameters in authz~credential requests, no credential_instance_identifiers

2. Credential offer with authorization details 
    a. wallet gets credential offer, uses credentials_supported_identifier to obtain a relevant part of a metadata
    b. wallet uses authorization details in authz~token requests and can simplify credential request by using only credential_instance_identifiers if AS/RS support it.

(steps 1a and 1b can obviously be used stand-alone without 1a and 2a.)

so in short, I think there is no further simplification needed in this PR.

Sakurann:
> IMO, this proposal can be greatly simplified by changing the type of credentials_supported from an array to an object. The object keys would be what is now credentials_supported_identifier. This would also avoid having to name this identifier.

I agree changing to a map looks cleaner, but I don't have a strong opinion - would like feedback from the WG.

Sakurann:
@TimoGlastra @nklomp can you please review if you get a chance..?

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: I agree with @cobward , it is the primary identifier and belongs to the top. I was also wondering why its not named `id` but I'm fine with the explanation

paulbastian (approved):
This is a great improvement.

It seems to me that there is some redundancy between `credentials_supported_identifier` ,  `scope` , `c_instance_identifier` and `authorization_details` but that become clearer once both PRs are merged.

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
Sakurann: fixed

Sakurann: the structure has changed to map, please check an updated language..

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: with `credentials_subject` structure changing to a map, there will be no more `credentials_supported_identifier` :)
@cobward I think this data structure change of a `credentials_supported_identifier` is a breaking change for your implementation.. please check.

Sakurann:
as part of changing the data structure to a map, I also merged two sub-sections within the issuer metadata section. so there is no more section dedicated to explaining parameters comprising `credentials_supported` object.

also, originally, `credentials_supported_identifier` was mandatory only when credential offer was used, but now it is always mandatory. so... what if the flow does not start with the credential offer? the wallet can obtain object values even without knowing the key value, but thought it is worth pointing out.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1159,41 +1159,32 @@ This specification defines the following Credential Issuer Metadata:
...
-    * `description`: OPTIONAL. String value of a description of the Credential.
-    * `background_color`: OPTIONAL. String value of a background color of the Credential represented as numerical color values defined in CSS Color Module Level 37 [@!CSS-Color].
-    * `text_color`: OPTIONAL. String value of a text color of the Credential represented as numerical color values defined in CSS Color Module Level 37 [@!CSS-Color].
-
-It is dependent on the Credential format where the available claims will appear and how they are represented (see (#format_profiles)).
+* `credentials_supported`: REQUIRED. An object that describes specifics of the Credential that the Credential Issuer supports issuance of. This object contains a list of name/value pairs, where each name is an identifier of the credential being described. This identifier is used in the Credential Offer as defined in (#credential_offer_parameters) to communicate to the Wallet which Credential is being offered. The value is an object that contains metadata about specific credential and contains the following parameters defined by this specification:
```
paulbastian, line 1165: If we have a breaking change here, it is a convenient opportunity for any renaming from #91 . This would also enable some backwards-compatiblity

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -446,7 +446,7 @@ Note: Applications MAY combine authorization details of type `openid_credential`
...
 
-When the Wallet does not know which scope value to use to request issuance of a certain credential, it can discover it using the `scope` Credential Issuer metadata parameter defined in (#credential-metadata-object). When the flow starts with a Credential Offer, the Wallet can use the information in the `credentials` Credential Offer parameter. When the entry of the `credentials` parameter is a string, it can be used as a scope value. When the entry of the `credentials` parameter is an object, it can be used to discover the desired scope value from the Credential Issuer metadata that can be obtained using `credential_issuer` Credential Offer parameter as defined in (#credential-issuer-wellknown).
+When the Wallet does not know which scope value to use to request issuance of a certain credential, it can discover it using the `scope` Credential Issuer metadata parameter defined in (#credential-issuer-parameters). When the flow starts with a Credential Offer, the Wallet can use the `credentials` parameter values to identify object(s) in the `credentials_supported` array in the Credential Issuer metadata parameter and use `scope` parameter value from that object.
```
TimoGlastra, line 449: credentials_supported structure has been changed to an object/map

```suggestion
When the Wallet does not know which scope value to use to request issuance of a certain credential, it can discover it using the `scope` Credential Issuer metadata parameter defined in (#credential-issuer-parameters). When the flow starts with a Credential Offer, the Wallet can use the `credentials` parameter values to identify object(s) in the `credentials_supported` object in the Credential Issuer metadata parameter and use `scope` parameter value from that object.
```

```suggestion
When the Wallet does not know which scope value to use to request issuance of a certain credential, it can discover it using the `scope` Credential Issuer metadata parameter defined in (#credential-issuer-parameters). When the flow starts with a Credential Offer, the Wallet can use the `credentials` parameter values to identify object(s) in the `credentials_supported` map in the Credential Issuer metadata parameter and use `scope` parameter value from that object.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1919,6 +1910,11 @@ The value of the `credential` claim in the Credential Response MUST be a JSON st
 
    [[ To be removed from the final specification ]]
    
+   -19
+  
+   * changed the structure of the `credentials_supported` parameter to a map from array of objects
+   * changed the structure of `credentials` parameter in Credential Offer to only be a string (no more objects) whose value is a key in the `credentials_supported` list entry
```
TimoGlastra, line 1916:
```suggestion
   * changed the structure of `credentials` parameter in Credential Offer to only be a string (no more objects) whose value is a key in the `credentials_supported` map
```

TimoGlastra (approved):
These changes look good to me 👍

danielfett (approved):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: good catch - thank you!

openid-4-verifiable-credential-issuance-1_0.md
nklomp, line 1165:
```suggestion
* `credentials_supported`: REQUIRED. An object that describes specifics of the Credential that the Credential Issuer supports issuance of. This object contains a list of name/value pairs, where each name is a unique identifier of the credential being described. This identifier is used in the Credential Offer as defined in (#credential_offer_parameters) to communicate to the Wallet which Credential is being offered. The value is an object that contains metadata about specific credential and contains the following parameters defined by this specification:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -446,7 +446,7 @@ Note: Applications MAY combine authorization details of type `openid_credential`
...
 
-When the Wallet does not know which scope value to use to request issuance of a certain credential, it can discover it using the `scope` Credential Issuer metadata parameter defined in (#credential-metadata-object). When the flow starts with a Credential Offer, the Wallet can use the information in the `credentials` Credential Offer parameter. When the entry of the `credentials` parameter is a string, it can be used as a scope value. When the entry of the `credentials` parameter is an object, it can be used to discover the desired scope value from the Credential Issuer metadata that can be obtained using `credential_issuer` Credential Offer parameter as defined in (#credential-issuer-wellknown).
+When the Wallet does not know which scope value to use to request issuance of a certain credential, it can discover it using the `scope` Credential Issuer metadata parameter defined in (#credential-issuer-parameters). When the flow starts with a Credential Offer, the Wallet can use the `credentials` parameter values to identify object(s) in the `credentials_supported` map in the Credential Issuer metadata parameter and use `scope` parameter value from that object.
```
nklomp, line 449: Previously it allowed for the param to either being a string, which then needed to be used as a scope, or it was the object, from which you then had to lookup the scope in the respective `credentials_supported` object.

scopes are however completely optional in `credentials_supported`. Since the string will now always map onto the respective key of the `credentials_supported` map, this change means we now always need to look at the `scrope` value. Is that desired, or should the text be reworked that you need to lookup the scope and if not present use the string value?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -314,7 +314,7 @@ For security considerations, see (#credential-offer-security).
...
-* `credentials`: REQUIRED. A JSON array, where every entry is a JSON object or a JSON string. If the entry is an object, the object contains the data related to a certain credential type the Wallet MAY request. Each object MUST contain a `format` Claim determining the format of the credential to be requested and further parameters characterizing the type of the credential to be requested as defined in (#format_profiles). If the entry is a string, the string value MUST be one of the `scope` values in one of the objects in the `credentials_supported` Credential Issuer metadata parameter. When processing, the Wallet MUST resolve this string value to the respective object.
+* `credentials`: REQUIRED. An array of strings that each identify one of the keys in the key/value pairs stored in the `credentials_supported` Credential Issuer metadata property. The Wallet uses this string value to obtain the respective object that contains information about the Credential being offered as defined in (#credential-issuer-parameters). For example, this string value can be used to obtain `scope` value to be used in the Authorization Request.
```
nklomp, line 317:
```suggestion
* `credentials`: REQUIRED. An array of unique strings that each identify one of the keys in the key/value pairs stored in the `credentials_supported` Credential Issuer metadata property. The Wallet uses this string value to obtain the respective object that contains information about the Credential being offered as defined in (#credential-issuer-parameters). For example, this string value can be used to obtain `scope` value to be used in the Authorization Request.
```

nklomp (approved):
Overall liking the change. Have some small comments to make it more clear the keys need to be unique and a remark about the scope, for which the lookup now has a bit of different behaviour.

But these are non-blocking

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: when saying unique, I think we need to be a little more specific..? I think we would want `unique within that issuer, but not globally unique` ?

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: I think it is a desired/intended behavior that if the wallet needs to look up the scope value, it uses the identifier to look it up. if it does not need scope value, it just uses the parameters other than scope values to formulate authorization_details request. allowing to use identifier as a scope value will lead to ambiguities when that is allowed and when it is not..

examples/credential_issuer_metadata_jwt_vc_json.json
```diff
@@ -0,0 +1,76 @@
+{
+    "credential_issuer": "",
+    "authorization_server": "",
+    "credential_endpoint": "",
+    "batch_credential_endpoint": "",
+    "deferred_credential_endpoint": "",
```
Sakurann, line 6: add values

openid-4-verifiable-credential-issuance-1_0.md
pmhsfelix: I suggest replacing `key` with `name` in when refering to members of JSON objects. I.e. `name/value pairs` instead of `key/value` pairs. Both https://www.json.org and https://www.rfc-editor.org/rfc/rfc8259.html seem to use the `name` terminology.

openid-4-verifiable-credential-issuance-1_0.md
pmhsfelix, line 449: Suggest "`credentials_supported` object in" instead of "`credentials_supported` map in" (i.e. use `object` instead of `map`).

openid-4-verifiable-credential-issuance-1_0.md
pmhsfelix, line 1165: `identifier of the credential being described` may collide with the different `credential_identifier` that we added in a different PR.
Perhaps `identifier of the supported credential being described`.

pmhsfelix (approved):
Approved with minor comments.

openid-4-verifiable-credential-issuance-1_0.md
nklomp: @Sakurann Yeah, but that is more or less assumed by just adding unique. But always better to be very explicit about it

openid-4-verifiable-credential-issuance-1_0.md
nklomp: Agreed, just pointing out the difference with previous version

examples/credential_issuer_metadata_jwt_vc_json.json
Sakurann:
```suggestion
    "credential_issuer": "https://credential-issuer.example.com",
    "credential_endpoint": "https://server.example.com",
    "batch_credential_endpoint": "https://server.example.com/batch_credential",
    "deferred_credential_endpoint": "https://server.example.com/deferred_credential",
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* `credentials`: REQUIRED. An array of unique strings that each identify one of the keys in the name/value pairs stored in the `credentials_supported` Credential Issuer metadata property. The Wallet uses this string value to obtain the respective object that contains information about the Credential being offered as defined in (#credential-issuer-parameters). For example, this string value can be used to obtain `scope` value to be used in the Authorization Request.
```
agreed to change to `name/value pairs` and to keep `unique` (don't have a better text proposal to clarify)

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: it used to say object and Timo suggested to change to map and I agreed. happy to change back if this comes up again..

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* `credentials_supported`: REQUIRED. An object that describes specifics of the Credential that the Credential Issuer supports issuance of. This object contains a list of name/value pairs, where each name is a unique identifier of the supported credential being described. This identifier is used in the Credential Offer as defined in (#credential_offer_parameters) to communicate to the Wallet which Credential is being offered. The value is an object that contains metadata about specific credential and contains the following parameters defined by this specification:
```

PR #87, opened 2023-10-19
Issuer metadata to indicate PoP requirement.

cobward:
Addresses #89

Currently a wallet has no way of knowing whether the issuing authority requires a proof of possession in the Credential Request. From my point of view there are two ways we can remedy this:

1. Add a parameter in the `credentials_supported` issuer metadata object that indicates whether a PoP is required.
2. Add some normative text which would require wallets to send a PoP if `proof_types_supported` in the `credentials_supported` issuer metadata object is not omitted and contains at least one element.

Personally I prefer explicit indication over implied so I have raised this PR showing option 1, but I am open to other opinions and options.


paulbastian:
There is also the option, that the Wallet may know this implicit, depending on whether the Issuer sends a `c_nonce` in the Token Response. But I think putting it in the Issuer metadata is the right direction, as there may also be more requirements on the keys and other requirements such as wallet attestations that must be reflected in the metadata

peppelinux:
It makes sense.
In italy [1] we mandate DPoP in the token endpoint, then the `c_nonce` is provided in the token endpoint response.

A signal about this could be taken with the presence of `dpop_signing_alg_values_supported` in the AS metadata.

I believe that we need more guidance to say how and when the token endpoint requires PoP and how and when the access token MUST be DPoP and not bearer, and how we may communicate it as a required or optional for a good interop.

[1] https://italia.github.io/eudi-wallet-it-docs/versione-corrente/en/pid-eaa-issuance.html#token-request

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1177,6 +1177,7 @@ This section defines the structure of the objects that appear in the `credential
...
 * `proof_types_supported`: OPTIONAL. A JSON array of case sensitive strings, each representing `proof_type` that the Credential Issuer supports. Supported values include those defined in (#proof_types) or other values defined in a profile of this specification or elsewhere. If omitted, the default value is `jwt`. `proof_type` claim is defined in (#credential_request).
+* `proof_of_possession_required`: OPTIONAL. A JSON boolean that indicates whether a proof of possession must be submitted when requesting a credential of this format. If this field is omitted it can assumed to be `false`.
```
peppelinux, line 1180:
```suggestion
* `proof_of_possession_required`: OPTIONAL. A JSON boolean that indicates whether a proof of possession must be submitted when requesting a Credential of this format. If this field is omitted it can assumed to be `false`.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -689,7 +689,7 @@ For cryptographic binding, the Client has the following options to provide crypt
 A Client makes a Credential Request to the Credential Endpoint by sending the following parameters in the entity-body of an HTTP POST request using the `application/json` media type.
 
 * `format`: REQUIRED. Format of the Credential to be issued. This Credential format identifier determines further parameters required to determine the type and (optionally) the content of the credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles).
-* `proof`: OPTIONAL. JSON object containing proof of possession of the key material the issued Credential shall be bound to.  The `proof` object MUST contain a following claim:
+* `proof`: OPTIONAL. JSON object containing proof of possession of the key material the issued Credential shall be bound to.  The `proof` object is REQUIRED if the `proof_of_possession_required` parameter in the issuer metadata for the submitted credential format is `true`. The `proof` object MUST contain a following claim:
```
peppelinux, line 692:
```suggestion
* `proof`: OPTIONAL. JSON object containing proof of possession of the key material the issued Credential SHALL be bound to.  The `proof` object is REQUIRED if the `proof_of_possession_required` parameter in the issuer metadata for the submitted Credential format is `true`. The `proof` object MUST contain a following claim:
```

Sakurann: Jacob, please pull the most recent version of the spec. the text has been updated to `proof: OPTIONAL. Object containing the proof of possession of the cryptographic key material the issued Credential would be bound to.`

Giuseppe, this specification uses MUST, not shall for normative statements.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* `proof_of_possession_required`: OPTIONAL. Boolean value specifying whether Credential Issuer requires a proof of possession to be submitted in the Credential Request, with `true` indicating support. If omitted, the default value is `true`.
```
aligning with the rest of the language. also, why is default `false`..?

openid-4-verifiable-credential-issuance-1_0.md
danielfett: Since we decided in #97 to drop the "JSON" in most places, please remove it here as well.

openid-4-verifiable-credential-issuance-1_0.md
danielfett: Since we decided in #97 to drop the "JSON" in most places, please remove it here as well.

openid-4-verifiable-credential-issuance-1_0.md
cobward: Done.

openid-4-verifiable-credential-issuance-1_0.md
cobward: A boolean default is usually false, so this is the most natural to me. Also it makes sense that the issuer would need to include this field to require PoP, rather than the issuer needing to include this field to allow for no PoP requirement.

Sakurann (commented):
discussed in a WG call. agreed that there is a need to clarify when PoP is needed but with an improvement. Specific suggestion is as following:
there are currently two parameters: proof_types_supported and proof_of_possession_required. the presence of proof_types_supported alone can be used to indicate that PoP is required because for the same credential type, PoP is either required or not (no use-case when same type is sometimes issued without PoP and other times with PoP), but `_supported` has a different nuance from `_required` so suggetion is to rename proof_types_supported to `proof_types` and clarify that when it is omitted, it means "PoP not required".

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1194,6 +1194,7 @@ This specification defines the following Credential Issuer Metadata:
...
   * `proof_types_supported`: OPTIONAL. A JSON array of case sensitive strings, each representing `proof_type` that the Credential Issuer supports. Supported values include those defined in (#proof_types) or other values defined in a profile of this specification or elsewhere. If omitted, the default value is `jwt`. `proof_type` claim is defined in (#credential_request).
+  * `proof_of_possession_required`: OPTIONAL. Boolean value specifying whether Credential Issuer requires a proof of possession to be submitted in the Credential Request, with `true` indicating support. If this field is omitted it can assumed to be `false`.
```
Sakurann, line 1197: merge into one parameter `proof_types`, where if omitted, the default value is PoP not required.
```suggestion
  * `proof_types`: REQUIRED when Credential Issuer requires a proof of possession of the cryptographic key material to be submitted in the Credential Request. Array of case sensitive strings, each representing `proof_type` that the Credential Issuer supports as defined in (#credential_request). Supported values include those defined in (#proof_types) or other values defined in a profile of this specification or elsewhere. If omitted, Credential Issuer does not require proof of possession of the cryptographic key material.
```
something like this?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,7 +710,7 @@ For cryptographic binding, the Client has the following options to provide crypt
...
 
 * `format`: REQUIRED when the `credential_identifier` was not returned from the Token Response. MUST NOT be used otherwise. String that determines the format of the Credential to be issued, which may determine the type and any other information related to the Credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles). When this parameter is used, `credential_identifier` parameter MUST NOT be present.
-* `proof`: OPTIONAL. Object containing the proof of possession of the cryptographic key material the issued Credential would be bound to.  The `proof` object MUST contain a following claim:
+* `proof`: OPTIONAL. Object containing the proof of possession of the cryptographic key material the issued Credential would be bound to.  The `proof` object is REQUIRED if the `proof_of_possession_required` parameter in the issuer metadata for the submitted credential format is `true`. The `proof` object MUST contain a following claim:
```
```suggestion
* `proof`: OPTIONAL. Object containing the proof of possession of the cryptographic key material the issued Credential would be bound to.  The `proof` object is REQUIRED if the `proof_types` parameter is present in the issuer metadata. The `proof` object MUST contain a following claim:
```
aligning with below

openid-4-verifiable-credential-issuance-1_0.md
jogu: @Sakurann your suggestions looks good to me.

cobward:
@Sakurann I've made some changes based on your suggestions, but I tried to streamline the language a little.

jogu (approved):

tlodderstedt (changes requested):

cobward:
This is how the text currently reads:

> Array of case sensitive strings, each representing a `proof_type` that the Credential Issuer **_supports_** as defined in (#credential_request), one of which MUST be used in the Credential Request. If this array is non-empty and present, the Credential Issuer requires proof of possession of the cryptographic key material. If the parameter is omitted or array is empty, Credential Issuer does not require proof of possession of the cryptographic key material.

I didn't quite understand your comment Torsten. Do you want the highlighted word to be changed from `supports` to `requires`?

Sakurann:
If you accepted my suggestion, @tlodderstedt agreed during the WG call that the language suggested addressed his concerns.

I think a sentence quoted in Torsten's comment is not the one he meant - and he was looking to clarify if the wallet need to support all or at least one of the proof types supported by the issuer.

cobward:
> If you accepted my suggestion, @tlodderstedt agreed during the WG call that the language suggested addressed his concerns.
> 
> I think a sentence quoted in Torsten's comment is not the one he meant - and he was looking to clarify if the wallet need to support all or at least one of the proof types supported by the issuer.
> 
> so this one should be good to go - thank you, Jacob!

Great, thanks for clearing that up Kristina

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1196,7 +1196,7 @@ This specification defines the following Credential Issuer Metadata:
...
-  * `proof_types_supported`: OPTIONAL. A JSON array of case sensitive strings, each representing `proof_type` that the Credential Issuer supports. Supported values include those defined in (#proof_types) or other values defined in a profile of this specification or elsewhere. If omitted, the default value is `jwt`. `proof_type` claim is defined in (#credential_request).
+  * `proof_types`: OPTIONAL. Array of case sensitive strings, each representing a `proof_type` that the Credential Issuer supports as defined in (#credential_request), one of which MUST be used in the Credential Request. If this array is non-empty and present, the Credential Issuer requires proof of possession of the cryptographic key material. If the parameter is omitted or array is empty, Credential Issuer does not require proof of possession of the cryptographic key material.
```
tplooker, line 1199:
```suggestion
  * `proof_types`: OPTIONAL. Array of case sensitive strings, each representing a `proof_type` that the Credential Issuer supports as defined in (#credential_request), one of which MUST be used in the Credential Request. If this array is non-empty and present, the Credential Issuer requires proof of possession of the cryptographic key material. If the parameter is omitted or the array is empty, the Credential Issuer does not require proof of possession of the cryptographic key material.
```

tplooker (approved):
Approved with one minor editorial nit

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -713,7 +713,7 @@ For cryptographic binding, the Client has the following options to provide crypt
...
 
 * `format`: REQUIRED when the `credential_identifier` was not returned from the Token Response. MUST NOT be used otherwise. String that determines the format of the Credential to be issued, which may determine the type and any other information related to the Credential to be issued. Credential Format Profiles consisting of the Credential format specific set of parameters are defined in (#format_profiles). When this parameter is used, `credential_identifier` parameter MUST NOT be present.
-* `proof`: OPTIONAL. Object containing the proof of possession of the cryptographic key material the issued Credential would be bound to.  The `proof` object MUST contain a following claim:
+* `proof`: OPTIONAL. Object containing the proof of possession of the cryptographic key material the issued Credential would be bound to.  The `proof` object is REQUIRED if the `proof_types` parameter is non-empty and present in the `credentials_supported` map of the issuer metadata for the requested credential. The `proof` object MUST contain a following claim:
```
tlodderstedt, line 716:
```suggestion
* `proof`: OPTIONAL. Object containing the proof of possession of the cryptographic key material the issued Credential would be bound to.  The `proof` object is REQUIRED if the `proof_types` parameter is present in the `credentials_supported` map of the issuer metadata for the requested credential and non-empty. The `proof` object MUST contain a following claim:
```

tlodderstedt (approved):

paulbastian (approved):

PR #95, opened 2023-11-01
add user_pin_length and user_pin_description to Credential Offer

### LINKED ISSUE #10: [OpenID4VCI] UserPIN description and length
### Original Reporter: pwlb

In Pre-authorized code flow the OpenID4VCI protocol offers the option to use a userPIN to further protect the code from theft/replay. The spec \(rightfully\) does not make any further assumptions about how and where this userPIN is communicated between Issuer and Holder. But this also means that situations might occur, where the Holder is prompted for the UserPIN in his wallet and he might be clueless of what the PIN is or where it might be communicated. The issuer might give this information on the website, where the QR-Code is displayed but

a\) in a same-device scenario the website is not visible anymore for the user

b\) in a cross-device scenario the user’s attention is now on the wallet \(we have seen problems in the past with DIDComm cross-device flows\)

UserPINs might be send as SMS Code or an email might not be aware to the user in this situation.

I therefore propose to extend the properties in the Credential Offer with either

1. add \`user\_pin\_description\` that is a short text describing how and where to get the userPIN OR
2. add \`user\_pin\_description\` in the metadata of the issuer

Additionally wallet implementers told us, they would favor a `user_pin_length` that says how many digits are expected for the userPIN such that this could be displayed in the app and might give further indications for the user what the correct PIN might be and improve the UI/UX in general.

\`user\_pin\_description\` might need i18n similar to what is already existing in the metadata, otherwise the issuer might know the language from the browser metadata.

‌

paulbastian:
Closes #10 
The proposal follows discussions in the issue and in the DCP workshop before iiw#37.


examples/credential_offer_multiple_credentials.json
```diff
@@ -12,8 +12,10 @@
          "issuer_state": "eyJhbGciOiJSU0Et...FYUaBy"
       },
       "urn:ietf:params:oauth:grant-type:pre-authorized_code": {
-         "pre-authorized_code": "adhjhdjajkdkhjhdj",
-         "user_pin_required": true
+         "pre-authorized_code": "oaKazRN8I0IbtZ0C7JuMn5",
+         "user_pin_required": true,
+         "user_pin_length": 4,
+         "user_pin_description": "Please provide the PIN, which was send via eMail"
```
danielfett, line 18:
```suggestion
         "user_pin_description": "Please provide the PIN which was sent via e-mail"
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -323,7 +323,9 @@ The following values are defined by this specification:
...
+  * `user_pin_length`: OPTIONAL. JSON Integer specifying the length of the PIN. This helps the Wallet to render the PIN screen and improve the user experience. When `user_pin_required` is `false`, this parameter MUST NOT be present.
+  * `user_pin_description`: OPTIONAL. JSON String containing guidance for the Holder of the Wallet how to obtain the PIN, e.g. specifying over which communication channel the PIN is delivered. The Wallet is RECOMMENDED to display this description next to the PIN entry form to improve the user experience. The length of the JSON String MUST be limited to 300 characters. The `user_pin_description` does not support internationalization (i18n), however the Issuer may detect the Holder's language by previous communication or an HTTP Accept-Language header within an HTTP GET request for a  `credential_offer_uri`. When `user_pin_required` is `false`, this parameter MUST NOT be present.
```
danielfett, line 328:
```suggestion
  * `user_pin_length`: OPTIONAL. Integer specifying the length of the PIN. This helps the Wallet to render the PIN screen and improve the user experience. When `user_pin_required` is `false`, this parameter MUST NOT be present.
  * `user_pin_description`: OPTIONAL. String containing guidance for the Holder of the Wallet how to obtain the PIN, e.g., specifying over which communication channel the PIN is delivered. The Wallet is RECOMMENDED to display this description next to the PIN entry form to improve the user experience. The length of the JSON String MUST be limited to 300 characters. The `user_pin_description` does not support internationalization (i18n), however the Issuer may detect the Holder's language by previous communication or an HTTP Accept-Language header within an HTTP GET request for a  `credential_offer_uri`. When `user_pin_required` is `false`, this parameter MUST NOT be present.
```

paulbastian: While I don't object to this, the specification is not very consistent on whether to name things "String" or "JSON String". Is there a consensus on this?

danielfett: My point of view is that it should be avoided to say "JSON String" (because the encoding is defined on the outer object and here we're just talking about the data structure).

jogu: Note that the 'JSON String' and 'JSON Integer' language are used throughout OpenID Connect and https://datatracker.ietf.org/doc/html/rfc7519 etc.

We should probably agree on the policy and use that policy consistently rather than changing specific instances in unrelated pull requests.

danielfett: I'll open an issue!

peppelinux (changes requested):

examples/credential_offer_multiple_credentials.json
```diff
@@ -12,8 +12,10 @@
          "issuer_state": "eyJhbGciOiJSU0Et...FYUaBy"
       },
       "urn:ietf:params:oauth:grant-type:pre-authorized_code": {
-         "pre-authorized_code": "adhjhdjajkdkhjhdj",
-         "user_pin_required": true
+         "pre-authorized_code": "oaKazRN8I0IbtZ0C7JuMn5",
```
pmhsfelix, line 15: Did you consider grouping these properties in an object?
~~~
"user_pin": {
    "required": true,
    "length": 4,
    "description": "Please provide the PIN which was sent via e-mail"
}
~~~
It may result in a small string and logically seems to make sense. We could even remove the `"required": true` which would be implicit if the object is present.
The downside is that it would be a breaking change.

pmhsfelix: While I agree that in most cases the locale can be determined from the context and be implicit, I'm afraid that we are making it impossible for an implementation to provide offers with more than one locale, in the eventual rare cases where it is needed.
Also, this seems to be the only place in this set of specs where we are providing localized user-facing strings without locale information. 
Note sure if saving some bytes justifies this exception.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -323,10 +323,12 @@ The following values are defined by this specification:
...
```
pmhsfelix, line 328: Detail: "JSON String containing guidance for the Holder of the Wallet *on* how to obtain the PIN".

examples/credential_offer_multiple_credentials.json
Sakurann: +1 to this structure. 
```
"user_pin": {
    "required": true,
    "length": 4,
    "description": "Please provide the PIN which was sent via e-mail"
}
```

examples/credential_offer_multiple_credentials.json
```diff
@@ -12,8 +12,10 @@
...
+         "user_pin_required": true,
+         "user_pin_length": 4,
+         "user_pin_description": "Please provide the PIN which was sent via e-mail"
```
paulbastian, line 18:
```suggestion
         "user_pin" : {
             "required": true,
             "length": 4,
             "description": "Please provide the PIN which was sent via e-mail"
         }
```

openid-4-verifiable-credential-issuance-1_0.md
danielfett: Since we agreed in #97 to drop "JSON", I have reopened this suggestion.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -323,10 +323,12 @@ The following values are defined by this specification:
...
   * `interval`: OPTIONAL. The minimum amount of time in seconds that the Wallet SHOULD wait between polling requests to the token endpoint (in case the Authorization Server responds with error code `authorization_pending` - see (#token_error_response)). If no value is provided, Wallets MUST use 5 as the default.
 
-The following non-normative example shows a Credential Offer object where the Credential Issuer can offer the issuance of two Credentials of different formats, one as JSON string ("UniversityDegree_JWT") and the other one as JSON object:
+The following non-normative example shows a Credential Offer object where the Credential Issuer offers the issuance of two Credentials of different formats, one as JSON string ("UniversityDegree_JWT") and the other one as JSON object:
```
danielfett, line 331:
```suggestion
The following non-normative example shows a Credential Offer object where the Credential Issuer offers the issuance of two Credentials of different formats, one as a string ("UniversityDegree_JWT") and the other one as an object:
```

examples/credential_offer_multiple_credentials.json
paulbastian: @pmhsfelix @Sakurann I applied the suggestion moving the 3 claims into one object.

paulbastian: @pmhsfelix The discussion on i18n already happened in #10 and in DCP workshop at iiw. I'm unsure if this is now the best place to open this discussion again. Otherwise please make a concrete proposal, because this PR was the consensus

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: applied

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: applied

peppelinux (approved):

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
paulbastian: applied

examples/credential_offer_multiple_credentials.json
paulbastian: If this is blocking for you please leave this in a dedicated review

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -560,7 +562,7 @@ Upon receiving a successful Authorization Response, a Token Request is made as d
 The following are the extension parameters to the Token Request used in a Pre-Authorized Code Flow defined by this specification:
 
 * `pre-authorized_code`: The code representing the authorization to obtain Credentials of a certain type. This parameter MUST be present if the `grant_type` is `urn:ietf:params:oauth:grant-type:pre-authorized_code`.
-* `user_pin`: OPTIONAL. String value containing a user PIN. This value MUST be present if `user_pin_required` was set to `true` in the Credential Offer. The string value MUST consist of a maximum of eight numeric characters (the numbers 0 - 9). This parameter MUST only be used if the `grant_type` is `urn:ietf:params:oauth:grant-type:pre-authorized_code`.
+* `user_pin`: OPTIONAL. String value containing a user PIN. This value MUST be present if `user_pin` object was present in the Credential Offer. The string value MUST consist of a maximum of eight numeric characters (the numbers 0 - 9). This parameter MUST only be used if the `grant_type` is `urn:ietf:params:oauth:grant-type:pre-authorized_code`.
```
danielfett, line 565:
```suggestion
* `user_pin`: OPTIONAL. String value containing a user PIN. This value MUST be present if a `user_pin` object was present in the Credential Offer (including if the object was empty). The string value MUST consist of a maximum of eight numeric characters (the numbers 0 - 9). This parameter MUST only be used if the `grant_type` is `urn:ietf:params:oauth:grant-type:pre-authorized_code`.
```

examples/credential_offer_multiple_credentials.json
pmhsfelix: Agree with the current design. I guess we can add a new internationalized field if that is needed in the future.

pmhsfelix (approved):

c2bo (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1937,6 +1939,7 @@ The value of the `credential` claim in the Credential Response MUST be a JSON st
    
    -19
   
+   * added `user_pin_description` and `user_pin_length` to Credential Offer
```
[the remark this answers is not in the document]
jogu:
```suggestion
   * replaced `user_pin_required` in Credential Offer with a `user_pin` object that also now contains `description` and `length`
```

Sakurann (changes requested):
Can we add pin type? It is needed to decide whether to show the alphanumeric or numeric keyboard to the user.

Also would like to see restructure changes suggested by Pedro

paulbastian:
> Can we add pin type? It is needed to decide whether to show the alphanumeric or numeric keyboard to the user.

Right now it is only allowed to have numeric PIN with 4-8 digits. I'm open to change these constraints.

paulbastian:
Also, we should think about incorporating the bike-shedding from #96 before merging this

paulbastian:
I applied changes from the bikeshedding discussion #96 .

Do we have consensus on adding the alphabet?
I propose something like:
```
 "urn:ietf:params:oauth:grant-type:pre-authorized_code": {
         "pre-authorized_code": "oaKazRN8I0IbtZ0C7JuMn5",
         "tx_code" : {
            "alphabet" : "numeric"
            "length": 4,
            "description": "Please provide the PIN which was sent via e-mail"
         }
      }
```
Adding `alphabet` under `tx_code` with possible values `numeric` and `alphanumeric`

jogu:
There seem to be about 20 instances of PIN still in the spec that I think at least some of still need to be changed?

jogu:
> Adding `alphabet` under `tx_code` with possible values `numeric` and `alphanumeric`

`alphabet` seems a bit awkward as a key name. Perhaps "inputmode" (as used in HTML) or "type"?

`alphanumeric` sounds like it excludes punctation and perhaps non-Latin characters. We might want to be explicit about exactly what is meant.

paulbastian:
> There seem to be about 20 instances of PIN still in the spec that I think at least some of still need to be changed?

solved

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -324,10 +324,12 @@ The following values are defined by this specification:
...
+  * `tx_code`: OPTIONAL. An object specifying whether the AS expects presentation of a Transaction Code by the End-User along with the Token Request in a Pre-Authorized Code Flow. If the AS does not expect a Transaction Code, this object is absent, this is the default. The Transaction Code is intended to bind the Pre-Authorized Code to a certain transaction to prevent replay of this code by an attacker that, for example, scanned the QR code while standing behind the legitimate End-User. It is RECOMMENDED to send the Transaction Code via a separate channel. If the Wallet decides to use the Pre-Authorized Code Flow, the Transaction Code value MUST be sent in the `tx_code` parameter with the respective Token Request. If no `length` or `description` is given, this object may be empty, indicating that a Transaction Code is required.
```
Sakurann, line 327:
```suggestion
  * `tx_code`: OPTIONAL. An object specifying whether the AS expects presentation of a Transaction Code by the End-User along with the Token Request in a Pre-Authorized Code Flow. If the AS does not expect a Transaction Code, this object is absent, this is the default. The Transaction Code is intended to bind the Pre-Authorized Code to a certain transaction to prevent replay of this code by an attacker that, for example, scanned the QR code while standing behind the legitimate End-User. It is RECOMMENDED to send the Transaction Code via a separate channel. If the Wallet decides to use the Pre-Authorized Code Flow, the Transaction Code value MUST be sent in the `tx_code` parameter with the respective Token Request as defined in (#token_request). If no `length` or `description` is given, this object may be empty, indicating that a Transaction Code is required.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -324,10 +324,12 @@ The following values are defined by this specification:
...
+    * `length`: OPTIONAL. Integer specifying the length of the Transaction Code. This helps the Wallet to render the input screen and improve the user experience.
+    * `description`: OPTIONAL. String containing guidance for the Holder of the Wallet on how to obtain the Transaction Code, e.g. describing over which communication channel it is delivered. The Wallet is RECOMMENDED to display this description next to the Transaction Code input screen to improve the user experience. The length of the String MUST NOT exceed 300 characters. The `description` does not support internationalization (i18n), however the Issuer MAY detect the Holder's language by previous communication or an HTTP Accept-Language header within an HTTP GET request for a  `credential_offer_uri`.
```
Sakurann, line 329:
```suggestion
    * `description`: OPTIONAL. String containing guidance for the Holder of the Wallet on how to obtain the Transaction Code, e.g. describing over which communication channel it is delivered. The Wallet is RECOMMENDED to display this description next to the Transaction Code input screen to improve the user experience. The length of the string MUST NOT exceed 300 characters. The `description` does not support internationalization (i18n), however the Issuer MAY detect the Holder's language by previous communication or an HTTP Accept-Language header within an HTTP GET request for a Credential Offer URI.
```

Sakurann:
> alphabet seems a bit awkward as a key name. Perhaps "inputmode" (as used in HTML)?

I like the suggestion. we can also use `text`, `decimal`, and `numeric` as defined [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode)

paulbastian:
added text on input_mode, please recheck @jogu @Sakurann

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -324,10 +324,13 @@ The following values are defined by this specification:
...
+  * `tx_code`: OPTIONAL. An object specifying whether the AS expects presentation of a Transaction Code by the End-User along with the Token Request in a Pre-Authorized Code Flow. If the AS does not expect a Transaction Code, this object is absent, this is the default. The Transaction Code is intended to bind the Pre-Authorized Code to a certain transaction to prevent replay of this code by an attacker that, for example, scanned the QR code while standing behind the legitimate End-User. It is RECOMMENDED to send the Transaction Code via a separate channel. If the Wallet decides to use the Pre-Authorized Code Flow, the Transaction Code value MUST be sent in the `tx_code` parameter with the respective Token Request as defined in (#token_request). If no `length` or `description` is given, this object may be empty, indicating that a Transaction Code is required.
+    * `input_mode` : OPTIONAL. String specifying the input characters set. Possible values are `numeric` (only digits) and `text` (any character). The default is `numeric`.
```
Sakurann, line 328: should this say that this input_mode can also control the keyboard being displayed to the user?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -324,10 +324,13 @@ The following values are defined by this specification:
...
+    * `length`: OPTIONAL. Integer specifying the length of the Transaction Code. This helps the Wallet to render the input screen and improve the user experience.
+    * `description`: OPTIONAL. String containing guidance for the Holder of the Wallet on how to obtain the Transaction Code, e.g. describing over which communication channel it is delivered. The Wallet is RECOMMENDED to display this description next to the Transaction Code input screen to improve the user experience. The length of the string MUST NOT exceed 300 characters. The `description` does not support internationalization (i18n), however the Issuer MAY detect the Holder's language by previous communication or an HTTP Accept-Language header within an HTTP GET request for a Credential Offedefiningr URI.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
    * `description`: OPTIONAL. String containing guidance for the Holder of the Wallet on how to obtain the Transaction Code, e.g. describing over which communication channel it is delivered. The Wallet is RECOMMENDED to display this description next to the Transaction Code input screen to improve the user experience. The length of the string MUST NOT exceed 300 characters. The `description` does not support internationalization (i18n), however the Issuer MAY detect the Holder's language by previous communication or an HTTP Accept-Language header within an HTTP GET request for a Credential Offer URI.
```

PR #116, opened 2023-11-17
support HTTP Accept-Language in the request for Credential Issuer Metadata to request a subset for display data

### LINKED ISSUE #84: Use HTTP Accept-Language Header to request Issuer Metadata
I propose to extend [10.2.2. ](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-10.2.2)[Credential Issuer Metadata Retrieval](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-credential-issuer-metadata-) with the recommendation for the Holder/Wallet to pass an HTTP Accept-Language Header according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language

This would save both Wallet and Issuer a lot of traffic, if the issuer supports many languages for the display properties. The Issuer may pre-generate language-specific issuer metadata files for individual languages. If the Wallet does not send an Accept-Language header, the Issuer may choose to send any language (e.g. english or all).

paulbastian:
Closes #84

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1172,16 +1172,16 @@ The Credential Issuer's configuration can be retrieved using the Credential Issu
 
 Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-credential-issuer` to the Credential Issuer Identifier. If the Credential Issuer value contains a path component, any terminating `/` MUST be removed before appending `/.well-known/openid-credential-issuer`.
 
-To fetch the Credential Issuer Metadata, a requester MUST send a HTTP GET to the path formed following the steps above utilizing TLS. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and a HTTP Status Code 200.
+To fetch the Credential Issuer Metadata, a requester MUST send a HTTP request using the GET method and the path formed following the steps above. The URL used for this purpose MUST use the https scheme and SHOULD NOT contain query parameter. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and a HTTP Status Code 200.
 
-The Wallet is RECOMMENDED to send an Accept-Language Header in the HTTP GET request to indicate the particular languages he is interested in. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested languages and using the HTTP Content-Language Header. The Credential Issuer MAY ignore the Accept-Language Header and send all supported languages or a any chosen default subset.
+The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the particular languages it is interested in. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested languages and using the HTTP `Content-Language` Header. The requested language MUST use the values defined in [RFC3066]. The Credential Issuer MAY ignore the `Accept-Language` Header and send all supported languages or a any chosen default subset.
 
 Below is a non-normative example of a Credential Issuer Metadata request:
 
 ```
 GET /.well-known/openid-credential-issuer HTTP/1.1
 Host: server.example.com
-Accept-Language: fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
+Accept-Language: fr-ch, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
 ```
 
 ### Credential Issuer Metadata Parameters {#credential-issuer-parameters}
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1160,6 +1160,8 @@ If the Credential Issuer is unable to perform discovery of the Wallet's Credenti
 
 ## Credential Issuer Metadata {#credential-issuer-metadata}
 
+The Credential Issuer Metadata contains information on the Credential Issuer's techical capabilities, supported credentials and (internationalized) display information.
+
 ### Credential Issuer Identifier {#credential-issuer-identifier}
 
 A Credential Issuer is identified by a case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components. 
@@ -1168,9 +1170,19 @@ A Credential Issuer is identified by a case sensitive URL using the `https` sche
 
 The Credential Issuer's configuration can be retrieved using the Credential Issuer Identifier.
 
-Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-credential-issuer` to the Credential Issuer Identifier. If the Credential Issuer value contains a path component, any terminating `/` MUST be removed before appending `/.well-known/openid-credential-issuer`. 
+Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-credential-issuer` to the Credential Issuer Identifier. If the Credential Issuer value contains a path component, any terminating `/` MUST be removed before appending `/.well-known/openid-credential-issuer`.
 
-The path formed following the steps above MUST point to a JSON document compliant with this specification. The document MUST be returned using the `application/json` media type.
+To fetch the Credential Issuer Metadata, a requester MUST send a HTTP GET to the path formed following the steps above utilizing TLS. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and a HTTP Status Code 200.
+
+The Wallet is RECOMMENDED to send an Accept-Language Header in the HTTP GET request to indicate the particular languages he is interested in. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested languages and using the HTTP Content-Language Header. The Credential Issuer MAY ignore the Accept-Language Header and send all supported languages or a any chosen default subset.
+
+Below is a non-normative example of a Batch Credential Request:
+
+```
+GET /.well-known/openid-credential-issuer HTTP/1.1
+Host: server.example.com
+Accept-Language: de-DE
+```
 
 ### Credential Issuer Metadata Parameters {#credential-issuer-parameters}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1176,7 +1176,12 @@ Communication with the Credential Issuer Metadata Endpoint MUST utilize TLS.
 
 To fetch the Credential Issuer Metadata, a requester MUST send a HTTP request using the GET method and the path formed following the steps above. The URL SHOULD NOT contain any query parameters. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and a HTTP Status Code 200.
 
-The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the particular language(s) it is interested in. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested language(s) and indicate the language(s) returned using the HTTP `Content-Language` Header. The requested language(s) MUST use the values defined in [@!RFC3066]. The Credential Issuer MAY ignore the `Accept-Language` Header and send all supported languages or any chosen default subset.
+The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the language(s) preferred for display. It is up to the Credential Issuer whether to
+
+* send a subset the metadata containing internationalized display data for one or all of the requested languages and indicate returned languages using the HTTP `Content-Language` Header.
+* ignore the `Accept-Language` Header and send all supported languages or any chosen default subset.
+
+The language(s) in HTTP `Accept-Language` and `Content-Language` Headers MUST use the values defined in [@!RFC3066]. 
 
 Below is a non-normative example of a Credential Issuer Metadata request:
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1160,7 +1160,7 @@ If the Credential Issuer is unable to perform discovery of the Wallet's Credenti
 
 ## Credential Issuer Metadata {#credential-issuer-metadata}
 
-The Credential Issuer Metadata contains information on the Credential Issuer's technical capabilities, supported credentials and (internationalized) display information.
+The Credential Issuer Metadata contains information on the Credential Issuer's technical capabilities, supported Credentials and (internationalized) display information.
 
 ### Credential Issuer Identifier {#credential-issuer-identifier}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1160,7 +1160,7 @@ If the Credential Issuer is unable to perform discovery of the Wallet's Credenti
 
 ## Credential Issuer Metadata {#credential-issuer-metadata}
 
-The Credential Issuer Metadata contains information on the Credential Issuer's techical capabilities, supported credentials and (internationalized) display information.
+The Credential Issuer Metadata contains information on the Credential Issuer's technical capabilities, supported credentials and (internationalized) display information.
 
 ### Credential Issuer Identifier {#credential-issuer-identifier}
 
@@ -1176,12 +1176,12 @@ To fetch the Credential Issuer Metadata, a requester MUST send a HTTP GET to the
 
 The Wallet is RECOMMENDED to send an Accept-Language Header in the HTTP GET request to indicate the particular languages he is interested in. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested languages and using the HTTP Content-Language Header. The Credential Issuer MAY ignore the Accept-Language Header and send all supported languages or a any chosen default subset.
 
-Below is a non-normative example of a Batch Credential Request:
+Below is a non-normative example of a Credential Issuer Metadata request:
 
 ```
 GET /.well-known/openid-credential-issuer HTTP/1.1
 Host: server.example.com
-Accept-Language: de-DE
+Accept-Language: fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
 ```
 
 ### Credential Issuer Metadata Parameters {#credential-issuer-parameters}
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1172,9 +1172,11 @@ The Credential Issuer's configuration can be retrieved using the Credential Issu
 
 Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-credential-issuer` to the Credential Issuer Identifier. If the Credential Issuer value contains a path component, any terminating `/` MUST be removed before appending `/.well-known/openid-credential-issuer`.
 
-To fetch the Credential Issuer Metadata, a requester MUST send a HTTP request using the GET method and the path formed following the steps above. The URL used for this purpose MUST use the https scheme and SHOULD NOT contain query parameter. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and a HTTP Status Code 200.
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1160,6 +1160,8 @@ If the Credential Issuer is unable to perform discovery of the Wallet's Credenti
 
 ## Credential Issuer Metadata {#credential-issuer-metadata}
 
+The Credential Issuer Metadata contains information on the Credential Issuer's techical capabilities, supported credentials and (internationalized) display information.
```
jogu, line 1163: typo: techical -> technical

```suggestion
The Credential Issuer Metadata contains information on the Credential Issuer's technical capabilities, supported credentials and (internationalized) display information.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1168,9 +1170,19 @@ A Credential Issuer is identified by a case sensitive URL using the `https` sche
...
+Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-credential-issuer` to the Credential Issuer Identifier. If the Credential Issuer value contains a path component, any terminating `/` MUST be removed before appending `/.well-known/openid-credential-issuer`.
+
+To fetch the Credential Issuer Metadata, a requester MUST send a HTTP GET to the path formed following the steps above utilizing TLS. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and a HTTP Status Code 200.
+
+The Wallet is RECOMMENDED to send an Accept-Language Header in the HTTP GET request to indicate the particular languages he is interested in. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested languages and using the HTTP Content-Language Header. The Credential Issuer MAY ignore the Accept-Language Header and send all supported languages or a any chosen default subset.
```
jogu, line 1177: Small improvements to wording:

```suggestion
The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the language(s) preferred for display. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested languages and using the HTTP `Content-Language` Header. The Credential Issuer MAY ignore the `Accept-Language` Header and send all supported languages or an any chosen default subset.
```

I'm not sure about the 'MAY' here. It seems like saying you MAY ignore the Accept-Language header conflicts with the previous sentence that says it's RECOMMENDED to process it.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1168,9 +1170,19 @@ A Credential Issuer is identified by a case sensitive URL using the `https` sche
...
+
+Below is a non-normative example of a Batch Credential Request:
 
-The path formed following the steps above MUST point to a JSON document compliant with this specification. The document MUST be returned using the `application/json` media type.
+```
+GET /.well-known/openid-credential-issuer HTTP/1.1
+Host: server.example.com
+Accept-Language: de-DE
```
jogu, line 1184: I believe it's permitted to have more complex values, so I think we may want an example that's more complex.

```suggestion
Accept-Language: fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1168,9 +1170,19 @@ A Credential Issuer is identified by a case sensitive URL using the `https` sche
...
+Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-credential-issuer` to the Credential Issuer Identifier. If the Credential Issuer value contains a path component, any terminating `/` MUST be removed before appending `/.well-known/openid-credential-issuer`.
+
+To fetch the Credential Issuer Metadata, a requester MUST send a HTTP GET to the path formed following the steps above utilizing TLS. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and a HTTP Status Code 200.
+
+The Wallet is RECOMMENDED to send an Accept-Language Header in the HTTP GET request to indicate the particular languages he is interested in. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested languages and using the HTTP Content-Language Header. The Credential Issuer MAY ignore the Accept-Language Header and send all supported languages or a any chosen default subset.
+
+Below is a non-normative example of a Batch Credential Request:
```
jogu, line 1179: That's just wrong :-)

```suggestion
Below is a non-normative example of a Credential Issuer Metadata request:
```

jogu (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
pmhsfelix: > and using the HTTP `Content-Language` Header
- The display entries already contain `locale` information, so the `Content-Language` header on the response is not strictly needed, IMO.
- Perhaps we should add that if the `Content-Language` header is present, it should be compatible with the locales used inside the metadata (`Content-Language` can have more than one language).

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1160,6 +1160,8 @@ If the Credential Issuer is unable to perform discovery of the Wallet's Credenti
 
 ## Credential Issuer Metadata {#credential-issuer-metadata}
 
+The Credential Issuer Metadata contains information on the Credential Issuer's technical capabilities, supported credentials and (internationalized) display information.
```
peppelinux, line 1163:
```suggestion
The Credential Issuer Metadata contains information on the Credential Issuer's technical capabilities, supported Credentials and (internationalized) display information.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1168,9 +1170,19 @@ A Credential Issuer is identified by a case sensitive URL using the `https` sche
 
 The Credential Issuer's configuration can be retrieved using the Credential Issuer Identifier.
 
-Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-credential-issuer` to the Credential Issuer Identifier. If the Credential Issuer value contains a path component, any terminating `/` MUST be removed before appending `/.well-known/openid-credential-issuer`. 
+Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-credential-issuer` to the Credential Issuer Identifier. If the Credential Issuer value contains a path component, any terminating `/` MUST be removed before appending `/.well-known/openid-credential-issuer`.
+
+To fetch the Credential Issuer Metadata, a requester MUST send a HTTP GET to the path formed following the steps above utilizing TLS. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and a HTTP Status Code 200.
```
peppelinux, line 1175:
```suggestion
To fetch the Credential Issuer Metadata, a requester MUST send an HTTP request using the GET method and the path formed following the steps above. The URL used for this purpose MUST use the https scheme and MUST NOT contain port, path or query parameter. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and a HTTP Status Code 200.
```

openid-4-verifiable-credential-issuance-1_0.md
peppelinux, line 1177:
```suggestion
The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the particular languages it is interested in. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested languages and using the HTTP `Content-Language` Header. The requested language MUST use the values defined in ISO 639 alpha-2.The Credential Issuer MAY ignore the `Accept-Language` Header and send all supported languages or a any chosen default subset.
```

peppelinux: Is there somethign else to say regarding ISO 639 alpha-2?
Is this correct?
Do we have to add a ref to that ISO specs?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1168,9 +1170,19 @@ A Credential Issuer is identified by a case sensitive URL using the `https` sche
...
+
+The Wallet is RECOMMENDED to send an Accept-Language Header in the HTTP GET request to indicate the particular languages he is interested in. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested languages and using the HTTP Content-Language Header. The Credential Issuer MAY ignore the Accept-Language Header and send all supported languages or a any chosen default subset.
+
+Below is a non-normative example of a Credential Issuer Metadata request:
 
-The path formed following the steps above MUST point to a JSON document compliant with this specification. The document MUST be returned using the `application/json` media type.
+```
+GET /.well-known/openid-credential-issuer HTTP/1.1
+Host: server.example.com
+Accept-Language: fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
```
peppelinux, line 1184:
```suggestion
Accept-Language: fr-ch, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
```

peppelinux: according to the codes defined in ISO 639-1

jogu: The language tags are case insensitive as per https://datatracker.ietf.org/doc/html/rfc3066#section-2.1

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: Why do we mandate not using port or path parameters in the URL? Infact, to OpenID rules, the /.well-known path might not be at the domain root but contain a path

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: @jogu Do you mean RECOMMENDED and MAY oppose each other? So then whats our opinion here? Issuers are RECOMMENDED OR MAY send the according subset?

paulbastian: @pmhsfelix thats true. But its also only RECOMMENDED. What do you propose?

jogu: > @jogu Do you mean RECOMMENDED and MAY oppose each other? 

Yes, exactly.

> So then whats our opinion here? Issuers are RECOMMENDED OR MAY send the according subset?

I'm not actually sure. I think the real position is that it's probably only worth the complexity of doing this if your metadata is pretty large, e.g. over 30 kilobytes with compression applied? (Or put another way, I'm not sure there's much point in doing this kind of optimisations until the data is notably larger than a single fairly typical photo thumbnail on a website) - so probably for many issuers it'd probably not be worth the effort. So I guess either "MAY" or perhaps "RECOMMENDED if the metadata size exceeds XXX"?

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: good catch @paulbastian

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: good catch @jogu, that's why I was asking if we're considering ISO 639-1

It seems to me that in your suggestion you brough RFC3066 as the ref to be included in this PR (I have updated my prev comments accordingly, thx)

> [ISO 3166] recommends that country codes are capitalized (MN Mongolia), while [ISO 639] recommends that language codes are written in lower case (mn Mongolian).

jogu: We're not considering anything :-) The behaviour of the Accept-language tag is already well defined ( https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.5 ) - we're just using it as it is already defined I believe.

peppelinux: @jogu I was asking since I didn't see RFC7231 in the normative or informative specs referenced in this specification

jogu: I think it is pretty clear that the specification already relies on HTTP 1.1 :-) There is an already a normative reference as a result of the normative reference to OAuth2 (and maybe others) that have normative references to HTTP RFCs. Similarly RFC3066 is a normative reference of HTTP RFC. (I don't know what the rule is on if we need our own normative reference as well.)

peppelinux: I learned something, thank you @jogu

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1168,9 +1170,19 @@ A Credential Issuer is identified by a case sensitive URL using the `https` sche
 
 The Credential Issuer's configuration can be retrieved using the Credential Issuer Identifier.
 
-Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-credential-issuer` to the Credential Issuer Identifier. If the Credential Issuer value contains a path component, any terminating `/` MUST be removed before appending `/.well-known/openid-credential-issuer`. 
+Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-credential-issuer` to the Credential Issuer Identifier. If the Credential Issuer value contains a path component, any terminating `/` MUST be removed before appending `/.well-known/openid-credential-issuer`.
+
```
paulbastian, line 1174:
```suggestion

Communication with the Credential Issuer Metadata Endpoint MUST utilize TLS.

```

paulbastian: @peppelinux this is the wording used throughout the spec

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1168,9 +1170,19 @@ A Credential Issuer is identified by a case sensitive URL using the `https` sche
 
 The Credential Issuer's configuration can be retrieved using the Credential Issuer Identifier.
 
-Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-credential-issuer` to the Credential Issuer Identifier. If the Credential Issuer value contains a path component, any terminating `/` MUST be removed before appending `/.well-known/openid-credential-issuer`. 
+Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string `/.well-known/openid-credential-issuer` to the Credential Issuer Identifier. If the Credential Issuer value contains a path component, any terminating `/` MUST be removed before appending `/.well-known/openid-credential-issuer`.
```
peppelinux, line 1173: I agree about appending the well-known path to enable multitenant implementations based on webpaths

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1168,9 +1170,19 @@ A Credential Issuer is identified by a case sensitive URL using the `https` sche
...
+
+To fetch the Credential Issuer Metadata, a requester MUST send a HTTP request using the GET method and the path formed following the steps above. The URL used for this purpose MUST use the https scheme and SHOULD NOT contain query parameter. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and a HTTP Status Code 200.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
To fetch the Credential Issuer Metadata, a requester MUST send an HTTP request using the GET method and the path formed following the steps above. The URL used for this purpose MUST use the `https` scheme and SHOULD NOT contain query parameter. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and a HTTP Status Code 200.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1168,9 +1170,19 @@ A Credential Issuer is identified by a case sensitive URL using the `https` sche
...
+
+The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the particular languages it is interested in. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested languages and using the HTTP `Content-Language` Header. The requested language MUST use the values defined in [RFC3066]. The Credential Issuer MAY ignore the `Accept-Language` Header and send all supported languages or a any chosen default subset.
```
Sakurann, line 1177:
```suggestion
The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the particular language(s) it is interested in. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested language(s) and indicate the language(s) returned using the HTTP `Content-Language` Header. The requested language(s) MUST use the values defined in [@!RFC3066]. The Credential Issuer MAY ignore the `Accept-Language` Header and send all supported languages or any chosen default subset.
```

Sakurann (commented):
approving assuming my proposed changes are accepted.

Sakurann (approved):

openid-4-verifiable-credential-issuance-1_0.md
c2bo: I agree that a more complex example than 1 language makes sense, but I am not sure the weighted example is a good fit here or rather might be confusing. Instead of using a weighted request, wouldn't we instead just provide the list of languages we are interested in and the response would include all the fitting languages supported by the issuer?

If I understand weighted accept-language usage correctly, it is supposed to indicate a favorite language and the response is the best fit the server could find (in one language).

paulbastian: I agree, we should make clear what the weighting means for us.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: agreed on the WG call that the spirit is `MAY`. @Sakurann to make a suggestion and @paulbastian @jogu to help bikeshed.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: discussed on the WG call. agreed that wallet can use weighted multiple languages, but hte issuer is free to return whatever, even if it is only one of the requested languages. @paulbastian to make sure this is reflected in the spec text. good to merge after that.

c2bo (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1168,9 +1170,21 @@ A Credential Issuer is identified by a case sensitive URL using the `https` sche
...
+
+Communication with the Credential Issuer Metadata Endpoint MUST utilize TLS.
+
+To fetch the Credential Issuer Metadata, a requester MUST send a HTTP request using the GET method and the path formed following the steps above. The URL SHOULD NOT contain any query parameters. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and a HTTP Status Code 200.
+
+The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the particular language(s) it is interested in. The Credential Issuer is RECOMMENDED to send a subset of internationalized display data containing the requested language(s) and indicate the language(s) returned using the HTTP `Content-Language` Header. The requested language(s) MUST use the values defined in [@!RFC3066]. The Credential Issuer MAY ignore the `Accept-Language` Header and send all supported languages or any chosen default subset.
```
Sakurann, line 1179:
```suggestion
The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the language(s) preferred for display. It is up to the Credential Issuer whether to

* send a subset the metadata containing internationalized display data for one or all of the requested languages and indicate returned languages using the HTTP `Content-Language` Header
* ignore the `Accept-Language` Header and send all supported languages or any chosen default subset.
```

c2bo: I like those changes! That is a lot clearer for implementers imho.

paulbastian: I like it as well. Shall we keep the RFC reference?

jogu: I'm good with the proposed language, thanks Kristina!

Sakurann: @paulbastian can you please commit suggestion if you agree and request re-review from Joseph so that we can merge this PR during tomorrow's call?

Sakurann: for the RFC reference, I think it applies to both Headers, so I added a sentence saying so?

jogu (approved):

PR #119, opened 2023-11-21
Enable publishing IACA certificates

### LINKED ISSUE #62: IACA Metadata for Credential Issuers
This issue proposes an extension to OpenID4VCI Credential Issuer Metadata to publish Issuing Authority Certificate Authority (IACA) certificates.  These certificates are defined in ISO 18013-5 (mDL).

A key decision is whether to provide the certificates by value in the Credential Issuer Metadata itself or whether to provide it by reference at a different URI.  I propose to provide it by reference so that the timing of updates can be controlled by normal Web Cache-Control directives.  (This is the same reason that `jwks_uri` is used in OpenID Connect OP metadata to provide keys by reference.)

I propose that IANA certificates be provided using this Credential Issuer Metadata element:

> `mdoc_iacas_uri`:  OPTIONAL. URL of the Credential Issuer’s IACA set document. This document contains the IACA certificates [ISO 18013-5] the used to validate signatures for issued mDOCs.

The referenced document has this structure:
> {"iacas":
  [
    {"certificate": "base64-encoded DER representation of IACA certificate"},
    …
  ]
}

The top-level structure is intentionally parallel to a JWK Set – which uses a `keys` element whose value is an array of JWKs.  The array elements are structures with a required `certificate` element, allowing other certificate metadata to be added to the structures, such as expiration time, as needed.

selfissued:
Fixes #62

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1918,7 +1918,7 @@ The value of the `credential` claim in the Credential Response MUST be a string 
 
 ### IACA Certificate Metadata
 
-This profile defines an OpenID4VCI Credential Issuer Metadata parameter to publish Issuing Authority Certificate Authority (IACA) certificates.  These certificates are defined in ISO 18013-5 (mDL).  They are needed to verify mdoc certificate chains.  This parameter is:
+This profile defines an OpenID4VCI Credential Issuer Metadata parameter to publish Issuing Authority Certificate Authority (IACA) certificates.  These X.509 certificates are defined in ISO 18013-5 (mDL). They are needed to verify mdoc X.509 certificate chains. This parameter is:
 
 * `mdoc_iacas_uri`: A URL using the `https` scheme referencing a resource containing a JSON-encoded object containing IACA certificates.
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1916,6 +1916,27 @@ The following is a non-normative example of a Credential Request with Credential
 
 The value of the `credential` claim in the Credential Response MUST be a string that is the base64url-encoded representation of the issued Credential.
 
+### IACA Certificate Metadata
+
+This profile defines an OpenID4VCI Credential Issuer Metadata parameter to publish Issuing Authority Certificate Authority (IACA) certificates.  These certificates are defined in ISO 18013-5 (mDL).  They are needed to verify mdoc certificate chains.  This parameter is:
+
+* `mdoc_iacas_uri`: A URL using the `https` scheme referencing a resource containing a JSON-encoded object containing IACA certificates.
+
+The contents of this resource is a JSON object containing an `iacas` element whose value is an array of objects.  Each object in the array contains a "certificate" member whose value is the base64-encoded DER representation of an IACA certificate.
+
+Below is a non-normative example of the contents an IACA certificates resource:
+
+```
+{
+  "iacas": [
+    {"certificate": "base64-encoded DER representation of an IACA certificate"},
+    ...
+  ]
+}
+```
+
+Other members MAY be defined and used in both the top-level object and the enclosed objects.
+
 # Document History
 
    [[ To be removed from the final specification ]]


tplooker (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1916,6 +1916,27 @@ The following is a non-normative example of a Credential Request with Credential
 
 The value of the `credential` claim in the Credential Response MUST be a string that is the base64url-encoded representation of the issued Credential.
 
+### IACA Certificate Metadata
+
+This profile defines an OpenID4VCI Credential Issuer Metadata parameter to publish Issuing Authority Certificate Authority (IACA) certificates.  These certificates are defined in ISO 18013-5 (mDL).  They are needed to verify mdoc certificate chains.  This parameter is:
+
```
peppelinux, line 1922: If I got it well, I recap:

- mdoc_iacas_uri points to an HTTPs resource that publishes a json object containing trust anchors' certificates needed to verify the mdoc issued by a VCI

may we use the certificate's CN/Subject as key name, instead of `"certificate"`? that would help the resolution of the issuer in a human readable way.

or, wdyt about having the certificate issuer's entity id in addition to that "certificate" member name?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1916,6 +1916,27 @@ The following is a non-normative example of a Credential Request with Credential
 
 The value of the `credential` claim in the Credential Response MUST be a string that is the base64url-encoded representation of the issued Credential.
 
+### IACA Certificate Metadata
+
+This profile defines an OpenID4VCI Credential Issuer Metadata parameter to publish Issuing Authority Certificate Authority (IACA) certificates.  These certificates are defined in ISO 18013-5 (mDL).  They are needed to verify mdoc certificate chains.  This parameter is:
```
peppelinux, line 1921:
```suggestion
This profile defines an OpenID4VCI Credential Issuer Metadata parameter to publish Issuing Authority Certificate Authority (IACA) certificates.  These X.509 certificates are defined in ISO 18013-5 (mDL). They are needed to verify mdoc X.509 certificate chains. This parameter is:
```

Sakurann:
there is no agreement on the issue to do this PR

Sakurann (changes requested):
even if we agree to add something like this, I am not sure this should be an mdoc specific parameter. if we agree, this is useful, it should be a generic parameter in the issuer metadata section. otherwise, it should go into an mdoc profile, not VCI.

selfissued:
As I saw it, the discussion in the issue was mostly about how to do it - not whether it was needed - hence the PR to give us a concrete proposal on a way to do it to discuss.

Unless you can get the root signing keys, you can't evaluate the X.509 certificate chain, so there's clearly a need for something like this.

If someone wants to propose a concrete way of generalizing this so that it's not mdoc specific, should that be the direction that people want to go, we could also discuss that.

Anyway, it seems like putting this topic on a call agenda would be productive.

tplooker:
I agree with @selfissued without a mechanism like the proposed the issuance of mdocs is at best incomplete because there is no way to verify the credentials that are issued, we should discuss it more on the next editors call

tlodderstedt (changes requested):
I object against this proposal. I don't think this is in scope for the oid4vci spec.

awoie:
IMO, the IACA could go into the issuer metadata and more specifically the credential format-specific metadata where also claims are described. The reason is why this makes sense is that different types can have different X.509 cert profiles and issuers.

awoie:
> IMO, the IACA could go into the issuer metadata and more specifically the credential format-specific metadata where also claims are described. The reason is why this makes sense is that different types can have different X.509 cert profiles and issuers.

I made a proposal here https://github.com/openid/OpenID4VCI/pull/161 by using @selfissued language. @selfissued please feel free to update this PR using https://github.com/openid/OpenID4VCI/pull/161 if the group feels more comfortable making this a credential issuer mso_mdoc specific metadata which I think is the right approach.

Sakurann:
closing this in a week, unless strong objections, because there is #161 and also not enough discussion on the topic.

PR #109, opened 2023-11-07
removing credential format specific parameters from Offer Credential in Appendix

paulbastian:
Closes #101 

Work left over from #86 
We could even delete the whole Credential Offer chapters from the Appendix.

PR #121, opened 2023-11-23
remove credential offer part from credential profiles

tlodderstedt:
After the recent changes, there is no longer any format specific part in a credential offer. This PR removes the respective definitions from the credential format profiles.


Sakurann:
When paul did a previous PR, I suggested not to remove entirely to show concrete examples how to use credential format specific identifiers, etc..
https://github.com/openid/OpenID4VCI/pull/109#pullrequestreview-1722170403

paulbastian:
I don't have strong opinions on this

I think having more than one example is nice and the credential offer in appendix could match to the credential issuer metadata. Especially having some examples for the changes in #95 would be nice.

On the other hand, we don't have examples on e.g. authroization response either and we could move some examples to the credential offer section if we want more examples

Sakurann (approved):
I came along with this.
but as @jogu said during the WG call, would like to see the example files unused elsewhere in the spec removed before we merge this.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1843,10 +1831,6 @@ When the `format` value is `jwt_vc_json-ld`, entire Credential Offer, Authorizat
 
 The definitions in (#server_metadata_ldp_vc) apply for metadata of credentials of this type as well. 
 
-#### Credential Offer
-
-The definitions in (#issuer_initiated_issuance_ldp_vc) apply for credentials of this type as well. 
-
 #### Authorization Details
 
 The definitions in (#issuer_initiated_issuance_ldp_vc) apply for credentials of this type as well.
```
Sakurann, line 1836:
```suggestion
The definitions in (#authorization_ldp_vc) apply for credentials of this type as well.
```
it does not compile otherwise

paulbastian:
I think it would be good to include at least one of the existing examples into the credential offer section, to see some variations of transaction code

paulbastian (approved):

paulbastian:
I did some changes, please review @tlodderstedt

PR #136, opened 2023-12-08
Merging credential_response_encryption parameters into a single object

### LINKED ISSUE #113: Credential Encryption Parameter Name Discrepancy
In the credential request we have the parameter names for the credential response encryption, defined in the text as follows:

- `credential_encryption_jwk`
- `credential_response_encryption_alg`
- `credential_response_encryption_enc`

`credential_encryption_jwk` does not include `response_` in it. Me and other implementers are wondering if this is intentional and why.

If this is not intentional, WDYT if we change the name to `credential_response_encryption_jwk`, for naming normalization? 

immagine

paulbastian:
Closes #113 

- [x] Is there a default value for `alg`, as there is a default value for `enc`?
- [ ] Shall we also merge objects in the Issuer Metadata?

I was unsure about the intention of the Issuer Metadata.  Issuer could offer  `credential_response_encryption_alg_values_supported` for offering the optionality to the wallet but set `require_credential_response_encryption` to false?

The current Issuer metadata:
* `credential_response_encryption_alg_values_supported`: OPTIONAL. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`alg` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
* `credential_response_encryption_enc_values_supported`: OPTIONAL. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
*: OPTIONAL. Boolean value specifying whether the Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request, with `true` indicating support. When the value is `true`, `credential_response_encryption_alg_values_supported` parameter MUST also be provided.  If omitted, the default value is `false`.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -718,8 +718,8 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
 * `credential_identifier`: REQUIRED when `credential_identifier` was returned from the Token Response. MUST NOT be used otherwise. String that identifies a Credential that is being requested to be issued. When this parameter is used, the `format` parameter and any other Credential format specific set of parameters such as those defined in (#format_profiles) MUST NOT be present.
 * `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If omitted, encryption is not performed.
     * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
-    * `alg`: OPTIONAL. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses. The default value is `xxx`.
-    * `enc`: OPTIONAL. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses. The default value is `A256GCM`.
+    * `alg`: REQUIRED. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses.
+    * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
 
 The `proof_type` claim is an extension point that enables the use of different types of proofs for different cryptographic schemes.
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -873,7 +873,7 @@ Credential Response can be immediate or deferred. The Credential Issuer MAY be a
 
 In other cases, the Credential Issuer MAY NOT be able to immediately issue a requested Credential and would want to send a `transaction_id` parameter to the Client to be used later to receive a Credential when it is ready. The HTTP status code MUST be 202 (section 10.2.3 of [@!RFC2616]).
 
-If the Client requested encrypted responses, the Credential Issuer MUST encode the information in the Credential Response as a JWT using the `credential_response_encryption` parameter in the Credential Request. If the Credential Response is encrypted, the media type of the response MUST bet set to `application/jwt`. If encryption was negotiated in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
+If the Client requested an encrypted response by including the `credential_response_encryption` object in the request, the Credential Issuer MUST encode the information in the Credential Response as a JWT using the  parameters from the `credential_response_encryption` object. If the Credential Response is encrypted, the media type of the response MUST bet set to `application/jwt`. If encryption was requested in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
 
 If the Credential Response is not encrypted, the media type of the response MUST be set to `application/json`.
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -716,7 +716,7 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
 * `proof`: OPTIONAL. Object containing the proof of possession of the cryptographic key material the issued Credential would be bound to.  The `proof` object MUST contain a following claim:
     * `proof_type`: REQUIRED. String denoting the key proof type. The value of this claim determines other claims in the key proof object and its respective processing rules. Key proof types defined in this specification can be found in (#proof_types).
 * `credential_identifier`: REQUIRED when `credential_identifier` was returned from the Token Response. MUST NOT be used otherwise. String that identifies a Credential that is being requested to be issued. When this parameter is used, the `format` parameter and any other Credential format specific set of parameters such as those defined in (#format_profiles) MUST NOT be present.
-* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If omitted, encryption is not performed.
+* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If this request element is not present, the corresponding credential response returned is not encrypted.
     * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
     * `alg`: REQUIRED. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses.
     * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -716,7 +716,7 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
 * `proof`: OPTIONAL. Object containing the proof of possession of the cryptographic key material the issued Credential would be bound to.  The `proof` object MUST contain a following claim:
     * `proof_type`: REQUIRED. String denoting the key proof type. The value of this claim determines other claims in the key proof object and its respective processing rules. Key proof types defined in this specification can be found in (#proof_types).
 * `credential_identifier`: REQUIRED when `credential_identifier` was returned from the Token Response. MUST NOT be used otherwise. String that identifies a Credential that is being requested to be issued. When this parameter is used, the `format` parameter and any other Credential format specific set of parameters such as those defined in (#format_profiles) MUST NOT be present.
-* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If omitted, no encryption is intended to be performed.
+* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If omitted, encryption is not performed.
     * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
     * `alg`: OPTIONAL. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses. The default value is `xxx`.
     * `enc`: OPTIONAL. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses. The default value is `A256GCM`.
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1945,7 +1945,8 @@ The value of the `credential` claim in the Credential Response MUST be a string 
    [[ To be removed from the final specification ]]
    
    -13
-  
+
+   * grouped `credential_encryption_jwk`, `credential_response_encryption_alg` and `credential_response_encryption_enc` from Credential Request into a single `credential_response_encryption` object
    * replaced `user_pin_required` in Credential Offer with a `tx_code` object that also now contains `description` and `length`
 
    -12
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -873,7 +873,7 @@ Credential Response can be immediate or deferred. The Credential Issuer MAY be a
 
 In other cases, the Credential Issuer MAY NOT be able to immediately issue a requested Credential and would want to send a `transaction_id` parameter to the Client to be used later to receive a Credential when it is ready. The HTTP status code MUST be 202 (section 10.2.3 of [@!RFC2616]).
 
-If the Client requested encrypted responses, the Credential Issuer MUST encode the information in the Credential Response as a JWT with `alg`, `enc` and using the public key specified by `jwk` from the `credential_response_encryption` object in the Credential Request. If the Credential Response is encrypted, the media type of the response MUST bet set to `application/jwt`. If encryption was negotiated in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
+If the Client requested encrypted responses, the Credential Issuer MUST encode the information in the Credential Response as a JWT using the `credential_response_encryption` parameter in the Credential Request. If the Credential Response is encrypted, the media type of the response MUST bet set to `application/jwt`. If encryption was negotiated in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
 
 If the Credential Response is not encrypted, the media type of the response MUST be set to `application/json`.
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -716,9 +716,10 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
 * `proof`: OPTIONAL. Object containing the proof of possession of the cryptographic key material the issued Credential would be bound to.  The `proof` object MUST contain a following claim:
     * `proof_type`: REQUIRED. String denoting the key proof type. The value of this claim determines other claims in the key proof object and its respective processing rules. Key proof types defined in this specification can be found in (#proof_types).
 * `credential_identifier`: REQUIRED when `credential_identifier` was returned from the Token Response. MUST NOT be used otherwise. String that identifies a Credential that is being requested to be issued. When this parameter is used, the `format` parameter and any other Credential format specific set of parameters such as those defined in (#format_profiles) MUST NOT be present.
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -716,9 +716,10 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
...
-* `credential_encryption_jwk`: OPTIONAL. An object containing a single public key as a JWK used for encrypting the Credential Response.
-* `credential_response_encryption_alg`: OPTIONAL. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] REQUIRED for encrypting Credential and/or Batch Credential Responses. If omitted, no encryption is intended to be performed. When the `credential_response_encryption_alg` is present, the `credential_encryption_jwk` MUST be present.
-* `credential_response_encryption_enc`: OPTIONAL. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] REQUIRED for encrypting Credential Responses. If `credential_response_encryption_alg` is specified, the default for this value is `A256GCM`. When `credential_response_encryption_enc` is included, `credential_response_encryption_alg` MUST also be provided.
+* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If omitted, no encryption is intended to be performed.
```
Sakurann, line 719:
```suggestion
* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If omitted, encryption is not performed.
```

Sakurann:
On the Credential request parameters, the logic between the presence of `jwk`, `enc`, and `alg` sounds weird. My suggestion would be to make all three mandatory, and remove the default value for `enc`.

On the Issuer metadata.... I don't think there is a need to group there, but we could either
1. move `require_credential_response_encryption` before `credential_response_encryption_alg_values_supported` and `credential_response_encryption_enc_values_supported` to make the order more natural, or
2. remove `require_credential_response_encryption` and say that the presence of both `credential_response_encryption_alg_values_supported` and `credential_response_encryption_enc_values_supported` (not either) means that "Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request".

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -716,9 +716,10 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
...
+    * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
+    * `alg`: OPTIONAL. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses. The default value is `xxx`.
+    * `enc`: OPTIONAL. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses. The default value is `A256GCM`.
```
peppelinux, line 722: I would go for having this REQUIRED as well, making default values according to specific algs, today, we may require errata in the future when something might became vulnerable or deprecable.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -872,7 +873,7 @@ Credential Response can be immediate or deferred. The Credential Issuer MAY be a
...
 
-If the Client requested encrypted responses, the Credential Issuer MUST encode the information in the Credential Reponse as a JWT with `credential_response_encryption_alg` and `credential_response_encryption_enc`. The JWT MUST be encrypted using the public key specified by `credential_encryption_jwk` in the Credential Request. If the Credential Response is encrypted, the media type of the response MUST bet set to `application/jwt`. If encryption was negotiated in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
+If the Client requested encrypted responses, the Credential Issuer MUST encode the information in the Credential Response as a JWT with `alg`, `enc` and using the public key specified by `jwk` from the `credential_response_encryption` object in the Credential Request. If the Credential Response is encrypted, the media type of the response MUST bet set to `application/jwt`. If encryption was negotiated in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
```
peppelinux, line 876: Does this mean that the the response is an encrypted json and not a Netsted JWT, signed and then encrypted?
I would give more clarification, as OIDC with the userinfo response (or id token) made:

> If the UserInfo Response is signed and/or encrypted, then the Claims are returned in a JWT and the content-type MUST be application/jwt. The response MAY be encrypted without also being signed. If both signing and encryption are requested, the response MUST be signed then encrypted, with the result being a Nested JWT, as defined in [[JWT]](https://openid.net/specs/openid-connect-core-1_0.html#JWT).

@sakimura @selfissued ^

paulbastian: I think this is about encryption only, this is how I read it

awoie:
> On the Credential request parameters, the logic between the presence of `jwk`, `enc`, and `alg` sounds weird. My suggestion would be to make all three mandatory, and remove the default value for `enc`.
> 
> On the Issuer metadata.... I don't think there is a need to group there, but we could either
> 
> 1. move `require_credential_response_encryption` before `credential_response_encryption_alg_values_supported` and `credential_response_encryption_enc_values_supported` to make the order more natural, or
> 2. remove `require_credential_response_encryption` and say that the presence of both `credential_response_encryption_alg_values_supported` and `credential_response_encryption_enc_values_supported` (not either) means that "Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request".

We cannot make all three keys REQUIRED since encryption is OPTIONAL. If we group them, then we can make the group key OPTIONAL and then we probably don't have to say anything about the sub-keys (jwk, enc, alg) anymore.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -716,9 +716,10 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
...
-* `credential_response_encryption_alg`: OPTIONAL. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] REQUIRED for encrypting Credential and/or Batch Credential Responses. If omitted, no encryption is intended to be performed. When the `credential_response_encryption_alg` is present, the `credential_encryption_jwk` MUST be present.
-* `credential_response_encryption_enc`: OPTIONAL. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] REQUIRED for encrypting Credential Responses. If `credential_response_encryption_alg` is specified, the default for this value is `A256GCM`. When `credential_response_encryption_enc` is included, `credential_response_encryption_alg` MUST also be provided.
+* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If omitted, encryption is not performed.
+    * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
+    * `alg`: OPTIONAL. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses. The default value is `xxx`.
```
```suggestion
    * `jwk`: Object containing a single public key as a JWK used for encrypting the Credential Response.
    * `alg`: JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses. The default value is `xxx`.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -716,9 +716,10 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
...
+    * `enc`: OPTIONAL. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses. The default value is `A256GCM`.
```
paulbastian, line 722:
```suggestion
    * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
```

```suggestion
    * `enc`: JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses. The default value is `A256GCM`.
```

paulbastian:
> > On the Credential request parameters, the logic between the presence of `jwk`, `enc`, and `alg` sounds weird. My suggestion would be to make all three mandatory, and remove the default value for `enc`.
> > On the Issuer metadata.... I don't think there is a need to group there, but we could either
> > 
> > 1. move `require_credential_response_encryption` before `credential_response_encryption_alg_values_supported` and `credential_response_encryption_enc_values_supported` to make the order more natural, or
> > 2. remove `require_credential_response_encryption` and say that the presence of both `credential_response_encryption_alg_values_supported` and `credential_response_encryption_enc_values_supported` (not either) means that "Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request".
> 
> We cannot make all three keys REQUIRED since encryption is OPTIONAL. If we group them, then we can make the group key OPTIONAL and then we probably don't have to say anything about the sub-keys (jwk, enc, alg) anymore.

Is the current proposal sufficient? I think when the object is Optional and claims within are required, the rules seem descriptive enough to me.

openid-4-verifiable-credential-issuance-1_0.md
awoie, line 876:
```suggestion
If the Client requested encrypted responses, the Credential Issuer MUST encode the information in the Credential Response as a JWT using the `credential_response_encryption` parameter in the Credential Request. If the Credential Response is encrypted, the media type of the response MUST bet set to `application/jwt`. If encryption was negotiated in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
```

paulbastian:
> > On the Credential request parameters, the logic between the presence of `jwk`, `enc`, and `alg` sounds weird. My suggestion would be to make all three mandatory, and remove the default value for `enc`.
> > On the Issuer metadata.... I don't think there is a need to group there, but we could either
> > 
> > 1. move `require_credential_response_encryption` before `credential_response_encryption_alg_values_supported` and `credential_response_encryption_enc_values_supported` to make the order more natural, or
> > 2. remove `require_credential_response_encryption` and say that the presence of both `credential_response_encryption_alg_values_supported` and `credential_response_encryption_enc_values_supported` (not either) means that "Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request".
> 
> We cannot make all three keys REQUIRED since encryption is OPTIONAL. If we group them, then we can make the group key OPTIONAL and then we probably don't have to say anything about the sub-keys (jwk, enc, alg) anymore.

I think I also tend to favor grouping for Issuer metadata.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -716,9 +716,10 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
...
-* `credential_response_encryption_alg`: OPTIONAL. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] REQUIRED for encrypting Credential and/or Batch Credential Responses. If omitted, no encryption is intended to be performed. When the `credential_response_encryption_alg` is present, the `credential_encryption_jwk` MUST be present.
-* `credential_response_encryption_enc`: OPTIONAL. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] REQUIRED for encrypting Credential Responses. If `credential_response_encryption_alg` is specified, the default for this value is `A256GCM`. When `credential_response_encryption_enc` is included, `credential_response_encryption_alg` MUST also be provided.
+* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If omitted, encryption is not performed.
+    * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
+    * `alg`: REQUIRED. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses.
+    * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
```
awoie, line 722:
```suggestion
* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If omitted, encryption is not performed.
    * `jwk`: Object containing a single public key as a JWK used for encrypting the Credential Response.
    * `alg`: JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses.
    * `enc`: JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
```

awoie: I don't think we will need the keywords here. Those are CONDITIONAL based on the presence of their parent.

awoie (changes requested):

awoie:
> > > On the Credential request parameters, the logic between the presence of `jwk`, `enc`, and `alg` sounds weird. My suggestion would be to make all three mandatory, and remove the default value for `enc`.
> > > On the Issuer metadata.... I don't think there is a need to group there, but we could either
> > > 
> > > 1. move `require_credential_response_encryption` before `credential_response_encryption_alg_values_supported` and `credential_response_encryption_enc_values_supported` to make the order more natural, or
> > > 2. remove `require_credential_response_encryption` and say that the presence of both `credential_response_encryption_alg_values_supported` and `credential_response_encryption_enc_values_supported` (not either) means that "Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request".
> > 
> > 
> > We cannot make all three keys REQUIRED since encryption is OPTIONAL. If we group them, then we can make the group key OPTIONAL and then we probably don't have to say anything about the sub-keys (jwk, enc, alg) anymore.
> 
> I think I also tend to favor grouping for Issuer metadata.

Let's do that in a separate PR.

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: I disagree, currently all subclaims are specified as OPTIONAL or REQUIRED throughout the whole spec

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: @peppelinux i think here we are trying to talk about the case when response is only encrypted and not signed. and that probably can be made clearer.

openid-4-verifiable-credential-issuance-1_0.md
awoie: @paulbastian @Sakurann Ok, then let's keep it the way it is written now. My intention is not to change the entire spec. I withdraw this comment.

awoie (approved):
It would be nice if my other suggestions would be committed otherwise looks good..

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -716,9 +716,10 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
...
-* `credential_encryption_jwk`: OPTIONAL. An object containing a single public key as a JWK used for encrypting the Credential Response.
-* `credential_response_encryption_alg`: OPTIONAL. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] REQUIRED for encrypting Credential and/or Batch Credential Responses. If omitted, no encryption is intended to be performed. When the `credential_response_encryption_alg` is present, the `credential_encryption_jwk` MUST be present.
-* `credential_response_encryption_enc`: OPTIONAL. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] REQUIRED for encrypting Credential Responses. If `credential_response_encryption_alg` is specified, the default for this value is `A256GCM`. When `credential_response_encryption_enc` is included, `credential_response_encryption_alg` MUST also be provided.
+* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If omitted, encryption is not performed.
```
tplooker, line 719:
```suggestion
* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If this request element is not present, the corresponding credential response returned is not encrypted.
```

openid-4-verifiable-credential-issuance-1_0.md
jogu: On editors call we seemed to have a consensus around "REQUIRED when credential_response_encryption parameter is present."

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -872,7 +873,7 @@ Credential Response can be immediate or deferred. The Credential Issuer MAY be a
...
 
-If the Client requested encrypted responses, the Credential Issuer MUST encode the information in the Credential Reponse as a JWT with `credential_response_encryption_alg` and `credential_response_encryption_enc`. The JWT MUST be encrypted using the public key specified by `credential_encryption_jwk` in the Credential Request. If the Credential Response is encrypted, the media type of the response MUST bet set to `application/jwt`. If encryption was negotiated in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
+If the Client requested encrypted responses, the Credential Issuer MUST encode the information in the Credential Response as a JWT using the `credential_response_encryption` parameter in the Credential Request. If the Credential Response is encrypted, the media type of the response MUST bet set to `application/jwt`. If encryption was negotiated in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
```
jogu, line 876: Reworded slightly:

- make it clear that credential_response_encryption is how the client requests encryption
- change 'negotiated' to 'requested' as I think the latter better reflects what happens

```suggestion
If the Client requested an encrypted response by including the `credential_response_encryption` object in the request, the Credential Issuer MUST encode the information in the Credential Response as a JWT using the  parameters from the `credential_response_encryption` object. If the Credential Response is encrypted, the media type of the response MUST bet set to `application/jwt`. If encryption was requested in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
```

jogu (approved):

openid-4-verifiable-credential-issuance-1_0.md
awoie: if we keep those key words in general, then we need to make all of them REQUIRED.

paulbastian: This is already resolved in the PR, closing this conversation

Sakurann (approved):

Issue #80, opened 2023-09-30
Binding Issuer to a service provider that the Issuer is using to host Issuer Metadata


Sakurann:
Current spec is written under the assumption that the issuer hosts the metadata itself. meaning that the entity whose metadata can be found at the path formed based on `credential_issuer` identifier URL is also the owner of that URL. However, the Issuer might be using managed services to host Issuer metadata. In which case, Entity A (actual issuer)'s metadata is being hosted under domain of an Entity B that provides that managed service. 

The issue we are facing is how to communicate to the wallet Entity A's domain/identifier so that it could be displayed to the user when obtaining user consent to get the credential issued.

One option we explored is to sign the metadata using the key controlled by Entity A. In our case, that key is expressed using a DID (did:web) so that the wallet obtains the key from a DID Document and also discovers a JWS signed by us that attests the binding between the DID of entity A and entity A's domain. which i realize is a lot... 

Maybe alternative could be, including domain of the actual issuer in the issuer metadata that is hosted by another entity B? I don't think this needs to be per credential, so could probably be a new top-level `issuer_domain` parameter

David-Chadwick:
_"Maybe alternative could be, including domain of the actual issuer in the issuer metadata that is hosted by another entity B?"_
I don't see how this can work as it will allow anyone (B) to masquerade as being the owner of A's metadata

jogu:
> The issue we are facing is how to communicate to the wallet Entity A's domain/identifier so that it could be displayed to the user when obtaining user consent to get the credential issued.

Relying on a user verifying that the displayed domain name is the expected one has proved to be a bad pattern.

That aside, I'm struggling to see what is the issue here that's a new problem in VC? Hosted openid connect providers have been around for a long time, usually relying on the customer setting entries in their DNS to point at the IP of the hosted services provider?

Sakurann:
@David-Chadwick you are right. so I think signing the request with B's key seems to be the only option

Sakurann:
@jogu 
> Relying on a user verifying that the displayed domain name is the expected one has proved to be a bad pattern.

maybe it is not ideal, but it's the most basic way. just like users are expected not to type in their credit card information when the SSL cert has not been verified or http is used.

> That aside, I'm struggling to see what is the issue here that's a new problem in VC? Hosted openid connect providers have been around for a long time, usually relying on the customer setting entries in their DNS to point at the IP of the hosted services provider?

[...]

PR #140, opened 2023-12-09
add `signed_credential_issuer_metadata` metadata parameter

Sakurann:
resolves issue #80.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1207,6 +1207,7 @@ This specification defines the following Credential Issuer Metadata:
 * `credential_response_encryption_enc_values_supported`: OPTIONAL. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
 * `require_credential_response_encryption`: OPTIONAL. Boolean value specifying whether the Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request, with `true` indicating support. When the value is `true`, `credential_response_encryption_alg_values_supported` parameter MUST also be provided.  If omitted, the default value is `false`.
 * `credential_identifiers_supported`: OPTIONAL. Boolean value specifying whether the Credential Issuer supports returning `credential_identifiers` parameter in the `authorization_details` Token Response parameter, with `true` indicating support. If omitted, the default value is `false`.
+* `signed_credential_issuer_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be digitally signed or MACed JWS using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At)claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_credential_issuer_metadata` metadata value SHOULD NOT appear as a claim in the JWT. Because metadata is only partially signed, the usage of DPoP [@!RFC9449] of MTLS [@!RFC8705] is RECOMMENDED to prevent an attacker from modifying the value of the Credential Endpoint to steal Access Tokens.
 * `display`: OPTIONAL. An array of objects, where each object contains display properties of a Credential Issuer for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
   * `name`: OPTIONAL. String value of a display name for the Credential Issuer.
   * `locale`: OPTIONAL. String value that identifies the language of this object represented as a language tag taken from values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
@@ -1313,6 +1314,10 @@ Server-provided nonces are an effective means for further reducing the chances f
 Implementations MUST follow [@!BCP195].
 
 Whenever TLS is used, a TLS server certificate check MUST be performed, per [@!RFC6125].
+
+##
+
+DPoP or MTLS needs to be recommended because metadata is only partially signed and attacker could do MITM by changing the credential endpoint value as suggested.
 
 # Implementation Considerations
 


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1314,6 +1315,10 @@ Implementations MUST follow [@!BCP195].
 
 Whenever TLS is used, a TLS server certificate check MUST be performed, per [@!RFC6125].
 
+##
+
+DPoP or MTLS needs to be recommended because metadata is only partially signed and attacker could do MITM by changing the credential endpoint value as suggested.
+
```
Sakurann, line 1321:
```suggestion
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1207,6 +1207,7 @@ This specification defines the following Credential Issuer Metadata:
...
+* `signed_credential_issuer_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be digitally signed or MACed JWS using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At)claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_credential_issuer_metadata` metadata value SHOULD NOT appear as a claim in the JWT. Because metadata is only partially signed, the usage of DPoP [@!RFC9449] of MTLS [@!RFC8705] is RECOMMENDED to prevent an attacker from modifying the value of the Credential Endpoint to steal Access Tokens.
```
Sakurann, line 1210:
```suggestion
* `signed_credential_issuer_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be digitally signed or MACed JWS using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_credential_issuer_metadata` metadata value SHOULD NOT appear as a claim in the JWT. Because metadata is only partially signed, the usage of DPoP [@!RFC9449] of MTLS [@!RFC8705] is RECOMMENDED to prevent an attacker from modifying the value of the Credential Endpoint to steal Access Tokens.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1207,6 +1207,7 @@ This specification defines the following Credential Issuer Metadata:
...
+* `signed_credential_issuer_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be digitally signed or MACed JWS using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_credential_issuer_metadata` metadata value SHOULD NOT appear as a claim in the JWT. Because metadata is only partially signed, the usage of DPoP [@!RFC9449] of MTLS [@!RFC8705] is RECOMMENDED to prevent an attacker from modifying the value of the Credential Endpoint to steal Access Tokens.
```
peppelinux, line 1210: I didn't get this 

> Because metadata is only partially signed, the usage of DPoP [@!RFC9449] of MTLS [@!RFC8705] is RECOMMENDED to prevent an attacker from modifying the value of the Credential Endpoint to steal Access Tokens.

can we add more clarification for this?

Since there may be some cases where the metadata attester and signer is different from the credential issuer, in the vest of the Trust Third Party like it could be a National Accreditation Body, in this case we must explode a sublist with:

- signed_credential_issuer_metadata_uri can be possibile, then providing the metadata by reference
- when metadata iss is a trusted third party, the sub MUST be included. The exp should be present as well, because a signed artifact should have an expiration time.

all good, even if I have to admit that's not clear how to establish the trust to the metadata issuer ... yes, out of scope here (but still in scope in our minds).

peppelinux:
@Sakurann using OpenID Federation the credential issuer publishes its metadata in its openid-federation endpoint (.well-known) and the trust is stablished with the federation trust chain, where signatures are verified and policies applied.

I'm wondering how this parameter would be used in an harmonic way using OpenID Federation

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: what I meant was tha tif the signed_metadata JWT does not contain `authorization_servers` and `credential_issuer` claims, and they are in the JSON object, an attacker can insert attacker's `credential_issuer` value and trick the wallet to send an access token from a good AS to attacker's RS, so that attacker can use it at the good RS. ping @danielfett if he could suggest a better language..?

> The exp should be present as well, because a signed artifact should have an expiration time.
I am not sure what `exp` time I would put as an issuer to the metadata. what would cause the the issuer metadata to expire..?

> that's not clear how to establish the trust to the metadata issuer ...
I agree that's something we will need to reiterate on later. for now, the goal is to put a general purpose mechanism that implementers can start using.

Sakurann:
> using OpenID Federation the credential issuer publishes its metadata in its openid-federation endpoint (.well-known)

VCI spec mandates the issuer metadata to be hosted at the `/.well-known/openid-credential-issuer`. How does the wallet know to use /.well-known/openid-federation and not  /.well-known/openid-credential-issuer? Isn't it more modular to host issuer metadata at  /.well-known/openid-credential-issuer and use /.well-known/openid-federation for trust establishment?

paulbastian (changes requested):
Assuming this is for Issuance-as-a-Service, imagine a scenario where CocaCola wants to issue credentials but doesn't want to deploy/host the services themselves but hires Microsoft to do so. In this case, I believe we have two other options that are already in the specification, hence don't require changes and are much cleaner overall:

Option A:
CocaCola creates a new subdomain and sets A/AAAA record to the static IP where Microsoft runs the issuer on their behalf.

Option B:
CocaCola deploys static .well-known files on their webserver. The Authroization and credential endpoint both point to Microsoft issuer domain where Microsoft runs the issuer on their behalf.

The approach in this PR seems odd to me, adding a lot of complexity and potentially opening doors for phishing attacks as well-known data need to be mixed and matched and it seems difficult to see the consequences of this change right now.

Sakurann:
No, Paul, the main use-case is to be able to sign the metadata values in zero-trust posture - RFC8414 already defines "signed_metadata" and this PR introduces the equivalent. The description of the parameter in this PR should be pretty clear.

As for the use-case you describe, it's definitely one Microsoft is also interested in, and unfortunately, what you describe does not work for Microsoft because we use DIDs. It has been discussed at length in the calls, at IIW and IETF Prague, and majority of the discussion is documented in the issue #80.

openid-4-verifiable-credential-issuance-1_0.md
tplooker: I don't think we should be putting this quite general recommendation around usage of DPoP in the definition of a credential issuer metadata parameter. I think we should shift this to a seperate security consideration.

peppelinux: _Memento OpenID Federation_

An `openid_credential_issuer` metadata derived from a trust chain actually doesn't bring all these worries about how to establish the trust with the metadata "issuer" or attester :-)

peppelinux:
> > using OpenID Federation the credential issuer publishes its metadata in its openid-federation endpoint (.well-known)
> 
> VCI spec mandates the issuer metadata to be hosted at the `/.well-known/openid-credential-issuer`. How does the wallet know to use /.well-known/openid-federation and not /.well-known/openid-credential-issuer? Isn't it more modular to host issuer metadata at /.well-known/openid-credential-issuer and use /.well-known/openid-federation for trust establishment?

Modularity is optional, security requirements not. The problem I see here is that we're considering a signed metadata because we're facing a lack of trust to the self-issued issuer metadata. Even if the metadata issuer is another entity, the problem is moved: <<How to establish the trust to that entity?>>

> Isn't it more modular to host issuer metadata at /.well-known/openid-credential-issuer and use /.well-known/openid-federation for trust establishment?

Yes, how?
openid-federation allows void metadata types like `"metadata": {"openid_credential_issuer": {}}`. How OpenID4VCI may notify that the trust should be established using openid-federation? (out-of-scope for this PR even if this PR alone doesn't give the big picture of the trust solution)

openid-4-verifiable-credential-issuance-1_0.md
danielfett: I agree with Tobias that DPoP/MTLS should be general consideration that we can point to here. The focus in the description should be to say that DPoP/MTLS prevent the misuse of tokens after they are sent to manipulated (attacker-controlled) endpoints.

Another note: In the sentence 'A `signed_credential_issuer_metadata` metadata value SHOULD NOT appear as a claim' I would make that a MUST NOT as otherwise the processing is undefined.

Sakurann: > Another note: In the sentence 'A signed_credential_issuer_metadata metadata value SHOULD NOT appear as a claim' I would make that a MUST NOT as otherwise the processing is undefined.

@danielfett why so..?

awoie (changes requested):
This PR should also explain what happens if the issuer metadata is different from the metadata in the `signed_credential_issuer_metadata`. I can see also some potential issues with credential format/type specific credential issuer metadata configurations that link to external resources. How are those external resources secured by this parametrer?

Sakurann:
> This PR should also explain what happens if the issuer metadata is different from the metadata in the signed_credential_issuer_metadata.

there is already a sentence saying `If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements.`

> I can see also some potential issues with credential format/type specific credential issuer metadata configurations that link to external resources. How are those external resources secured by this parametrer?

links to the external resources are secured the same way regardless whether the parameters are signed or not.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1251,6 +1251,7 @@ This specification defines the following Credential Issuer Metadata:
...
+* `signed_credential_issuer_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be digitally signed or MACed JWS using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_credential_issuer_metadata` metadata value SHOULD NOT appear as a claim in the JWT.
```
Sakurann, line 1254:
```suggestion
* `signed_credential_issuer_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be digitally signed or MACed JWS using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_credential_issuer_metadata` metadata value SHOULD NOT appear as a claim in the JWT. Required Credential Issuer metadata parameters MUST be present as plain JSON elements regardless of whether this parameter is present or not.
```
per @danielfett's comment during the editor's call

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1251,6 +1251,7 @@ This specification defines the following Credential Issuer Metadata:
...
+* `signed_credential_issuer_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be digitally signed or MACed JWS using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_credential_issuer_metadata` metadata value SHOULD NOT appear as a claim in the JWT. Required Credential Issuer metadata parameters MUST be present as plain JSON elements regardless of whether this parameter is present or not.
```
awoie, line 1254:
```suggestion
* `signed_credential_issuer_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using a third-party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_credential_issuer_metadata` metadata value MUST NOT appear as a claim in the JWT. Required Credential Issuer metadata parameters MUST be present in the Credential Issuer metadata  regardless of whether this parameter is present or not.
```

awoie (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
awoie: I'm having also issues with parsing this sentence:
> This JWT can also be used to identify the actual Credential Issuer when it is using a third-party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider.

openid-4-verifiable-credential-issuance-1_0.md
jogu: Could @danielfett possibly record his reasoning please? :-)

Sakurann:
in case it is not clear, this parameter was modelled after https://www.rfc-editor.org/rfc/rfc8414.html#section-2.1

paulbastian (changes requested):
As mentioned in the DCP Call, I am in favor of signed metadata in general, but it seems more reasonable to me having either the whole metadata unsigned or signed and distinguish with MIME types. Having parts signed and parts unsigned makes things complicated for the wallet as it has to parse both and then merge them. My feeling is that this seems to open several attack vectors.

David-Chadwick:
Sorry, but I do not understand why only part of the metadata needs to be signed, and not all of it. If the entity is trusted, then unsigned should be OK, but if not trusted, then all of it should be signed. 

However, referring to RFC 8414, this states "Consumers of the metadata MAY ignore the signed metadata if they do not support this feature.  If the consumer of the  metadata supports signed metadata, metadata values conveyed in the  signed metadata MUST take precedence over the corresponding values  conveyed using plain JSON elements."
So this implies that all the metadata must be unsigned, so that it is usable by all the clients that trust the publisher. For those undecided clients that sort-of trust the publisher, then the most sensitive bits of metadata may be signed as a JWT to increase trustworthiness in the most sensitive metadata parameters.

Providing this PR follows the guidance from RFC 8414 then introducing the signed metadata parameter as an optional parameter seems to be consistent with current practice.

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt, line 1254: It seems the wallet can ignore the signed data (e.g. if it does not support signed metadata) even though the the text states "If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements." since it also says "Required Credential Issuer metadata parameters MUST be present as plain JSON elements regardless of whether this parameter is present or not." I think that is dangerous. As an issuer not fully trusting the service provider hosting my metadata, I would like to ensure ANY wallet uses the signed data. Otherwise, an attacker could modify the unsigned data and circumvent the whole mechanism.

Sakurann: @awoie, @tlodderstedt, @danielfett please review the text below

```suggestion
* `signed_credential_issuer_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be digitally signed or MACed JWS using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to establish the binding between the actual Credential Issuer and the third party service provider used as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. Metadata values conveyed in the signed JWT MUST NOT be duplicated as plain JSON elements. A `signed_credential_issuer_metadata` metadata value SHOULD NOT appear as a claim in the JWT.
```

Sakurann:
@paulbastian @David-Chadwick

We discussed with Torsten and the suggestion is to say that `Metadata values conveyed in the signed JWT MUST NOT be duplicated as plain JSON elements.` instead of `If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements.` and remove a sentence `Required Credential Issuer metadata parameters MUST be present as plain JSON elements regardless of whether this parameter is present or not.`

with this approach, the wallet can still use the same media type application/json to fetch a JSON with only one claim `signed_credential_issuer_metadata` when the entire metadata is signed.

peppelinux:
in this PR it is not clear:

- how the trust to the signer is established
- which keys must/should be intended to use for signing the metadata, since different keys should be used for different purposes

I suggest another endpoint where the signed metadata are provided, allowing to obtain directly the signed metadata.
I suggest to use openid federation entity configuration for this since this is perfectly suitable for this scopes, [here an example ](https://github.com/italia/eudi-wallet-it-docs/blob/versione-corrente/docs/en/pid-eaa-issuance.rst#entity-configuration-credential-issuer)of the final result, considering that:

- sub and iss are the same, because the VCI self-signs the metadata
- authority hints points to the trusted third parties that will confirm the signed data and the keys to verify the signed metadata
- the top level jwks are the keys used for the signature and these are different from the one used for the credential issuance protocol

even using a progressive approach where the implementations don't fully support federation, I believe that's reasonable the use this openid federation artifact.

if we don't define how to establish the trust with the signed metadata issuer we risk to inflate the specs and move the problems without giving a solution for the real-world implementations

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: suggestion to make it clear that trust establishment is out of scope for now and should be defined in the profiles, because different implementations use different mechanisms and throughout the spec we have not recommended one over the other...

Sakurann:
```suggestion
* `signed_credential_issuer_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_credential_issuer_metadata` metadata value SHOULD NOT appear as a claim in the JWT.
```

Sakurann:
@peppelinux, @awoie @tlodderstedt @paulbastian please review if the most recent changes address/resolve your concerns.

paulbastian (changes requested):
The current proposal suggests that Wallets must support this feature, which is odd and not compliant with RFC 8414.
What is a reason not to sign the whole metadata?
Why do you want to partially sign metadata?

OpenID4VCI does not make any assumptions about the trust mechanisms. Here we are defining a mechanism that is heavily depending on such a trust mechanisms without defining it. In general the security of OAuth and OpenID4VCI seems to depend a lot on DNS and the well-known mechanisms, making the exception here feels like blowing a hole into the security model.

The purpose of this feature remains unclear to me, therefore I'm having trouble to find an alternative.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1340,6 +1340,7 @@ This specification defines the following Credential Issuer Metadata:
...
+* `signed_credential_issuer_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_credential_issuer_metadata` metadata value SHOULD NOT appear as a claim in the JWT. How the trust in the signer of the metadata can be established, and how to obtain the keys to validate the signature is out of scope of this specification and may be defined in the profiles of this specification.
```
selfissued, line 1343:
```suggestion
* `signed_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_credential_issuer_metadata` metadata value SHOULD NOT appear as a claim in the JWT. How the trust in the signer of the metadata can be established, and how to obtain the keys to validate the signature is out of scope of this specification and may be defined in the profiles of this specification.
```

selfissued (changes requested):
If we're going to introduce the same signed metadata concept as defined at https://www.rfc-editor.org/rfc/rfc8414.html#section-2.1, we should use the same name for it.

Sakurann:
@paulbastian it is not the intention of this PR to make this mandatory for the wallets. Can you please clarify what made you think so and what needs to be clarified.

The reason not to sign the whole metadata is to be able to use the same media type for the metadata regardless whether it is signed or not (always application/json without introducing application/jwt sometimes)


this PR intentionally allows multiple trust mechanisms (trust lists, OpenID federation, .well-known, etc) Please clarify what in this PR defines a trust mechanism and how that can be clarified.

The purpose of this feature has been discussed number of times on the call/issue/PR, it is to be able to establish trust in the metadata (concrete mechanisms remain open) and to integrity protect the metadata. Hope that's clearer.

jogu:
@paulbastian 

> OpenID4VCI does not make any assumptions about the trust mechanisms. Here we are defining a mechanism that is heavily depending on such a trust mechanisms without defining it. In general the security of OAuth and OpenID4VCI seems to depend a lot on DNS and the well-known mechanisms, making the exception here feels like blowing a hole into the security model.

I think we have to keep security and trust as separate topics, albeit they strongly interrelate.

Like OAuth, OID4VC assumes that the parties figure out a way to trust each other, in both specs how that trust is established is out of scope.

As @selfissued points out, OAuth has exactly the same signed meta data capability, so we're not creating any new holes in the security.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: the original rationale for choosing a different parameter name was because `signed_metadata` is defined for AS metadata and this is about RS metadata, but if there is an agreement that there is no need to differentiate, I don't have too strong of an opinion.

Sakurann:
```suggestion
* `signed_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_metadata` metadata value SHOULD NOT appear as a claim in the JWT. How the trust in the signer of the metadata can be established, and how to obtain the keys to validate the signature is out of scope of this specification and MUST be defined in the profiles of this specification.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2292,7 +2297,8 @@ Wallet Providers may also provide a market place where Issuers can register to b
    [[ To be removed from the final specification ]]
    
    -13
-
+  
+   * added `signed_credential_issuer_metadata` parameter
```
c2bo, line 2301:
```suggestion
   * added `signed_metadata` parameter
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1447,6 +1448,10 @@ Implementations MUST follow [@!BCP195].
 
 Whenever TLS is used, a TLS server certificate check MUST be performed, per [@!RFC6125].
 
+## Token Binding
+
+The usage of DPoP [@!RFC9449] or MTLS [@!RFC8705] is RECOMMENDED for the detection of replay attacks with access and refresh tokens.
+
```
Sakurann, line 1454:
```suggestion
```
will move current 13.6 to the security considerations in a different PR.

selfissued (approved):

paulbastian (commented):
I am still not convinced about this, but I won't block it for ID-1 as it somehow seems very important to some people

awoie (approved):

jogu (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1340,6 +1340,7 @@ This specification defines the following Credential Issuer Metadata:
...
+* `signed_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_metadata` metadata value SHOULD NOT appear as a claim in the JWT. The Wallet MUST establish trust in the signer of the metadata, and obtain the keys to validate the signature. The concrete mechanism how to do that is out of scope of this specification and may be defined in the profiles of this specification.
```
tlodderstedt, line 1343:
```suggestion
* `signed_metadata`: OPTIONAL. String that is a signed JWT. This JWT contains Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim denoting the third party service provider. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_metadata` metadata value SHOULD NOT appear as a claim in the JWT. The Wallet MUST establish trust in the signer of the metadata, and obtain the keys to validate the signature. The concrete mechanism how to do that is out of scope of this specification and may be defined in the profiles of this specification.
```

tlodderstedt, line 1343: I'm not sure about the security implications of making this feature optional for the wallet. I think if the issuer adds signed metadata, the issuer may want to force the wallet to use this data.

tlodderstedt (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: How are you suggesting to make it mandatory for the wallet?

jogu: I am not keen to make this mandatory for the wallet:

1. It can't be implemented without an ecosystem profile that makes clear how to check signatures, and that profile can also make it mandatory to implement for the wallet.
2. If we make it mandatory to implement it then probably becomes a thing we have to conformance test and that makes the conformance testing process more complicated for everyone, even the people that don't need this feature.
3. As an issuer you can essentially force wallets to support signed meta data by publishing all (or at least some critical part) of it's metadata only inside the signed object.
4. It would be inconsistent with https://datatracker.ietf.org/doc/html/rfc8414#section-2.1

paulbastian, line 1343: Why is "A `signed_metadata` metadata value SHOULD NOT appear as a claim in the JWT." only a SHOULD NOT and not MUST NOT?

paulbastian:
If a `signed_metadata` exists that contains `display` properties for an issuer, this would on the one hand be more secure to really identify the Crednetial Issuer towards the Wallet, because right now in unsigned metadata they are only self-asserted but used in the Wallets.

What prevents a malicious issuer to copy that `signed_metadata` and put that into his own metadata and impersonate the legitimate issuer?

paulbastian:
From the discussion in #202 it became clear that people have different ideas and motivations for `signed_metadata`. As I was asking earlier about the motivation of people, I ask it here again, as it seems not clear.

awoie (commented):
I reviewed the PR again and I have some questions related to `signed_metadata`.

The `signed_metadata` has the following claims according to this PR:

`iss`: denoting the party attesting to the claims in the signed metadata
`sub`: (optional) denoting the third party service provider

Can you give a concrete example? I thought this PR wants to enable third-party services to host (parts of) the metadata. 

I'm confused about who is hosting whose metadata now. IMO, it makes sense if the third-party service hosts the JWT but then the issuer should sign it with their key that is trusted by some trust framework.

I don't want this PR to suggest any form of attestation/accreditation.

awoie:
> I reviewed the PR again and I have some questions related to `signed_metadata`.
> 
> The `signed_metadata` has the following claims according to this PR:
> 
> `iss`: denoting the party attesting to the claims in the signed metadata `sub`: (optional) denoting the third party service provider
> 
> Can you give a concrete example? I thought this PR wants to enable third-party services to host (parts of) the metadata.
> 
> I'm confused about who is hosting whose metadata now. IMO, it makes sense if the third-party service hosts the JWT but then the issuer should sign it with their key that is trusted by some trust framework.
> 
> I don't want this PR to suggest any form of attestation/accreditation.

I'm also unclear how .well-known metadata discovery would work based on the `credential_issuer` identifier in this case.

Sakurann:
> What prevents a malicious issuer to copy that signed_metadata and put that into his own metadata and impersonate the legitimate issuer?

Nothing. But copying will be meaningless because the wallet will be able to identify the mismatch under the correct framework.

> Is sub supposed to match with the Crednetial Issuer base URL? Can we have an example for the third-party hosting strategy? The text alone does not seem obvious enough to me.

I think it is up to the profile/ecosystem. In some cases sub can match with the base URL, in some cases it can be the DID, etc.

paulbastian:
I propose to mandate that `sub` matches the `crednetial_issuer` origin of the Third Party. In that way you prevent malicious issuers impersonating `signed_metadata` from somebody else. By this means, you have basic assurance and you don't rely 100% on trust mechanisms defined by external profile and iIt matches very nicely the security concept of OAuth.

Sakurann:
> I'm also unclear how .well-known metadata discovery would work based on the credential_issuer identifier in this case. I double-checked https://datatracker.ietf.org/doc/html/rfc8414#section-2.1 and does this mean that signed_metadata is not obtained via HTTPS from the issuer but instead the whole object that includes the signed_metadata parameter is obtained through some out-of-band mechanism?

Issuer metadata will be obtained by appending .well-known to the credential_issuer identifier. That metadata is JSON and one of the claims in that JSON can be the signed_metadata parameter. This should be clear because signed_metadata parameter is defined as one of the metadata parameters?

Sakurann:
> I'm confused about who is hosting whose metadata now. IMO, it makes sense if the third-party service hosts the JWT but then the issuer should sign it with their key that is trusted by some trust framework.

If the actual-issuer.com is using provider.com to host the metadata, credential_issuer identifier to which .well-known gets appended. The JWT in the signed_metadata is signed by actual-issuer.com (how the wallet establishes trust in that public key is out of scope for VCI) and that JWT's sub claim contains actual-issuer's identifier.

awoie:
> > I'm confused about who is hosting whose metadata now. IMO, it makes sense if the third-party service hosts the JWT but then the issuer should sign it with their key that is trusted by some trust framework.
> 
> If the actual-issuer.com is using provider.com to host the metadata, credential_issuer identifier to which .well-known gets appended. The JWT in the signed_metadata is signed by actual-issuer.com (how the wallet establishes trust in that public key is out of scope for VCI) and that JWT's sub claim contains actual-issuer's identifier.

That makes a lot of sense. I didn't read the definition of `sub` like this since the PR currently says:
> sub: (optional) denoting the third party service provider

This should be changed to `credential_issuer` identifier, right?

Sakurann:
> I propose to mandate that sub matches the crednetial_issuer origin of the Third Party. In that way you prevent malicious issuers impersonating signed_metadata from somebody else.

I think this could work.

paulbastian:
I also read the current text that `iss` can be actual-issuer.com but could also be trust-provider.com, depending how the profile works, correct?

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: This is from RFC8414. I don't know why RFC8414 made it a should @selfissued, would you know?

Sakurann:
> I also read the current text that iss can be actual-issuer.com but could also be trust-provider.com, depending how the profile works, correct?

Yes, if there is a federation trust chain for example, I think.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1340,6 +1340,7 @@ This specification defines the following Credential Issuer Metadata:
...
+* `signed_metadata`: OPTIONAL. String that is an entire signed JWT. This JWT can contain Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim matching the Credentials Issuer identifier. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_metadata` metadata value SHOULD NOT appear as a claim in the JWT. The Wallet MUST establish trust in the signer of the metadata, and obtain the keys to validate the signature. The concrete mechanism how to do that is out of scope of this specification and may be defined in the profiles of this specification.
```
Sakurann, line 1343:
```suggestion
* `signed_metadata`: OPTIONAL. String that is a signed JWT. This JWT contains Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim matching the Credentials Issuer identifier. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_metadata` metadata value SHOULD NOT appear as a claim in the JWT. The Wallet MUST establish trust in the signer of the metadata, and obtain the keys to validate the signature. The concrete mechanism how to do that is out of scope of this specification and may be defined in the profiles of this specification.
```

jogu:
re:

> This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim matching the Credentials Issuer identifier.

My understanding of Paul's comment ("I propose to mandate that sub matches the crednetial_issuer origin of the Third Party. In that way you prevent malicious issuers impersonating signed_metadata from somebody else. ") was he was suggesting 'MUST' style clause.

Is there a reason to use 'in which case' and 'SHOULD'? It seems like we really could just say:

> `sub` MUST be the Credential Issuer Identifier" 

and that completely avoids any concern that a signed metadata could be taken and put on a different domain and appear to still be valid.

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: I'm ok with adding text stating something along the lines of bullet 3.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1340,6 +1340,7 @@ This specification defines the following Credential Issuer Metadata:
...
+* `signed_metadata`: OPTIONAL. String that is a signed JWT. This JWT contains Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim matching the Credentials Issuer identifier. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_metadata` metadata value SHOULD NOT appear as a claim in the JWT. The Wallet MUST establish trust in the signer of the metadata, and obtain the keys to validate the signature. The concrete mechanism how to do that is out of scope of this specification and may be defined in the profiles of this specification.
```
tlodderstedt, line 1343:
```suggestion
* `signed_metadata`: OPTIONAL. String that is a signed JWT. This JWT contains Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using a third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim matching the Credentials Issuer identifier. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. If the Issuer wants to enforce use of signed metadata, it SHOULD omit the respective metadata parameters from the unsigned part of the credential issuer metadata. A `signed_metadata` metadata value SHOULD NOT appear as a claim in the JWT. The Wallet MUST establish trust in the signer of the metadata, and obtain the keys to validate the signature. The concrete mechanism how to do that is out of scope of this specification and may be defined in the profiles of this specification.
```

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: added a suggestion

tlodderstedt (approved):
approved with the suggested change

openid-4-verifiable-credential-issuance-1_0.md
awoie: I agree with @tlodderstedt , otherwise we won't have interop. I added some editorial suggestions plus saying that the `signed_metadata` MUST NOT appear in the `signed_metadata`.

```suggestion
* `signed_metadata`: OPTIONAL. The `signed_metadata` parameter is a signed JSON Web Token (JWT) containing Credential Issuer metadata parameters as claims. The `signed_metadata` parameter MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST include an `iat` (Issued At) claim and an `iss` (Issuer) claim, indicating the party attesting to the claims in the `signed_metadata` parameter. The `signed_metadata` parameter can serve to identify the actual Credential Issuer when using a third-party service provider as a domain to host its Credential Issuer metadata. In such cases, the `signed_metadata` parameter SHOULD include a `sub` (Subject) claim matching the Credential Issuer identifier. If the Wallet supports the `signed_metadata` parameter, the metadata parameters conveyed as claims in the `signed_metadata` parameter MUST take precedence over the corresponding metadata parameters in the unsigned embedding Credential Issuer metadata. If the Issuer wishes to enforce the use of the `signed_metadata` parameter, it MUST exclude the respective metadata parameters contained as claims in the `signed_metadata` parameter from the unsigned Credential Issuer metadata. The `signed_metadata` parameter MUST NOT contain a `signed_metadata` claim. The Wallet MUST establish trust in the signer of the JWT and obtain the keys to validate the signature before utilizing the Credential Issuer metadata parameter values conveyed as claims in the `signed_metadata` parameter value. The specific mechanism for doing so is beyond the scope of this specification and MAY be defined in the profiles associated with this specification.
```

awoie: >  If the Issuer wants to enforce use of signed metadata, it SHOULD omit the respective metadata parameters from the unsigned part of the credential issuer metadata.

@tlodderstedt IMO, this is unnecessary since the `signed_metadata` takes precedence. What happens if a wallet cannot validate the JWS and finds the same parameter in the `signed_metadata` and the unsigned metadata? According to this spec, it MUST fail since `signed_metadata` takes precedence. Do we need to clarify this better?

awoie: Should the unsigned metadata be a fallback in case the wallet cannot validate the JWS, or do we accept the fact that in that case wallets that don't trust the keys cannot proceed? In the latter case we should probably change the MUST to a "SHOULD establish trust".

awoie, line 1343: I read the proposal described in https://github.com/openid/OpenID4VCI/issues/80#issuecomment-1786501377 again and I noticed that the current proposal in this PR is different. The difference is that in https://github.com/openid/OpenID4VCI/issues/80#issuecomment-1786501377, the suggestion is to have the following JWT claim values in `signed_metadata`: 
- `iss` is actual credential issuer
- `sub` is the third-party service hosting the metadata aka `signed_metadata`

The proposal above ^ suggests that the credential issuer is attesting/issuing the JWT, whereas in this PR, it is the opposite as follows:
- `iss` is the third-party service hosting the metadata aka `signed_metadata`
- `sub` is the credential issuer identifier

I thought about the proposal in this PR and I cannot get my around it. If the `signed_metadata` is hosted under the domain of a third-party service provider which is different from the actual credential issuer's domain, how would you perform .well-known discovery? IMO, this won't be possible unless one has someone has for example a DNS redirect. If this assumption is wrong and the idea is to keep the same domain for credential issuer and third-party service provider, why would you need this feature at all?

awoie (changes requested):
IMO supportive in general but I don't fully understand the proposal and the implications. Could you please clarify my questions.

Sakurann (commented):
The current state of this PR reflects the discussion in this PR, so I don't think we should be hang up on the discussion in the issue 80. Bottomline, there are two entities involved and one is the iss and another is sub that allows to establish trust in the metadata.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: If the wallet supports signed metadata it must use that data only if it trusts the public key, otherwise it should use the plaintext values. So if the issuer enforces all the metadata being signed and the wallet can't establish trust in the public key transaction should fail. So current MUST is fine.

awoie:
> The current state of this PR reflects the discussion in this PR, so I don't think we should be hang up on the discussion in the issue 80. Bottomline, there are two entities involved and one is the iss and another is sub that allows to establish trust in the metadata.

Proposal in this PR is as follows:
- assuming we have a credential issuer identifier = https://gov.com/issuer
- then, .well-known discovery is https://gov.com/.well-known/openid-credential-issuer
- this metadata contains a `signed_metadata` with the following claims:

```
  "iss" = "https://1234.acme.com/issuer"
  "sub" = "https://gov.com"
  "display": { "name" = "gov.com" }
```

What is the purpose of this additional JWS if gov.com hosts the metadata themselves, under their own domain? Perhaps I'm missing something but that does not address what was described in #80.

Proposal in #80 would work as follows:

- assuming we have a credential issuer identifier = https://1234.acme.com/issuer
- then, .well-known discovery is https://1234.acme.com/.well-known/issuer/openid-credential-issuer
- this metadata contains a `signed_metadata` with the following:

```
  "iss = https://gov.com
  "sub = https://1234.acme.com/issuer
  "display": { "name" = "gov.com" }
```

Assuming credentials will be issued by https://1234.acme.com/issuer, proposal 2 perhaps makes more sense if the goal is to enable a third-party to host the metadata for gov.com, be the actual issuer but change the display name to gov.com.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: The current state of this PR reflects the discussion in this PR, so I don't think we should be hang up on the discussion in the issue 80. Bottomline, there is an entity that issues JWT (iss) and an entity that hosts the metadata (sub)  - how do you call each of those (actual issuer or the provider) does not matter as long as sub == credential issuer identifier

Sakurann:
> What is the purpose of this additional JWS if gov.com hosts the metadata themselves, under their own domain? 

if gov.com hosts the metadata themselves and the wallet trusts gov.com directly there is no need to use this signed metadata parameter.

awoie:
> > What is the purpose of this additional JWS if gov.com hosts the metadata themselves, under their own domain?
> 
> if gov.com hosts the metadata themselves and the wallet trusts gov.com directly there is no need to use this signed metadata parameter.

Exactly, and if the domain is different, then .well-known would not work, right?

awoie:
I think what needs to be changed is that the term 'actual issuer' is not well defined. I believe we talk about the same if we replace 'actual issuer' with something like the issuer referenced by the display.name although i'm not super happy with that definition either. I won't block the PR anymore as long as we make it a bit more clear what 'actual' means.

awoie (commented):
Not blocking anymore if 'actual' is defined better.

openid-4-verifiable-credential-issuance-1_0.md
awoie, line 1343:
```suggestion
* `signed_metadata`: OPTIONAL. String that is a signed JWT. This JWT contains Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim matching the Credentials Issuer identifier. In this case, the `iss` claim also defines the actual Credential Issuer. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_metadata` metadata value SHOULD NOT appear as a claim in the JWT. The Wallet MUST establish trust in the signer of the metadata, and obtain the keys to validate the signature. The concrete mechanism how to do that is out of scope of this specification and may be defined in the profiles of this specification.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1340,6 +1340,7 @@ This specification defines the following Credential Issuer Metadata:
...
+* `signed_metadata`: OPTIONAL. String that is a signed JWT. This JWT contains Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim matching the Credentials Issuer identifier. In this case, the `iss` claim also defines the actual Credential Issuer. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_metadata` metadata value SHOULD NOT appear as a claim in the JWT. The Wallet MUST establish trust in the signer of the metadata, and obtain the keys to validate the signature. The concrete mechanism how to do that is out of scope of this specification and may be defined in the profiles of this specification.
```
Sakurann, line 1343:
```suggestion
* `signed_metadata`: OPTIONAL. String that is a signed JWT. This JWT contains Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim matching the Credentials Issuer identifier. In this case, the `iss` claim also defines the actual Credential Issuer. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. If the Issuer wants to enforce use of signed metadata, it SHOULD omit the respective metadata parameters from the unsigned part of the credential issuer metadata. A `signed_metadata` metadata value SHOULD NOT appear as a claim in the JWT. The Wallet MUST establish trust in the signer of the metadata, and obtain the keys to validate the signature. The concrete mechanism how to do that is out of scope of this specification and may be defined in the profiles of this specification.
```

Sakurann:
@paulbastian what concrete changes are you requesting with the current request for changes? (sorry if I missed.)

paulbastian:
One more question came to my mind: If there are 3 different trust mechanisms that all use `signed_metadata` but have different structure within, how does the wallet know how to parse `signed_metadata`? Probably only by JWT type?

Although OpenID4VCI is trust agnostic, we force trust frameworks to use JWT as a mechanism. Are we sure that this fits to any of the existing mechanisms?

openid-4-verifiable-credential-issuance-1_0.md
paulbastian, line 1343:
```suggestion
* `signed_metadata`: OPTIONAL. String that is a signed JWT. This JWT contains Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim, an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata and a `sub` (Subject) claim matching the Credential Issuer's origin identifier. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. A `signed_metadata` metadata value MUST NOT appear as a claim in the JWT. The Wallet MUST establish trust in the signer of the metadata, and obtain the keys to validate the signature. The concrete mechanism how to do that is out of scope of this specification and may be defined in the profiles of this specification.
```

peppelinux (changes requested):
I'm not in favour of merging this PR, at least at this stage and with the current content, for the following reasons:

1. having unsigned metadata carrying signed metadata with the shared/overlapping parameters is a matrioska design I don't like. I suggest using a separate well-known endpoint for a specific content type like the signed metadata
2. Having a single .well known carrying the vci metadata and a single signed metadata embedded in it doesn't scale for entities that are part of multiple federation/context where more than a single trusted third party (signed metadata issuers) might be involved. I can't imagine that my VCI, with that wellknow endpoint can only host a single signed metadata signed by a single trusted third party. (Memento Federation: entity configurations allows multiple authority hints ...)

I don't want to be negative, I can give positive contributions on this field, however, with the current stage of the content I'm not in favor of merging this.

openid-4-verifiable-credential-issuance-1_0.md
awoie: After reading the text again, I think it should be fine. We could always be more explicit by saying something like the following but I'm not insisting:
```suggestion
* `signed_metadata`: OPTIONAL. String that is a signed JWT. This JWT contains Credential Issuer metadata parameters as claims. The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain an `iat` (Issued At) claim and an `iss` (Issuer) claim denoting the party attesting to the claims in the signed metadata. This JWT can also be used to identify the actual Credential Issuer when it is using a third party service provider as a domain to host its metadata file, in which case, the JWT SHOULD contain `sub` (Subject) claim matching the Credentials Issuer identifier. If the Wallet supports signed metadata, metadata values conveyed in the signed JWT MUST take precedence over the corresponding values conveyed using plain JSON elements. If the Issuer wants to enforce use of signed metadata, it SHOULD omit the respective metadata parameters from the unsigned part of the credential issuer metadata. A `signed_metadata` metadata value SHOULD NOT appear as a claim in the JWT. The Wallet MUST establish trust in the signer of the metadata, and obtain the keys to validate the signature. If the Wallet cannot establish trust in the signer of the metadata, the Wallet MUST reject the metadata parameters contained in the `signed_metadata` parameter value. The concrete mechanism how to do that is out of scope of this specification and may be defined in the profiles of this specification.
```

awoie: Resolved by latest update of this PR (see comment on "actual" issuer).

c2bo (approved):

paulbastian (commented):
not blocking the merge

peppelinux (approved):

paulbastian (approved):

Sakurann:
@peppelinux's concerns have been discussed and agreed to keep discussing how the requirement of multiple issuer metadata signers can be addressed.

PR #141, opened 2023-12-09
clarify that `logo_uri` can be of multiple schemes

### LINKED ISSUE #43: logo uri in the `credentials_supported` display metadata
### Original Reporter: KristinaYasuda

In the [Objects comprising credentials\_supported parameter](https://openid.bitbucket.io/connect/openid-4-verifiable-credential-issuance-1_0.html#section-10.2.3.1) section logo uri is defined as follows:

> `logo`: OPTIONAL. A JSON object with information about the logo of the Credential with a following non-exhaustive list of parameters that MAY be included:
>
> ‌    `url`: OPTIONAL. URL where the Wallet can obtain a logo of the Credential from the Credential Issuer.

We have a use-case to use data uri instead of an https URL. Would the WG be open to changing the definition to something like below?

> `uri`: OPTIONAL. URI where the Wallet can obtain a logo of the Credential from the Credential Issuer. Wallet needs to determine the scheme, since the URI value could use `httpss:` scheme, `data:` scheme, etc.

‌

Sakurann:
resolves #43.


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1220,7 +1220,7 @@ This specification defines the following Credential Issuer Metadata:
       * `name`: REQUIRED. String value of a display name for the Credential.
       * `locale`: OPTIONAL. String value that identifies the language of this object represented as a language tag taken from values defined in BCP47 [@!RFC5646]. Multiple `display` objects MAY be included for separate languages. There MUST be only one object for each language identifier.
       * `logo`: OPTIONAL. A JSON object with information about the logo of the Credential with a following non-exhaustive list of parameters that MAY be included:
-          * `url`: OPTIONAL. URL where the Wallet can obtain a logo of the Credential from the Credential Issuer.
+          * `uri`: OPTIONAL. String value that contains a URI where the Wallet can obtain a logo of the Credential from the Credential Issuer. Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
```
peppelinux, line 1223: I would add here the recommendation that the log should be provided in SVG format, to prevent scaling and rendering problems in different display resolutions. In the italian impl we have required it, while in this specs would just be recommend to make the implementer aware of this.

Sakurann: I think I understand where you are coming from, but I am not exactly comfortable with adding that spec text... for example, our implementation only supports .png, .jpeg, or .jpg.

tplooker: I think we need to be stricter here, having a URI that could take many forms means the wallet has to support all of these, including both hosted resources (resolved over https) and data URI's. Can we remove some of these options and also potentially limit the image formats too?

peppelinux: do you remember the webp exploit?
https://nvd.nist.gov/vuln/detail/CVE-2023-4863

probably an implementer would like to hav some filters or just some recommendation about how to mitigate the risk to have bad UX (rasters with different resolutions) or security problems

Sakurann: Ideally, it should be stricter, but unfortunately, I don't think it is realistically possible though I am ok adding an implementation considerations or security considerations why SVG is better,

danielfett: One option would be to make URIs an array such that an implementation can fallback to the second entry in that URL in case it doesn't support the first format (and so on). A mime type per entry would help to identify the proper entry to choose, but may also lead to new errors (wrong image parser being used).

@peppelinux I see the security issue largely independent from what we do here. SVGs have their own history of problems (like XSS and SSRF).

peppelinux: I agree that SVG is not more or less secure for any other format, since the vulnerability is in the interpreter and not in the format itself. I mentioned the possibility to give more guidance to the implementers. Out of scope from this PR

paulbastian:
As a sidetopic: I wonder why there is a logo for the credential, but not for the issuer?

peppelinux:
> As a sidetopic: I wonder why there is a logo for the credential, but not for the issuer?

Good catch @paulbastian

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1220,7 +1220,7 @@ This specification defines the following Credential Issuer Metadata:
...
           * `alt_text`: OPTIONAL. String value of an alternative text of a logo image.
```
peppelinux, line 1224:
```suggestion
          * `alt_text`: OPTIONAL. String value of an alternative text of a logo image.
          * `media_type`: RECOMMENDED. Media type of the logo image. It MUST be one of the types listed in the IANA Media Types registry [@IANA.MediaTypes]. Eg: `image/svg+xml` or `image/png`.
```

openid-4-verifiable-credential-issuance-1_0.md
tplooker: > Ideally, it should be stricter, but unfortunately, I don't think it is realistically possible though I am ok adding an implementation considerations or security considerations why SVG is better,

Whats the rational for why here? Just that certain existing implementations have made different decisions?

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: Is there therefore something to be added in the security consideration, something like

> Incorrect or mismatched media types can lead to security vulnerabilities. For example, if an application expects a file to be a certain type based on its media type, but the file's actual content is different, this could lead to issues like code injection attacks. By correctly specifying the media type, applications can better validate and sanitize incoming data, reducing the risk of such attacks.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: I would not expect one spec to be able to enforce one image format over the other on the web. Different systems have different restrictions and will use one format and not the other

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: an example of static file type inspection is using libmagic
https://linux.die.net/man/3/libmagic

Sakurann: I am more in favor of changing the structure to the array. with this approach, logo will have to be array of objects resulting in additional complexity. pointing to RFC2397 for data: scheme is sufficient IMO

awoie (changes requested):
I don't think we should introduce `media_type` and I don't think we should encourage people to use data URIs. This would not be great in terms of performance. IMO, it is fine to have one logo URI, and potentially one URL for the thumbnail. The media type of the image is determined by the HTTP response content-type.

Sakurann:
@awoie the current PR text does not introduce 'media_types'.

Some want to use data uris and I don't think it is up to us to prevent that.

Without an agreement in the industry on one image format, I don't see a better option than having an array, even if it might end up having one entry in the end.

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: I can buy this @awoie :+1:

awoie:
> @awoie the current PR text does not introduce 'media_types'.
> 
> Some want to use data uris and I don't think it is up to us to prevent that.
> 
> Without an agreement in the industry on one image format, I don't see a better option than having an array, even if it might end up having one entry in the end.

I want to avoid that a wallet has to fetch all logos from the array until the wallet finds a suitable media type. If it is really required to support multiple logos then I support @peppelinux suggestion and we have to include the media type in the metadata to avoid unnecessary fetching of logos of unsupported media types. However,  I'm questioning if multiple logos is needed in the first place.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1264,7 +1264,7 @@ This specification defines the following Credential Issuer Metadata:
       * `name`: REQUIRED. String value of a display name for the Credential.
       * `locale`: OPTIONAL. String value that identifies the language of this object represented as a language tag taken from values defined in BCP47 [@!RFC5646]. Multiple `display` objects MAY be included for separate languages. There MUST be only one object for each language identifier.
       * `logo`: OPTIONAL. A JSON object with information about the logo of the Credential with a following non-exhaustive list of parameters that MAY be included:
-          * `url`: OPTIONAL. URL where the Wallet can obtain a logo of the Credential from the Credential Issuer.
+          * `uris`: OPTIONAL. Array of values that each identify a URI where the Wallet can obtain a logo of the Credential from the Credential Issuer. Wallet needs to identify and pick the scheme that it supports, since the URI value could use `https:` scheme, `data:` scheme, etc.
```
Sakurann, line 1267:
```suggestion
          * `uris`: OPTIONAL. String value that contains a URI where the Wallet can obtain a logo of the Credential from the Credential Issuer. Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
```

Sakurann:
for the purpose of this PR, let's stick to making it a string and renaming to `uri` to prevent breaking changes before ID-1. any other feature suggestions, please open the issue.

tplooker:
> for the purpose of this PR, let's stick to making it a string and renaming to uri to prevent breaking changes before ID-1. any other feature suggestions, please open the issue.

It doesn't feel like there is consensus on this so its going to be hard to prevent a breaking change. Leaving this feature open to all image formats and both remote and local resolution isn't a solution either IMO, standards are about making decisions and this proposal at the moment is the opposite of a decision. I'm fine with merging the PR but only on the basis that post ID-1 we have agreement to revisit and make it stricter, otherwise we need to resolve it now.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1846,6 +1846,9 @@ The following additional Credential Issuer metadata are defined for this Credent
       * `display`: OPTIONAL. An array of objects, where each object contains display properties of a certain claim in the Credential for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
           * `name`: OPTIONAL. String value of a display name for the claim.
           * `locale`: OPTIONAL. String value that identifies language of this object represented as language tag values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
+          * `logo`: OPTIONAL. A JSON object with information about the logo of the Credential Issuer with a following non-exhaustive list of parameters that MAY be included:
+            * `uri`: OPTIONAL. String value that contains a URI where the Wallet can obtain a logo of the Credential from the Credential Issuer. Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
+            * `alt_text`: OPTIONAL. String value of an alternative text of a logo image.
```
awoie, line 1851: Please delete this. This is new and unrelated to the #43. If a mechanism for the issuer logo is needed, then this should not be defined format-specific.
```suggestion
```

awoie (changes requested):
I'm ok with the data URI since it is not a breaking change. Just explains that URL can contain also a data URI which is fine (although not my preference).

However, the language on the issuer logo has to be removed since this is unrelated. See my comment.

tlodderstedt (approved):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: aaaa good catch. I added this in the wrong place.

Sakurann: you are right. this was born out of Paul's comment on this PR, and should be in another PR. i removed this from this PR and moved to PR #170

jogu (approved):

awoie (approved):

Issue #103, opened 2023-11-06
Should the parameter `wallet_issuer` be changed to `wallet_provider`?


peppelinux:
WDYT about?

`wallet_issuer` -> `wallet_provider`

it seems more closer to the current language

_Originally posted by @peppelinux in https://github.com/openid/OpenID4VCI/pull/98#discussion_r1382582945_

peppelinux:
In the PR https://github.com/openid/OpenID4VCI/pull/142 I've proposed the parameter name `wallet_id` to uniquely identifying a wallet instance.

I correct myself by saying that the term Wallet Provider has a mean different from the Wallet instance.

The name `wallet_id` is more flexible and it means exactly a wallet identifier. This can be used for doing the discovery process defined in SIOPv2. Nothing prevents a wallet instance to provide its wallet provider unique identifier in the wallet_id value, if the wallet capabilities and endpoints are provided in the cloud by the wallet provider.

selfissued:
There's also a discussion in #142 about whether we need a wallet identifier at all distinct from the Client ID.

PR #142, opened 2023-12-09
fix!: Authz request parameter wallet_issuer renamed to wallet_id

peppelinux:
This PR:

- resolves https://github.com/openid/OpenID4VCI/issues/103, renaming `wallet_issuer` to `wallet_id`. The term Wallet Issuer may sound misleading since it may be confused with Wallet Provider that's a consolidated term within the eIDAS Wallet ecosystem
- aligns the terminology about End-User, in particular in the sequence diagrams
- removes the mention of OAuth 2.0 AS in the definition of the term Wallet, since in this specs the AS for authorizing the issuance of a credential is elsewhere from the Wallet that acts like a client. Another reason for this removal is that in the SIOPv2 specs the wallet instance acts as an OAuth 2.0 AS, while in this specification the wallet instance acts like an OAuth 2.0 Client.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -103,7 +103,7 @@ Biometrics-based Holder Binding:
 :  Ability of the Holder to prove legitimate possession of a Verifiable Credential by demonstrating a certain biometric trait, such as fingerprint or face. One example of a Verifiable Credential with Biometrics-based Holder Binding is a mobile driving license [@ISO.18013-5], which contains a portrait of the holder.
 
 Wallet:
-:  An entity used by the Holder to request, receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored and managed locally, or by using a remote self-hosted service, or a remote third-party service.
+:  An entity used by the Holder to request, receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored and managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, during Credential Issuance the Wallet acts as an OAuth 2.0 Client (see [@!RFC6749]) towards the Credential Issuer which acts as the OAuth 2.0 Authorization Server / Resource Server.
 
 Deferred Credential Issuance:
 :  Issuance of Credentials not directly in the response to a Credential issuance request but following a period of time that can be used to perform certain offline business processes.
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -103,7 +103,7 @@ Biometrics-based Holder Binding:
 :  Ability of the Holder to prove legitimate possession of a Verifiable Credential by demonstrating a certain biometric trait, such as fingerprint or face. One example of a Verifiable Credential with Biometrics-based Holder Binding is a mobile driving license [@ISO.18013-5], which contains a portrait of the holder.
 
 Wallet:
-:  An entity used by the Holder to receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored/managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Authorization Server (see [@!RFC6749]) towards the Credential Verifier which acts as the OAuth 2.0 Client.
+:  An entity used by the Holder to request, receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored and managed locally, or by using a remote self-hosted service, or a remote third-party service.
 
 Deferred Credential Issuance:
 :  Issuance of Credentials not directly in the response to a Credential issuance request but following a period of time that can be used to perform certain offline business processes.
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -103,7 +103,7 @@ Biometrics-based Holder Binding:
 :  Ability of the Holder to prove legitimate possession of a Verifiable Credential by demonstrating a certain biometric trait, such as fingerprint or face. One example of a Verifiable Credential with Biometrics-based Holder Binding is a mobile driving license [@ISO.18013-5], which contains a portrait of the holder.
 
 Wallet:
-:  An entity used by the Holder to receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored/managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Authorization Server (see [@!RFC6749]) towards the Credential Verifier which acts as the OAuth 2.0 Client.
+:  An entity used by the Holder to request, receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored and managed locally, or by using a remote self-hosted service, or a remote third-party service.
```
Sakurann, line 106:
```suggestion
:  An entity used by the Holder to request, receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored and managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, during Credential Issuance the Wallet acts as an OAuth 2.0 Client (see [@!RFC6749]) towards the Credential Issuer which acts as the OAuth 2.0 Authorization Server / Resource Server.
```
I am not sure we should remove this sentence entirely, as opposed to making it accurate.

paulbastian:
My colleagues and I never really understood the parameters wallet_issuer and user_hint and this PR doesn't solve that either. I think the specification is missing text on this, but unsure if this should be a target of this PR

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: I fully agree your rewording @Sakurann, now the text reflect precisely the context,  the roles and the scope of this specs. Text picked!

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -103,7 +103,7 @@ Biometrics-based Holder Binding:
...
 
 Wallet:
-:  An entity used by the Holder to receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored/managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Authorization Server (see [@!RFC6749]) towards the Credential Verifier which acts as the OAuth 2.0 Client.
+:  An entity used by the Holder to request, receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored and managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, during Credential Issuance the Wallet acts as an OAuth 2.0 Client (see [@!RFC6749]) towards the Credential Issuer which acts as the OAuth 2.0 Authorization Server / Resource Server.
```
peppelinux, line 106:
```suggestion
:  An entity used by the Holder to request, receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored and managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Client (see [@!RFC6749]) since the Credential Issuer serves as the OAuth 2.0 Resource Server and, in certain implementations, it may also perform the role of the OAuth 2.0 Authorization Server. 
```

peppelinux: @Sakurann WDYT?

peppelinux:
> My colleagues and I never really understood the parameters wallet_issuer and user_hint and this PR doesn't solve that either. I think the specification is missing text on this, but unsure if this should be a target of this PR

yes, your concerns are understandable and I have moved your comment in the following issue:

https://github.com/openid/OpenID4VCI/issues/145

tplooker:
With regard to the change around wallet_issuer -> wallet_provider, I still don't see a need to have an alternative identifier for the wallet. See my comment in issue #145, but I think the client_id should suffice.

peppelinux:
> With regard to the change around wallet_issuer -> wallet_provider, I still don't see a need to have an alternative identifier for the wallet. See my comment in issue #145, but I think the client_id should suffice.

it make sense, if I got it well, you're suggesting to replace `wallet_issuer` with `client_id`?
if yes, I agree and I can change this part of the text in this PR accordingly

Sakurann:
if we want to revisit the fact if `wallet_issuer` parameter itself is needed, I don't think this PR need to go in before ID-1.

tplooker:
> if we want to revisit the fact if `wallet_issuer` parameter itself is needed, I don't think this PR need to go in before ID-1.

Agreed. I think we should revisit this.

awoie:
> > if we want to revisit the fact if `wallet_issuer` parameter itself is needed, I don't think this PR need to go in before ID-1.
> 
> Agreed. I think we should revisit this.

I think it would be great if we could reuse `client_id` for this because this is what the wallet effectively is -> an OAuth2 client to the AS that wants to get access to a Credential Issuer.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -134,12 +134,12 @@ Existing OAuth 2.0 mechanisms are extended as following:
...
 * Client metadata is used to convey Wallet's metadata. A new metadata parameter `credential_offer_endpoint` is added to allow a Wallet (acting as OAuth 2.0 client) to publish its Credential Offer Endpoint (see (#client-metadata)).
-* Authorization Endpoint: An additional parameter `issuer_state` is added to convey state in the context of processing an issuer-initiated Credential Offer (see (#credential-authz-request)). Additional parameters `wallet_issuer` and `user_hint` are added to enable the Credential Issuer to request Verifiable Presentations from the calling Wallet during Authorization Request processing.
+* Authorization Endpoint: An additional parameter `issuer_state` is added to convey state in the context of processing an issuer-initiated Credential Offer (see (#credential-authz-request)). Additional parameters `wallet_id` and `user_hint` are added to enable the Credential Issuer to request Verifiable Presentations from the calling Wallet during Authorization Request processing.
```
peppelinux, line 137:
```suggestion
* Authorization Endpoint: An additional parameter `issuer_state` is added to convey state in the context of processing an issuer-initiated Credential Offer (see (#credential-authz-request)). In conjunction with `client_id`, the additional parameter `user_hint` enables the Credential Issuer to request Verifiable Presentations from the calling Wallet during the Authorization Request processing.
```

selfissued:
It seems to me that we should discuss whether to have a wallet identifier at all that is distinct from the Client ID before the Implementer's Draft, as it would result in normative changes.

It might be more appropriate to discuss this in issue #103 than this PR, because depending on the outcome of the discussion, this PR might not be used at all.

Sakurann:
closing in a week unless strong objections, since there seems to be a consensus that we should go back to the issue discussion whether wallet_issuer is needed at all.

PR #153, opened 2023-12-14
group `credential_response_encryption` parameters in the Issuer Metadata

### LINKED ISSUE #152: Merge Credential Response encryption Issuer Metadata
> > On the Credential request parameters, the logic between the presence of `jwk`, `enc`, and `alg` sounds weird. My suggestion would be to make all three mandatory, and remove the default value for `enc`.
> > On the Issuer metadata.... I don't think there is a need to group there, but we could either
> > 
> > 1. move `require_credential_response_encryption` before `credential_response_encryption_alg_values_supported` and `credential_response_encryption_enc_values_supported` to make the order more natural, or
> > 2. remove `require_credential_response_encryption` and say that the presence of both `credential_response_encryption_alg_values_supported` and `credential_response_encryption_enc_values_supported` (not either) means that "Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request".
> 
> We cannot make all three keys REQUIRED since encryption is OPTIONAL. If we group them, then we can make the group key OPTIONAL and then we probably don't have to say anything about the sub-keys (jwk, enc, alg) anymore.

I think I also tend to favor grouping for Issuer metadata.

_Originally posted by @paulbastian in https://github.com/openid/OpenID4VCI/issues/136#issuecomment-1849659868_

paulbastian:
Closes #152 

I want to note that we don't have an example for the usage of this parameter in the issuer metadata yet.
Should we add an example?


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1247,9 +1247,9 @@ This specification defines the following Credential Issuer Metadata:
...
+* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer requires encryption of the (Batch) Credential Response. If this request element is not present, the (Batch) Credential Response returned should not be encrypted.
+  * `alg_values_supported`: REQUIRED. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`alg` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
+  * `enc_values_supported`: REQUIRED. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
```
peppelinux, line 1252: I prefer the traditional pattern $whatever`_values_supported` since it was consolidated though years in OAuth 2.0 and OpenID Connect specs

paulbastian: I don't understand. Are you voting to reject the PR?

peppelinux:
@paulbastian yes an example would improve this PR

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1247,9 +1247,9 @@ This specification defines the following Credential Issuer Metadata:
...
-* `credential_response_encryption_enc_values_supported`: OPTIONAL. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
-* `require_credential_response_encryption`: OPTIONAL. Boolean value specifying whether the Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request, with `true` indicating support. When the value is `true`, `credential_response_encryption_alg_values_supported` parameter MUST also be provided.  If omitted, the default value is `false`.
+* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer requires encryption of the (Batch) Credential Response. If this request element is not present, the (Batch) Credential Response returned should not be encrypted.
```
Sakurann, line 1250: I think the current description is not sufficient. please expand/clarify.

```suggestion
* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer requires encryption of the (Batch) Credential Response on top of TLS. If this object is present, the Credential Issuer expects (Batch) Credential Response to be encrypted. If this object is omitted, (Batch) Credential Response should not be encrypted.
```

Sakurann (changes requested):
i don't think the current description is sufficient, and would like to see it expanded. having an example would be nice, but not mandatory to merge this PR IMO

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1247,9 +1247,9 @@ This specification defines the following Credential Issuer Metadata:
...
-* `credential_response_encryption_enc_values_supported`: OPTIONAL. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
-* `require_credential_response_encryption`: OPTIONAL. Boolean value specifying whether the Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request, with `true` indicating support. When the value is `true`, `credential_response_encryption_alg_values_supported` parameter MUST also be provided.  If omitted, the default value is `false`.
+* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer requires encryption of the (Batch) Credential Response on top of TLS. If this object is present, the Credential Issuer expects Credential and/or Batch Credential Response to be encrypted. If this object is omitted, Credential  and/or Batch Credential Response should not be encrypted.
```
paulbastian, line 1250:
```suggestion
* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer requires encryption of the Credential and Batch Credential Response on top of TLS. If this object is present, the Credential Issuer expects Credential and Batch Credential Response to be encrypted. If this object is omitted, Credential  and Batch Credential Response should not be encrypted.
```

Sakurann: please also add something about "Object containing... that consists of the following parameters"

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1247,9 +1247,9 @@ This specification defines the following Credential Issuer Metadata:
...
-* `credential_response_encryption_enc_values_supported`: OPTIONAL. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
-* `require_credential_response_encryption`: OPTIONAL. Boolean value specifying whether the Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request, with `true` indicating support. When the value is `true`, `credential_response_encryption_alg_values_supported` parameter MUST also be provided.  If omitted, the default value is `false`.
+* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer requires encryption of the Credential and Batch Credential Response on top of TLS. If this object is present, the Credential Issuer expects Credential and Batch Credential Response to be encrypted. If this object is omitted, Credential  and Batch Credential Response should not be encrypted.
```
paulbastian, line 1250:
```suggestion
* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer requires encryption of the Credential and Batch Credential Response on top of TLS. If this object is present, the Credential Issuer expects Credential and Batch Credential Response to be encrypted. If this object is omitted, Credential  and Batch Credential Response should not be encrypted. If present, the object consists of exactly two parameters:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -766,7 +766,7 @@ Authorization: BEARER czZCaGRSa3F0MzpnWDFmQmF0M2JW
 
 The Client MAY request encrypted responses by providing its encryption parameters in the Credential Request.
 
-The Credential Issuer MAY require encrypted responses by including `require_credential_response_encryption` in the Credntial Issuer Metadata.
+The Credential Issuer MAY require encrypted responses by including `credential_response_encryption` object in the Credential Issuer Metadata.
```
awoie, line 769:
```suggestion
The Credential Issuer MAY require encrypted responses by including `credential_response_encryption` in the Credential Issuer Metadata.
```

awoie: If we need to refer to this, we should use the word "parameter" as in Issuer Metadata parameter, Authorization Server metadata parameter, etc. "metadata values" might be also fine.

openid-4-verifiable-credential-issuance-1_0.md
awoie, line 1250:
```suggestion
* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer requires encryption of the Credential and Batch Credential Response on top of TLS. If this parameter is present, the Credential Issuer expects Credential and Batch Credential Response to be encrypted. If this parameter is omitted, Credential  and Batch Credential Response should not be encrypted.
```

awoie: See above the reasoning.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1247,9 +1247,9 @@ This specification defines the following Credential Issuer Metadata:
...
+* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer requires encryption of the Credential and Batch Credential Response on top of TLS. If this object is present, the Credential Issuer expects Credential and Batch Credential Response to be encrypted. If this object is omitted, Credential  and Batch Credential Response should not be encrypted.
+  * `alg_values_supported`: REQUIRED. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`alg` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
+  * `enc_values_supported`: REQUIRED. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
```
awoie, line 1252: We should probably add something on supported curves. Something like the following ...

```suggestion
  * `enc_values_supported`: REQUIRED. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
  * `jwk_values_supported`: REQUIRED. Array containing a list of objects describing the supported JWK values by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].  Each object MUST contain a (`kty` value) [@!RFC7517] and an OPTIONAL (`crv` value) [@!RFC8037]. For example, to indicate the issuer supports ECDH-ES with curve P-256, `jwk_values_supported` includes the following object: `{ "kty":"EC", "crv":"P-256" }`.
```

awoie: Should we add this info about public key types and curves. Otherwise, wallets won't know which public key type they have to provide.

examples/credential_issuer_metadata_jwt_vc_json.json
```diff
@@ -4,6 +4,14 @@
     "credential_endpoint": "https://credential-issuer.example.com",
     "batch_credential_endpoint": "https://credential-issuer.example.com/batch_credential",
     "deferred_credential_endpoint": "https://credential-issuer.example.com/deferred_credential",
+    "credential_response_encryption": {
+        "alg_values_supported" : [
+            "ECDH-ES"
+        ],
+        "enc_values_supported" : [
+            "A128GCM"
+        ]
```
awoie, line 13:
```suggestion
        "enc_values_supported" : [
            "A128GCM"
        ],
        "jwk_values_supported": [
            { "kty": "EC", "crv": "P-256" },
            { "kty": "EC", "crv": "X25519" },
            { "kty": "RSA" }            
        ],
```

bc-pi (changes requested):
I am in favor of grouping these parameters but I think the case of a credential issuer being able to encrypt a credential response and publishing the supported alg/enc values while not *requiring* the credential response be encrypted should be retained.

awoie (approved):
Generally looks. I made some comments. Please accept them. I'm ok if `jwk_values_supported` does not get accepted but I think it might be useful. Especially with brainpool etc.

examples/credential_issuer_metadata_jwt_vc_json.json
awoie: Only if people believe this is useful. See description of `jwk_values_supported` below.

bc-pi: kty for X25519 is OKP FWIW https://www.rfc-editor.org/rfc/rfc8037.html#section-2

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: This is an interesting question/point, which makes me wonder about other similar places where this kind of info might be useful or needed.

Sakurann:
@bc-pi I don't see anything in the PR that mandates the encryption when these metadata parameters are present.

bc-pi:
> @bc-pi I don't see anything in the PR that mandates the encryption when these metadata parameters are present.

These bits of text, for example, that use the word "require" read that way to me. 

> The Credential Issuer MAY require encrypted responses by including `credential_response_encryption` object in the Credential Issuer Metadata.

> `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer requires encryption of the Credential and Batch Credential Response on top of TLS.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
The Credential Issuer MAY require encrypted responses by including `credential_response_encryption` parameter in the Credential Issuer Metadata.
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: not sure why this is needed..?

Sakurann, line 1250: trying to address Brian's comment regarding the word "required"

```suggestion
* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer supports encryption of the Credential and Batch Credential Response on top of TLS. If this object is present, the Credential Issuer expects Credential and Batch Credential Response to be encrypted. If this object is omitted, Credential and Batch Credential Response should not be encrypted.
```

Sakurann:
@bc-pi I see your point - I made two suggestions to address them. changing "requires" to "supports" made a difference.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: I agree it is an interesting point. might be worth opening an issue and addressing this in a separate PR...

Sakurann (approved):
assuming something like my suggested changes will be accepted

examples/credential_issuer_metadata_jwt_vc_json.json
paulbastian: @selfissued is this also something that we could solve with fully-specified algorithms? To me the solution proposed by Oliver looks rather ugly. Unfortunately Mikes proposal only targets signatures and not yet encyprtion

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: I think you requested this in another conversation in this PR actually.

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: See my point above, is this solvable by Fully Specified Algorithms?

awoie: IANA does not have fully specified JWE algorithms like that. I suggest we create an issue and move my language to a separate PR.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1247,9 +1247,9 @@ This specification defines the following Credential Issuer Metadata:
...
-* `credential_response_encryption_enc_values_supported`: OPTIONAL. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and/or Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
-* `require_credential_response_encryption`: OPTIONAL. Boolean value specifying whether the Credential Issuer requires additional encryption on top of TLS for the Credential Response and expects encryption parameters to be present in the Credential Request and/or Batch Credential Request, with `true` indicating support. When the value is `true`, `credential_response_encryption_alg_values_supported` parameter MUST also be provided.  If omitted, the default value is `false`.
+* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer supports encryption of the Credential and Batch Credential Response on top of TLS. If this object is present, the Credential Issuer expects Credential and Batch Credential Response to be encrypted. If this object is omitted, Credential and Batch Credential Response should not be encrypted.
```
bc-pi, line 1250: The word 'expects' in the latter two sentences reads like a requirement and the sentences don't otherwise add meaningful info.

```suggestion
* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer supports encryption of the Credential and Batch Credential Response on top of TLS. 
```

Sakurann: agreed to remove this sentence and reincarnate the boolean that the issuer can use to indicate that the encryption is required. will potenitally lead to a sentence or two in this parameter description related to the reincarnated boolean. @paulbastian

examples/credential_issuer_metadata_jwt_vc_json.json
Sakurann: @awoie to file a separate issue.

Sakurann:
to clarify, the existence of the encryption related parameters signal that issuer supports the encryption, and the existence of the boolean signals that the issuer requires encryption

openid-4-verifiable-credential-issuance-1_0.md
awoie: Created an issue here #168

examples/credential_issuer_metadata_jwt_vc_json.json
awoie: Done, see #168

selfissued:
I plan to review this after the disposition of the set of suggested changes is known.

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: the phrase is removed in the latest proposal

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: applied in latest commit

paulbastian:
@bc-pi @selfissued please review

examples/credential_issuer_metadata_jwt_vc_json.json
jogu, line 13: `encryption_required` is marked as 'required' in the spec so should be present here

```suggestion
        ],
        "encryption_required": false
```

jogu (changes requested):

examples/credential_issuer_metadata_jwt_vc_json.json
paulbastian: thanks Joseph

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1335,9 +1335,10 @@ This specification defines the following Credential Issuer Metadata:
...
+* `credential_response_encryption`: OPTIONAL. Object containing information whether the Credential Issuer supports encryption of the Credential and Batch Credential Response on top of TLS.
+  * `alg_values_supported`: REQUIRED. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`alg` values) [@!RFC7518] supported by the Credential and Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
+  * `enc_values_supported`: REQUIRED. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential and Batch Credential Endpoint to encode the Credential or Batch Credential Response in a JWT [@!RFC7519].
+  * `encryption_required`: REQUIRED. Boolean value specifying whether the Credential Issuer requires the additional encryption on top of TLS for the Credential Response. If the value is `true`, the Credential Issuer requires the encryption for every Credential Response and therefore the Wallet MUST provide encryption keys in the Credential Request. If the value is `false`, the Wallet MAY chose whether it provides encryption keys or not.
```
Sakurann, line 1341: I think it logically makes more sense to bring this parameter before `enc_values_supported` and `alg_values_supported` when describing?

jogu (approved):

Sakurann:
DCP WG call - Jan-04-2024 agreed to merge once Brian's approval is in. agreed `encryption_required` is a good trade-off as a parameter name.

bc-pi (approved):

bc-pi:
Brian's approval is in

Issue #158, opened 2023-12-18
Replace mentions of PIN with new naming Transaction Code


cryptphil:
The naming "PIN" was replaced with "Transaction Code" with PR #95 but there were still two locations where a PIN was mentioned.

In concrete, the PIN is still visible in the pre-authorized code flow figure in [Step 4](https://github.com/openid/OpenID4VCI/blob/main/openid-4-verifiable-credential-issuance-1_0.md?plain=1#L272) and in the [example](https://github.com/openid/OpenID4VCI/blob/main/examples/credential_offer_multiple_credentials.json#L13) for a credential offer for multiple credentials.

Sakurann:
PR merged

PR #159, opened 2023-12-18
Fix: Replace mentions of PIN to Transaction Code

cryptphil:
Solves #158.

### DIFF
--- a/examples/credential_offer_multiple_credentials.json
+++ b/examples/credential_offer_multiple_credentials.json
@@ -10,7 +10,7 @@
          "tx_code" : {
             "length": 4,
             "input_mode" : "numeric",
-            "description": "Please provide the PIN which was sent via e-mail"
+            "description": "Please provide the Transaction Code which was sent via e-mail"
          }
       }
    }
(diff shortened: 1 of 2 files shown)


examples/credential_offer_multiple_credentials.json
```diff
@@ -10,7 +10,7 @@
          "tx_code" : {
             "length": 4,
             "input_mode" : "numeric",
-            "description": "Please provide the PIN which was sent via e-mail"
+            "description": "Please provide the Transaction Code which was sent via e-mail"
```
paulbastian, line 13: The Issuer doesn't necessarily need to call it Transaction Code in his communciation to the Holder, but PIN is probably confusing here. What about OTP?

cryptphil: True, transaction code may sound a bit too technical for end-users, I would also prefer OTP or rather the written word "one-time password".

cryptphil: I changed it to "one-time code", I think that is the clearest notation in this case. What do you think? @Sakurann @paulbastian

Sakurann (approved):

jogu (approved):

awoie (approved):

c2bo (approved):

PR #161, opened 2023-12-20
Add iacas_uri to mso_mdoc issuer metadata

### LINKED ISSUE #62: IACA Metadata for Credential Issuers

awoie:
Added iacas_uri to mso_mdoc issuer metadata. Fixes #62


selfissued (approved):

paulbastian (changes requested):
The PR changes data in the authorization details data, I think this is not the correct way, either these things belong into Issuer metadata or the mdoc itself.

peppelinux:
> The PR changes data in the authorization details data, I think this is not the correct way, either these things belong into Issuer metadata or the mdoc itself.

I'm in favor of having `iacas_uri` in the issuer metadata. 
I'm not in favor of having it in the mdoc cbor, since it would be too much self-referential for a single artifact.

jogu:
I think we should close this for now and continue discussion on the issue - there's unanswered questions on the issue and not consensus about why this is needed.

awoie:
> The PR changes data in the authorization details data, I think this is not the correct way, either these things belong into Issuer metadata or the mdoc itself.

In the mdoc, this is not possible because not allowed by ISO 18013-5. This PR adds the info to the issuer metadata, doesn't it?

Sakurann:
closing for now, to consolidate the discussion to the issue, so that we can agree on the direction.

PR #164, opened 2023-12-21
fix error in the wallet definition

### LINKED ISSUE #156: Wallet and VCI are not OAuth 2.0 Authorization Servers
In the current specifications, it's stated that the "Wallet acts as an OAuth 2.0 Authorization Server (AS)." It could be beneficial to review the usage and context of the term AS throughout the document.

From the perspective of the Verifiable Credential Issuer (VCI) specifications, a Credential Issuer technically functions as a Resource Server (RS) that may or may not incorporate an AS, as it could depend on an external AS. This RS only requires a valid Access Token to provide the protected resources to a consumer (Client).

However, from the Verifier's viewpoint, the Wallet acts more like an identity provider, supplying verifiable data (Verifiable Credentials) about its owner on their behalf. The interactions between a Verifier and a Wallet, while conceptually similar, are technically distinct considering the flows and artifacts defined in SIOPv2 and OpenID4VP, the wallet metadata, etc. These extend, modify, and replace several aspects defined in RFC6749, resulting in similarities between the profiles that, however, may be superficial or in apparence. 

For instance, from the Verifier perspective with OpenID4VP, the flows: auth code flow, implicit, hybrid ... Or the token endpoint, something at the base of RFC6749 are now removed and not mentioned, since not needed. The OpenID4VC specs are actually different, even if not so far from OAuth 2.0 in a way to be still considered within its ecosystem but in a specific place with specialized term, flows, metadata ... A completely new paradigm in a completely new model, so many things, similar but different.

side note: The resource owner of the protected resource (VCI) is the user. I love to think that the credential endpoint is something similar to the OIDC userinfo endpoint, and all this makes a VCI more similar to an OpenID Provider than to a AS. Considering that OpenID deals with digital identities of persons.

If there's room for this kind of analysis, we might consider producing a PR to provide more clarity on these fundamental aspects of the roles within the ecosystem and their interactions.

this issue has born here: 
- https://github.com/openid/OpenID4VCI/pull/155#discussion_r1428007812
- https://github.com/openid/OpenID4VCI/pull/155/files#r1428013982

Sakurann:
resolves #156

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -103,7 +103,7 @@ Biometrics-based Holder Binding:
 :  Ability of the Holder to prove legitimate possession of a Verifiable Credential by demonstrating a certain biometric trait, such as fingerprint or face. One example of a Verifiable Credential with Biometrics-based Holder Binding is a mobile driving license [@ISO.18013-5], which contains a portrait of the holder.
 
 Wallet:
-:  An entity used by the Holder to receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored/managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Client towards the Credential Issuer which acts as the OAuth 2.0 Authorization Server.
+:  An entity used by the Holder to request, receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored and managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Client (see [@!RFC6749]) since the Credential Issuer is the OAuth 2.0 Resource Server and, in certain cases, it may also implement an OAuth 2.0 Authorization Server.
 
 Deferred Credential Issuance:
 :  Issuance of Credentials not directly in the response to a Credential issuance request but following a period of time that can be used to perform certain offline business processes.
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -103,7 +103,7 @@ Biometrics-based Holder Binding:
 :  Ability of the Holder to prove legitimate possession of a Verifiable Credential by demonstrating a certain biometric trait, such as fingerprint or face. One example of a Verifiable Credential with Biometrics-based Holder Binding is a mobile driving license [@ISO.18013-5], which contains a portrait of the holder.
 
 Wallet:
-:  An entity used by the Holder to receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored/managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Authorization Server (see [@!RFC6749]) towards the Credential Verifier which acts as the OAuth 2.0 Client.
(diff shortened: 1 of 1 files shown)


danielfett (approved):

jogu (approved):

awoie (approved):

paulbastian (approved):

c2bo (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -103,7 +103,7 @@ Biometrics-based Holder Binding:
...
 
 Wallet:
-:  An entity used by the Holder to receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored/managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Authorization Server (see [@!RFC6749]) towards the Credential Verifier which acts as the OAuth 2.0 Client.
+:  An entity used by the Holder to receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored/managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Client towards the Credential Issuer which acts as the OAuth 2.0 Authorization Server.
```
peppelinux, line 106:
```suggestion
:  An entity used by the Holder to request, receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored and managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Client (see [@!RFC6749]) since the Credential Issuer is the OAuth 2.0 Resource Server and, in certain cases, it may also implement an OAuth 2.0 Authorization Server.
```

peppelinux (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: Here I want to remark that VCI is an RS that can optionally implement and include also an AS.
I'd not force the idea that VCI is a monolithic solution with AS and RS together. That would be only an implementative option.

c2bo: The RFC is mentioned above this part in the beginning of the terminology section (introducing all terms according to 6749), so I don't think we need to mention it here?

tlodderstedt (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -103,7 +103,7 @@ Biometrics-based Holder Binding:
...
 
 Wallet:
-:  An entity used by the Holder to receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored/managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Authorization Server (see [@!RFC6749]) towards the Credential Verifier which acts as the OAuth 2.0 Client.
+:  An entity used by the Holder to request, receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored and managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Client (see [@!RFC6749]) since the Credential Issuer is the OAuth 2.0 Resource Server and, in certain cases, it may also implement an OAuth 2.0 Authorization Server.
```
Sakurann, line 106:
```suggestion
:  An entity used by the Holder to request, receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored and managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Client since the Credential Issuer is the OAuth 2.0 Resource Server and, in certain cases, it may also implement an OAuth 2.0 Authorization Server.
```

Sakurann:
will merge once @peppelinux approves.

tplooker (approved):

peppelinux (approved):

peppelinux:
sorry for the late!

PR #170, opened 2023-12-22
adding credential issuer logo parameter in the issuer metadata

Sakurann:
PR born out of this comment on PR https://github.com/openid/OpenID4VCI/pull/141#issuecomment-1855644509


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1342,6 +1342,9 @@ This specification defines the following Credential Issuer Metadata:
 * `display`: OPTIONAL. Array of objects, where each object contains display properties of a Credential Issuer for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
   * `name`: OPTIONAL. String value of a display name for the Credential Issuer.
   * `locale`: OPTIONAL. String value that identifies the language of this object represented as a language tag taken from values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
+  * `logo`: OPTIONAL. A JSON object with information about the logo of the Credential Issuer with a following non-exhaustive list of parameters that MAY be included:
```
c2bo, line 1345:
```suggestion
  * `logo`: OPTIONAL. Object with information about the logo of the Credential Issuer with a following non-exhaustive list of parameters that MAY be included:
```

paulbastian (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1342,6 +1342,9 @@ This specification defines the following Credential Issuer Metadata:
...
+    * `uri`: OPTIONAL. String value that contains a URI where the Wallet can obtain a logo of the Credential from the Credential Issuer. Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
```
c2bo, line 1346:
```suggestion
    * `uri`: OPTIONAL. String value that contains a URI where the Wallet can obtain a logo of the Credential Issuer. Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
```

peppelinux, line 1346:
```suggestion
    * `uri`: REQUIRED when the `logo` object is present. String value that contains a URI where the Wallet can obtain a logo of the Credential from the Credential Issuer. Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
```

If the logo object is present but doesn't contain any properties (like uri), it would indeed result in "logo": {}. This would be nonsensical because the logo object is supposed to contain at least the uri property when it is present. Therefore, it's important to specify that uri is required when logo is present to avoid such situations.

peppelinux (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
selfissued, line 1345:
```suggestion
  * `logo`: OPTIONAL. A JSON object with information about the logo of the Credential Issuer with the following non-exhaustive list of parameters that MAY be included:
```

openid-4-verifiable-credential-issuance-1_0.md
selfissued, line 1346:
```suggestion
    * `uri`: OPTIONAL. String value that contains a URI where the Wallet can obtain a logo of the Credential from the Credential Issuer. The Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
  * `logo`: OPTIONAL. Object with information about the logo of the Credential Issuer with the following non-exhaustive list of parameters that MAY be included:
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: we should discuss this. `logo` object is optional. so the issuer can avoid "logo": {} by not including `logo` parameter at all.

Sakurann:
```suggestion
    * `uri`: OPTIONAL. String value that contains a URI where the Wallet can obtain a logo of the Credential Issuer. The Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
```

jogu: I think it's also about setting clear expectations for what we expect issuers to do, as well as making it clear what situations we might end up needing to conformance test that wallets correctly parse. So I think it's good to make it clear that we don't expect issuers to use `"logo": {}` if that is our expectation.

It would probably be clearer if the logo object was defined in a new section. Then `uri` could I think just be marked as 'REQUIRED' and I think it would be clear that means "if the logo object is present".

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1342,6 +1342,9 @@ This specification defines the following Credential Issuer Metadata:
 * `display`: OPTIONAL. Array of objects, where each object contains display properties of a Credential Issuer for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
   * `name`: OPTIONAL. String value of a display name for the Credential Issuer.
   * `locale`: OPTIONAL. String value that identifies the language of this object represented as a language tag taken from values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
+  * `logo`: OPTIONAL. Object with information about the logo of the Credential Issuer with the following non-exhaustive list of parameters that MAY be included:
+    * `uri`: OPTIONAL. String value that contains a URI where the Wallet can obtain a logo of the Credential Issuer. The Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
```
Sakurann, line 1346:
```suggestion
    * `uri`: REQURED. String value that contains a URI where the Wallet can obtain a logo of the Credential Issuer. The Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: I don't think a separate section is necessary. I changed `uri` parameter to `required`. it made more sense from the expectations perspective and from the convention of the spec perspective.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1342,6 +1342,9 @@ This specification defines the following Credential Issuer Metadata:
 * `display`: OPTIONAL. Array of objects, where each object contains display properties of a Credential Issuer for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
   * `name`: OPTIONAL. String value of a display name for the Credential Issuer.
   * `locale`: OPTIONAL. String value that identifies the language of this object represented as a language tag taken from values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
+  * `logo`: OPTIONAL. Object with information about the logo of the Credential Issuer with the following non-exhaustive list of parameters that MAY be included:
+    * `uri`: REQURED. String value that contains a URI where the Wallet can obtain a logo of the Credential Issuer. The Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
```
jogu, line 1346: Works for me - unfortunately there's a typo in 'requred':

```suggestion
    * `uri`: REQUIRED. String value that contains a URI where the Wallet can obtain a logo of the Credential Issuer. The Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
```

jogu (approved):

Sakurann:
Jan-04-2024 WG call. agreed on "required" terminology. agreed to change uri to required for credential_configurations_supported too.

peppelinux (approved):
fully approved

sudeshrshetty (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1343,6 +1343,9 @@ This specification defines the following Credential Issuer Metadata:
 * `display`: OPTIONAL. Array of objects, where each object contains display properties of a Credential Issuer for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
   * `name`: OPTIONAL. String value of a display name for the Credential Issuer.
   * `locale`: OPTIONAL. String value that identifies the language of this object represented as a language tag taken from values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
+  * `logo`: OPTIONAL. Object with information about the logo of the Credential Issuer with the following non-exhaustive list of parameters that MAY be included:
+    * `uri`: REQUIRED. String value that contains a URI where the Wallet can obtain a logo of the Credential Issuer. The Wallet needs to determine the scheme, since the URI value could use `https:` scheme, `data:` scheme, etc.
```
tplooker, line 1347: As I've raised in other forums before, I'm concerned at the shear number of options this definition permits, both in by value and by reference based resolution of the image based logo, it also allows an open ended number of image formats. Good standards make choices and this definition is avoiding making a choice which could meaningfully improve interoperability.
 
I also have concerns from a security perspective because many of these URI mechanisms can resolve to much more than just images and because we aren't limiting to certain image formats its likely implementations will leave themselves vulnerable in certain ways such as via RCE through mistakenly downloading and executing something like javascript. At an absolute minimum I'd like to see normative guidance which recommends implementations take precautions in the URI resolution process to improve the likelihood that they actually obtain an image as the result. That might include validating the mediatype of a data URI to check it is indicating its an image OR using headers when resolving an HTTPS based URI to request an image and validate a response.

jogu: I'd suggest raising a separate issue about this, or perhaps two issues (one about the normative security note, one to discuss if we should limit the options).

paulbastian: I agree with Tobias, wallets already have many responsibilities and options to implement, limiting the logo formats is definitely a good idea

Sakurann: @tplooker I think we agreed that we need to discuss limiting the logo formats, but that should not block this PR. Please open a separate issue on this topic as also agreed in #141.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1343,6 +1343,9 @@ This specification defines the following Credential Issuer Metadata:
...
+    * `alt_text`: OPTIONAL. String value of an alternative text of a logo image.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
  * `logo`: OPTIONAL. Object with information about the logo of the Credential Issuer. Below is a non-exhaustive list of parameters that MAY be included:
    * `uri`: REQUIRED. String value that contains a URI where the Wallet can obtain the logo of the Credential Issuer. The Wallet needs to determine the scheme, since the URI value could use the `https:` scheme, the `data:` scheme, etc.
    * `alt_text`: OPTIONAL. String value of the alternative text for the logo image.
```

Sakurann:
@selfissued accepted. please re-review

selfissued (approved):

c2bo (approved):

openid-4-verifiable-credential-issuance-1_0.md
tplooker: Opened in #205.

tplooker (approved):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: thank you!

PR #174, opened 2023-12-27
Results from running a spelling and grammar check

selfissued:
Applied corrections identified by running the HTML through Word's spelling and grammar checker.


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -603,14 +603,14 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
 
 Token Responses are made as defined in [@!RFC6749].
 
-The AS might decide to authorize issuance of multiple instances for each Credential requested in the Authorization Request. Each Credential instance is described using the same entry in the `credential_configurations_supported` Credential Issuer metadata, but contains different claim values or different subset of claims within the claimset identified by the Credential description.
+The AS might decide to authorize issuance of multiple instances for each Credential requested in the Authorization Request. Each Credential instance is described using the same entry in the `credential_configurations_supported` Credential Issuer metadata, but contains different claim values or different subset of claims within the Claims Set identified by the Credential description.
```
jogu, line 606: I'm not convinced "Claims Set" should be capitalised. It's certainly not a term that seems to be defined within this specification. (And you've not changed it, but I don't think 'Credential description' is correct here either, this is the only instance of 'Credential description' in the whole spec.)

selfissued: "JWT Claims Set" is a defined term from https://www.rfc-editor.org/rfc/rfc7519.html#section-2.  We use a lot of terms from other specs without redefining them here.  I mainly caught this because "claimset" is not a word or a defined term.

I'd be OK with this becoming any of "Claims Set", "JWT Claims Set", or "claims set", but it can't stay as-is.

jogu: Agree it shouldn't stay as it is. "JWT Claims Set" would be wrong I think as the credential may not be a JWT. "claims set" probably seems best to me.

I suggest replacing Credential description with `credential_configuration_id` too.

```suggestion
The AS might decide to authorize issuance of multiple instances for each Credential requested in the Authorization Request. Each Credential instance is described using the same entry in the `credential_configurations_supported` Credential Issuer metadata, but contains different claim values or different subset of claims within the claims set identified by the `credential_configuration_id`.
```

peppelinux (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -603,14 +603,14 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
...
+  * `credential_identifiers`: OPTIONAL. Array of strings, each uniquely identifying a Credential that can be issued using the Access Token returned in this response. Each of these Credentials corresponds to the same entry in the `credential_configurations_supported` Credential Issuer metadata but can contain different claim values or a different subset of claims within the Claims Set identified by that Credential type. This parameter can be used to simplify the Credential Request, as defined in (#credential_request), where the `credential_identifier` parameter replaces the `format` parameter and any other Credential format-specific parameters in the Credential Request. When received, the Wallet MUST use these values together with an Access Token in subsequent Credential Requests.
```
selfissued, line 613:
```suggestion
  * `credential_identifiers`: OPTIONAL. Array of strings, each uniquely identifying a Credential that can be issued using the Access Token returned in this response. Each of these Credentials corresponds to the same entry in the `credential_configurations_supported` Credential Issuer metadata but can contain different claim values or a different subset of claims within the claims set identified by that Credential type. This parameter can be used to simplify the Credential Request, as defined in (#credential_request), where the `credential_identifier` parameter replaces the `format` parameter and any other Credential format-specific parameters in the Credential Request. When received, the Wallet MUST use these values together with an Access Token in subsequent Credential Requests.
```

jogu (approved):

Sakurann (approved):
look ok

bc-pi (approved):

tplooker (approved):

PR #177, opened 2023-12-28
Update references in preparation for ID1

selfissued:
Updates references and the uses of them

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1991,7 +1991,7 @@ This section defines Credential Format Profiles for a few of the commonly used C
 
 ## W3C Verifiable Credentials
 
-Sections 6.1 and 6.2 of [@VC_DATA] define how Verifiable Credentials MAY or MAY NOT use JSON-LD [@!JSON-LD]. As acknowledged in Section 4.1 of [@VC_DATA], implementations can behave differently regarding processing of the `@context` property whether JSON-LD is used or not.
+Sections 6.1 and 6.2 of [@VC_DATA] define how Verifiable Credentials MAY or MAY NOT use JSON-LD [@JSON-LD]. As acknowledged in Section 4.1 of [@VC_DATA], implementations can behave differently regarding processing of the `@context` property whether JSON-LD is used or not.
 
 This specification therefore differentiates the following three Credential formats for W3C Verifiable Credentials:
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -52,7 +52,7 @@ Access to this API is authorized using OAuth 2.0 [@!RFC6749], i.e., the Wallet u
 
 ## Requirements Notation and Conventions
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 [@!RFC2119].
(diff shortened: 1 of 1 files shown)


bc-pi (approved):

Sakurann (approved):
looks fine

peppelinux (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -52,7 +52,7 @@ Access to this API is authorized using OAuth 2.0 [@!RFC6749], i.e., the Wallet u
 
 ## Requirements Notation and Conventions
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 [@!RFC2119].
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in BCP 14 [@!RFC2119] [@!RFC8174] when, and only when, they appear in all capitals, as shown here.
```
tplooker, line 55:
```suggestion
The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in BCP 14 [@!RFC2119] and [@!RFC8174] when, and only when, they appear in all capitals, as shown here.
```

tplooker (approved):

openid-4-verifiable-credential-issuance-1_0.md
selfissued: https://www.rfc-editor.org/rfc/rfc8174.html#section-2 contains this text exactly - without the "and".

Are you willing to withdraw this suggestion, @tplooker?

Sakurann: @tplooker already approved, so I assume, yes.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2012,7 +1991,7 @@ This section defines Credential Format Profiles for a few of the commonly used C
 
 ## W3C Verifiable Credentials
 
-Sections 6.1 and 6.2 of [@VC_DATA] define how Verifiable Credentials MAY or MAY NOT use JSON-LD. As acknowledged in Section 4.1 of [@VC_DATA], implementations can behave differently regarding processing of the `@context` property whether JSON-LD is used or not.
+Sections 6.1 and 6.2 of [@VC_DATA] define how Verifiable Credentials MAY or MAY NOT use JSON-LD [@!JSON-LD]. As acknowledged in Section 4.1 of [@VC_DATA], implementations can behave differently regarding processing of the `@context` property whether JSON-LD is used or not.
```
jogu, line 1994: If VC_DATA isn't a normative reference I'm not convinced that JSON-LD is a normative reference?

As JSON-LD is only mentioned in the context of being a normative dependency of VC_DATA, either they're both normative or they're both not?

selfissued: Agreed.  I'll make it non-normative.

selfissued, line 1994:
```suggestion
Sections 6.1 and 6.2 of [@VC_DATA] define how Verifiable Credentials MAY or MAY NOT use JSON-LD [@JSON-LD]. As acknowledged in Section 4.1 of [@VC_DATA], implementations can behave differently regarding processing of the `@context` property whether JSON-LD is used or not.
```

awoie (approved):

jogu (approved):

PR #179, opened 2023-12-29
Proofread Credential Offer Endpoint section

selfissued:
Correct issues identified while proofreading the Credential Offer Endpoint section.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -312,7 +312,7 @@ The Credential Offer object, which is a JSON-encoded object with the Credential 
 
 The Credential Offer contains a single URI query parameter, either `credential_offer` or `credential_offer_uri`:
 
-* `credential_offer`: JSON object with the Credential Offer parameters. This MUST NOT be present when the `credential_offer_uri` parameter is present.
+* `credential_offer`: Object with the Credential Offer parameters. This MUST NOT be present when the `credential_offer_uri` parameter is present.
 * `credential_offer_uri`: String that is a URL using the `https` scheme referencing a resource containing a JSON object with the Credential Offer parameters. This MUST NOT be present when the `credential_offer` parameter is present.
 
 The Credential Issuer MAY render a QR code containing the Credential Offer that can be scanned by the End-User using a Wallet, or a link that the End-User can click.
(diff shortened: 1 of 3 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -302,18 +302,18 @@ For more details and concrete mitigations, see (#security_considerations_pre-aut
...
 The Credential Issuer sends Credential Offer using an HTTP GET request or an HTTP redirect to the Wallet's Credential Offer Endpoint defined in (#client_metadata).
 
 The Credential Offer object, which is a JSON-encoded object with the Credential Offer parameters, can be sent by value or by reference.
 
-The Credential Offer contains a single URI query parameter `credential_offer` or `credential_offer_uri`:
+The Credential Offer contains a single URI query parameter, either `credential_offer` or `credential_offer_uri`:
 
-* `credential_offer`: Object with the Credential Offer parameters. This MUST NOT be present when `credential_offer_uri` parameter is present.
-* `credential_offer_uri`: String that is a URL using the `https` scheme referencing a resource containing a JSON-encoded object with the Credential Offer parameters. This MUST NOT be present when `credential_offer` parameter is present.
+* `credential_offer`: JSON object with the Credential Offer parameters. This MUST NOT be present when the `credential_offer_uri` parameter is present.
```
Sakurann, line 315: we agreed to use Object, not JSON object.

Sakurann (dismissed):
Can we also keep the abbreviation "AS" for the authorization server? latest oauth wg specs/drafts (like security BCP) use "AS" without any issue.

paulbastian (approved):
looks good to me, only the point that Kristina already raised

jogu:
> Can we also keep the abbreviation "AS" for the authorization server? latest oauth wg specs/drafts (like security BCP) use "AS" without any issue.

I'd personally suggest using "Authorization Server". I don't think we define the term "AS" anywhere and even if we did define it, I think it makes the specifications hard to understand for people new to the space.

openid-4-verifiable-credential-issuance-1_0.md
selfissued, line 315:
```suggestion
* `credential_offer`: Object with the Credential Offer parameters. This MUST NOT be present when the `credential_offer_uri` parameter is present.
```

selfissued: I changed it back to Object.

bc-pi (approved):

peppelinux (approved):

tplooker (approved):

awoie (approved):

PR #185, opened 2023-12-30
Proofread Security Considerations

selfissued:
Correct issues identified while proofreading the Security Considerations.


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1387,38 +1387,38 @@ Credential Issuers often want to know what Wallet they are issuing Credentials t
...
...provides apps with a certificate including a certificate chain asserting that a particular key is managed, for example, by a hardware security module [ref]. The Wallet can provide this data along with the proof of possession in the Credential Request (see (#credential_request) for an example) to allow the Credential Issuer to validate the key management policy. This requires the Credential Issuer to rely on the trust anchor of the certificate chain and the respective key management policy. Another variant of this concept is the use of a Qualified Electronic Signature as defined by the eIDAS regulation [ref]. This signature will not reveal the properties of the associated private key to the Credential Issuer. However, as one example, due to the regulatory regime of eIDAS, the Credential Issuer can deduce that the signing service manages the private keys according to this regime and fulfills very high security requirements. As another example, FIDO2 allows RPs to obtain an attestation along with the public key from a FIDO authenticator. That implicitly asserts the key management policy, since the assertion is bound to a certain authenticator model and its key management capabilities.
```
peppelinux, line 1390: Missing [ref]

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1430,20 +1430,19 @@ The Wallet is supposed to detect signs of fraudulent behavior related to the Cre
 
 ## Key Proof replay {#key-proof-replay}
 
-If an adversary is able to get hold of a key proof defined in (#proof_types), the adversary could get a Credential issued that is bound to a key pair controlled by the victim.
+If an adversary is able to get ahold of a key proof, as defined in (#proof_types), the adversary could get a Credential issued that is bound to a key pair controlled by the victim.
```
Sakurann, line 1433: can we please keep `hold` instead of `ahold`?

jogu: It's an awkward area of grammar that differs between USA & British English, and I agree with avoiding 'ahold'. Perhaps just use "obtain" instead?

selfissued, line 1433:
```suggestion
If an adversary is able to obtain a key proof, as defined in (#proof_types), the adversary could get a Credential issued that is bound to a key pair controlled by the victim.
```

selfissued: I applied Joseph's proposed wording.

bc-pi (approved):

peppelinux (approved):

paulbastian (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1431,20 +1431,19 @@ The Wallet is supposed to detect signs of fraudulent behavior related to the Cre
...
 
-Note: For the attacker to be able to present to the Verifier a Credential bound to a replayed Key Proof, the attacker also needs to obtain the victim's private key. To limit this, servers are RECOMMENDED to check how the Wallet protects the private keys, using mechanisms such as Attestation-Based Client Authentication defined in [@!I-D.ietf-oauth-attestation-based-client-auth].
+Note: For the attacker to be able to present a Credential bound to a replayed Key Proof to the Verifier, the attacker also needs to obtain the victim's private key. To limit this, servers are RECOMMENDED to check how the Wallet protects the private keys, using mechanisms such as Attestation-Based Client Authentication as defined in [@!I-D.ietf-oauth-attestation-based-client-auth].
 
-`nonce` parameter is the primary countermeasure against key proof replay. To further narrow down the attack vector, the Credential Issuer SHOULD bind a unique `nonce` parameter to the respective Access Token.
+The `nonce` parameter is the primary countermeasure against key proof replay. To further narrow down the attack vector, the Credential Issuer SHOULD bind a unique `nonce` parameter to the respective Access Token.
```
tplooker, line 1438: Perhaps a digression from the current purpose of this PR, but when re-reading this paragraph in isolation the guidance about a unique nonce parameter to the respective access token seems a bit off. Does this mean to imply we should only have one nonce for key proofs in the credential request per access token? What happens if the access token is valid for a long time?

awoie (approved):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: I think the mechanism of how c_nonce is provided by the issuer makes it pretty clear that it is up to the issuer when to refresh the nonce value independent from the access token lifecycle, but feel free to open an issue. I agree it is probably out of scope of this PR..

Sakurann (approved):

PR #188, opened 2023-12-30
fix: audience set with the url of the endpoint

peppelinux:
This PR restricts the domain of the audience to the specific endpoint where the JWT/CWT is sent.
This prevents that the same proof could be reused across different endpoints, like the credential endpoint and the batch endpoint.

There are cases where AS != VCI, and this is another good reason to restrict the audience to a specific endpoint and not the entire entity domain.


selfissued (commented):
I don't know that I agree with this change.  We tend to use issuer values to identify entities in the OpenID & OAuth worlds - not endpoint addresses.  For instance, we use the OAuth `issuer` identifier in https://www.rfc-editor.org/rfc/rfc8414.html.  Is there a reason to do a different thing here?

peppelinux:
@selfissued I'm aware that this is not an easy decision.

The evidence I bring is taken from some evidences we have in the "traditional" implementations where multiple endpoints, like introspection, revocation, token and any other, can be protected with a client_assertion (AKA: private_key_jwt in the OIDC realm).

considering that a single entity can expose multiple endpoints over multiple nodes in a cluster, behind its FQDN and its httpd frontend, it turns out that if the audience of the client_assertion is set to the entity id (https://FQDN) without any explicit path pointing to a specific endpoint, the signed JWT can then be re-used to other endpoints.

I'm not for breaking the rules and I know that this can have a considerable impact at the interop level. For this reason, to bring awareness and continue to preserve "traditional" practices, we would like to recommend, rather than mandate, using the endpoint instead of the entity id?

Or could we bring new wine into new wineskins for the wallet ecosystem and increase security by saying <<ok, it's time to change, the audience must be set with the endpoint>>?

jogu:
@selfissued 

> I don't know that I agree with this change. We tend to use issuer values to identify entities in the OpenID & OAuth worlds - not endpoint addresses. For instance, we use the OAuth `issuer` identifier in https://www.rfc-editor.org/rfc/rfc8414.html. Is there a reason to do a different thing here?

Whilst I've not come to a conclusion on which way we should go yet, I will note that the JWTs Guiseppe is changing in the PR are being sent to resource server endpoints protected by access tokens (rather than authorization server ones often protected by client authentication), so I don't think there's a notable precedent towards sending issuer identifiers in those. The most widely deployed standards-based JWT going to a resource server endpoint I can currently think of would be DPoP, where the `htu` claim (there is no `aud` claim in DPoP) is the endpoint url.

gffletch:
I think we be very careful about the context and when this might (or might not) be needed. Artifacts like `private_secret_jwt` are about client authentication and in that context jti + one-time-use semantics can be leveraged to prevent replay if needed. Also, the [Resource Indicators for OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc8707) specification allows for binding an access token to a particular resource. Would that not be a better solution? Maybe it would be sufficient to specify that the `aud` claim MAY be set using a resource indicator string?

selfissued:
I agree with George.  Resource indicator strings (which are the same as Resource Identifiers in draft-ietf-oauth-resource-metadata) are the way to go.

Also, the very fact that resources could expose multiple endpoints argues for using the Resource Identifier - not one of the endpoint URLs.

Sakurann:
@peppelinux is there a PR on this? Can we please have a discussion in the issue before doing a PR? Thank you.

Sakurann (changes requested):
I think we need to discuss first what security property we are trying to achieve here. before jumping to the "how" that the audience restriction of a proof JWT is a way to go...

peppelinux:
the security property we are trying to achieve here can be explained with the evidence that an AS can issue tokens for multiple Credential Issuers (RSs) and that multiple Credential Issuers (RSs) can have one or more AS with multiple endpoints as well.

an audience set with the entity id of an AS or an RS can be reused across multiple endpoint.
this PR opened the discussion about the security requirement of having audiences limited to the endpoints, while George and Mike pointed out that resource indicators could be a better solution.

if we confirm the security issue (problem) then we can define the requirement and then the solution. 
I started a quick PR to give some material example, I don't want to break the rules behind the editorial flow :-)

jogu:
@peppelinux can you open an issue that describes the problem you think we need to solve please? I think we could then close this PR until we have a consensus that there is a problem to solve and consensus on how to solve it.

jogu:
Thanks @peppelinux ! Let's close this PR for now and continue the discussion on #265 then.

Issue #172, opened 2023-12-24
Editorial Issues in Appendix E.2 and E.3


TakahikoKawasaki:
Editorial issues found in the working group draft built on December 22, 2023, published at https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html

### E.2.2. Credential Issuer Metadata, the example JSON

- The Japanese translations for `display[1].name` (在籍証明書) and `display[1].logo.alt_text` (大学のロゴ) are inappropriate. They mean "Certificate of Enrollment" and "Logo of University", respectively.

### E.3.2. Credential Issuer Metadata, the 1st paragraph

- `credentials_supported` -> `credential_configurations_supported`

### E.3.2. Credential Issuer Metadata, the example JSON

- `credentials_supported` -> `credential_configurations_supported`

### E.3.3. Authorization Details

- `vct` should not be required because `credential_configuration_id` is sufficient. (Like `doctype` is no longer required in the case of `mso_mdoc`.)
- The example JSON should contain `credential_configuration_id` and should not contain `format` and `vct`.

TakahikoKawasaki:
### E.3.2. Credential Issuer Metadata, the example JSON

- It's better to change `last_name` to `family_name` to align with [5.1. Standard Claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) of OIDC Core 1.0.
- It might be worth considering to change `is_over_18`, `is_over_21`, `is_over_65` to `age_over_18`, `age_over_21`, `age_over_65`, respectively, to align with "7.2.5 Age attestation: nearest "true" attestation above request" of "[ISO/IEC 18013-5:2021](https://www.iso.org/standard/69084.html)".

### Others

There are two `proof_types_supported` in the draft. They should be changed to `proof_types`.

selfissued:
Taka is of course right.  This especially matters in cultures like Korea, where the first name pronounced is the family name.  We should do this before the first Implementer's Draft.

jogu:
I believe @selfissued fixed the two issues in E.3.2 (now moved to A.3.2 in the latest version of the spec).

https://github.com/openid/OpenID4VCI/pull/216 will fix the Japanese text.

The remaining issue to be resolved is the one in E.3.3 (now A.3.3):

> vct should not be required because credential_configuration_id is sufficient. (Like doctype is no longer required in the case of mso_mdoc.)

Sakurann:
per discussion in #217, 

[...]

PR #195, opened 2023-12-31
Improve vc+sd-jwt metadata example

selfissued:
This addresses some of the issues raised by @TakahikoKawasaki in #172.

### DIFF
--- a/examples/credential_metadata_sd_jwt_vc.json
+++ b/examples/credential_metadata_sd_jwt_vc.json
@@ -52,9 +52,9 @@
                     "country": {}
                 },
                 "birthdate": {},
-                "age_18_or_over": {},
-                "age_21_or_over": {},
-                "age_65_or_over": {}
+                "is_over_18": {},
+                "is_over_21": {},
+                "is_over_65": {}
             }
         }
     }
--- a/examples/credential_metadata_sd_jwt_vc.json
+++ b/examples/credential_metadata_sd_jwt_vc.json
@@ -31,7 +31,7 @@
                         }
                     ]
                 },
-                "last_name": {
+                "family_name": {
                     "display": [
                         {
                             "name": "Surname",
@@ -52,9 +52,9 @@
                     "country": {}
                 },
                 "birthdate": {},
-                "is_over_18": {},
-                "is_over_21": {},
-                "is_over_65": {}
+                "age_18_or_over": {},
+                "age_21_or_over": {},
+                "age_65_or_over": {}
             }
         }
     }
--- a/examples/credential_metadata_sd_jwt_vc.json
+++ b/examples/credential_metadata_sd_jwt_vc.json
@@ -1,5 +1,5 @@
 {
-    "credentials_supported": {
+    "credential_configurations_supported": {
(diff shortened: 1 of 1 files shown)


Sakurann (approved):
looks ok. please make it clear in the issue, which on Taka's comments have been addressed and which have not, otherwise it will be harder to track later.

Sakurann (changes requested):
per 18013-5, it is `age_over_NN`. please change from `age_65_or_over` to `age_over_65`

paulbastian:
`age_xx_or__over` claims are not in the IANA registry. Should they be prepended with something like an URI to make them public claims?

selfissued:
"Age over 65" means "Age > 65" - which is 66 or greater, in the normal English reading of the phrase.  It's an example.  I'd rather that the example read in a natural way, than have readers think "I know what they meant but that's not what they actually said."

To Paul's point about registering the claims, these are example claims for illustrative purposes only.

Sakurann:
I am afraid this is not an English question. 18013-5 already defined  "age_over_xx" which is used by the governments and other implementations and we should just use it. I am against introducing a new proprietary claim "age_xx_or_over".

jogu:
> "Age over 65" means "Age > 65" - which is 66 or greater, in the normal English reading of the phrase. It's an example. I'd rather that the example read in a natural way, than have readers think "I know what they meant but that's not what they actually said."

I think the interpretation does vary; at least in the United Kingdom it is common to say "you can drink alcohol once you're over 18", which means ">= 18", or perhaps more precisely "> 18.000000".

> To Paul's point about registering the claims, these are example claims for illustrative purposes only.

Unfortunately illustrative examples in specifications have a tendency to get copied and used in real life implementations. If we are not confident in the names or they are contentious etc then I suggest we just remove the "over xx years old" claims from the example here (and perhaps refer people to the actual SD JWT VC spec for more detailed examples).

Sakurann:
I don't understand why there is a controversy around using "age_over_nn" claim which is a relatively established claim in this space.

examples/credential_metadata_sd_jwt_vc.json
```diff
@@ -52,9 +52,9 @@
                     "country": {}
                 },
                 "birthdate": {},
-                "is_over_18": {},
-                "is_over_21": {},
-                "is_over_65": {}
+                "age_18_or_over": {},
+                "age_21_or_over": {},
+                "age_65_or_over": {}
```
selfissued, line 57:
```suggestion
                "is_over_18": {},
                "is_over_21": {},
                "is_over_65": {}
```

selfissued:
OK, I switched it back to using the ISO claim names.  If it were us creating new misleading claim names, I'd stand by my request for the change, but given we're copying misleading claim names created by ISO, I'll hold my nose and live with it.

bc-pi (approved):

paulbastian (approved):

peppelinux (approved):

tplooker (approved):

Sakurann (approved):

Sakurann:
@selfissued please make it clear in the original issue which of the taka's issues have been addressed.

PR #196, opened 2023-12-31
Use event_description in notifications

### LINKED ISSUE #193: Use of error_description in notification messages is confusing
The notification request currently defines the use of the `error_description` parameter as:
> error_description: OPTIONAL. Human-readable ASCII text providing additional information, used to assist the Credential Issuer developer in understanding the error that occurred.

And an example of using `error_description` with `"event": "credential_deleted"` is given.  The problem I have with this is that it's not an error to delete a credential.

I propose to change the name of `error_description` to `event_description` and change "understanding the error that occurred" to "understanding the **event** that occurred".

selfissued:
Fixes #193

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1240,7 +1240,7 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 
 {
   "notification_id": "3fwe98js",
-  "event": "credential_deleted",
+  "event": "credential_failure",
   "event_description": "User pressed 'cancel' button"
 }
 ```
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1214,7 +1214,7 @@ The Wallet sends an HTTP POST request to the Notification Endpoint with the foll
 
 * `notification_id`: REQUIRED. String received in Credential Response or Batch Credential Response.
 * `event`: REQUIRED. Type of the notification event. It MUST be a case sensitive string whose value is either `credential_accepted`, `credential_failure` or `credential_deleted`. `credential_accepted` is to be used when the Credential was successfully stored in the Wallet, with or without user action. `credential_deleted` is to be used when the unsuccessful Credential issuance was caused by a user action. In all other unsuccessful cases, `credential_failure` is to be used.
-* `error_description`: OPTIONAL. Human-readable ASCII [@!USASCII] text providing additional information, used to assist the Credential Issuer developer in understanding the error that occurred. Values for the `error_description` parameter MUST NOT include characters outside the set `%x20-21 / %x23-5B / %x5D-7E`.
+* `event_description`: OPTIONAL. Human-readable ASCII [@!USASCII] text providing additional information, used to assist the Credential Issuer developer in understanding the event that occurred. Values for the `event_description` parameter MUST NOT include characters outside the set `%x20-21 / %x23-5B / %x5D-7E`.
 
 Below is a non-normative example of a Notification Request when a credential was successfully accepted by the End-User:
 
@@ -1241,7 +1241,7 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 {
   "notification_id": "3fwe98js",
   "event": "credential_deleted",
-  "error_description": "..."
+  "event_description": "..."
 }
 ```
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1241,7 +1241,7 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 {
   "notification_id": "3fwe98js",
   "event": "credential_failure",
-  "event_description": "Could not retrieve issuer key"
+  "event_description": "Could not retrieve issuer key."
 }
 ```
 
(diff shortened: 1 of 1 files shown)


jogu (approved):

bc-pi (approved):

Sakurann (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1241,7 +1241,7 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 {
   "notification_id": "3fwe98js",
   "event": "credential_deleted",
-  "error_description": "..."
+  "event_description": "..."
```
paulbastian, line 1244: Either input some text her or omit the optional parameter

jogu: I think having an example is good; perhaps:

```suggestion
  "event_description": "User pressed 'cancel' button"
```

paulbastian: does cancel button match with the event "credential_deleted" ? Re-reading the options for event, "credential_deleted" was confusing to me.

paulbastian: If I'm correct, this is only about the acknowledgement of the credential response? So isn't the credential denied, refused or rejected instead of deleted?

jogu: It's pretty much the only thing that can trigger 'deleted' I think, the spec defines deleted as "to be used when the unsuccessful Credential issuance was caused by a user action".

jogu: > If I'm correct, this is only about the acknowledgement of the credential response?

Yes, as the spec is currently written.

> So isn't the credential denied, refused or rejected instead of deleted?

I would suggest not reopening that argument in this PR :-)

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1241,7 +1241,7 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 {
   "notification_id": "3fwe98js",
   "event": "credential_deleted",
-  "error_description": "..."
+  "event_description": "User pressed 'cancel' button"
```
Sakurann, line 1244: suggestion was to use “credential_faliure” event and put a description “could not retrieve issuer key”.

Sakurann:
Jan-04-2024 wg call - merge once the last change to the example is applied

paulbastian (approved):
approved with the suggested changes

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1241,7 +1241,7 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 {
   "notification_id": "3fwe98js",
   "event": "credential_deleted",
```
selfissued, line 1243:
```suggestion
  "event": "credential_failure",
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1240,8 +1240,8 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 
 {
   "notification_id": "3fwe98js",
-  "event": "credential_deleted",
-  "error_description": "..."
+  "event": "credential_failure",
+  "event_description": "User pressed 'cancel' button"
```
selfissued, line 1244:
```suggestion
  "event_description": "Could not retrieve issuer key"
```

jogu (approved):

bc-pi (approved):

peppelinux (approved):

tplooker (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1240,8 +1240,8 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 
 {
   "notification_id": "3fwe98js",
-  "event": "credential_deleted",
-  "error_description": "..."
+  "event": "credential_failure",
+  "event_description": "Could not retrieve issuer key"
```
Sakurann, line 1244:
```suggestion
  "event_description": "Could not retrieve issuer key."
```

Sakurann (approved):

Sakurann (changes requested):
I am not comfortable with this error message because this sets the expectation that the wallet needs to verify the credential it received from the issuer. 
Would prefer a message like "Could not store the Credential because out of space".

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1240,8 +1240,8 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 
 {
   "notification_id": "3fwe98js",
-  "event": "credential_deleted",
-  "error_description": "..."
+  "event": "credential_failure",
+  "event_description": "Could not retrieve issuer key."
```
Sakurann, line 1244:
```suggestion
  "event_description": "Could not store the Credential. Out of storage."
```

Sakurann (changes requested):
I am not comfortable with this error message because this sets the expectation that the wallet needs to verify the credential it received from the issuer. 
Would prefer a message like "Could not store the Credential because out of space".

jogu:
> I re-reviewed and I am not comfortable with this error message because this sets the expectation that the wallet needs to verify the credential it received from the issuer.

The wallet does have to in at least some circumstances I thought? At least it seems potentially dangerous (or at a minimum a bad user experience) to store a credential that hasn't had the signature validated and has potentially been received over an unauthenticated channel without any other signature over the response. I guess maybe it's not required by the spec.

I've fine with the suggested change, anyway, just slightly surprised by the rational :)

tlodderstedt (approved):

Sakurann (approved):

Issue #147, opened 2023-12-12
Localization for display values should not be implemented with an array


Dindexx:
Currently, the spec defines the localization of display values with an array.

For example:

``` 
"display": [
    {
        "name": "Example University",
        "locale": "en-US"
    },
    {
        "name": "Example Université",
        "locale": "fr-FR"
    }
] 
```
This forces us to iterate over the array in order to find the respective localization. It would be easier if we would change this to a JSON object.

For example:

```
"display": {
    "en-US": {
        "name": "Example University"
    },
    "fr-FR": {
        "name": "Example Université"
    }
}
```

For implementers, this would mean we can go from this:

```
// Pseudocode

foreach(display in displays) {
    if (display.locale == "en-US") {
        result = display
    }
}
```
to this

```
// Pseudocode

result = display["en-US"]
```

Alternatively, if no extra fields are needed in the object, we could also just define it as a field, which would make it even more easier but also more difficult to extend for the future:

```
"display": {
    "en-US": "Example University",
    "fr-FR": "Example Université"
}
```

peppelinux:
yes, it is more smart, this enables interesting code generalizations like `display.get($user-agent-lang, {})].get("name")` I love.
It definitively simplifies the implementations according to the way we access to each element, even if it is less correct from a formal point of view (as it was before)... it is definitively more smart for who develop the product.

jogu:
I just want to point out that the code examples are, I think, over simplified.
E.g. if the user's language is "en-us" then they won't return a value tagged as "en", but I believe they should.

Perhaps the conclusion is the same anyway...

Sakurann:
I agree this change makes sense. who is willing to volunteer to do a PR?

selfissued:
No matter which representation we choose, we may want to say something about the ability to use language identifiers without a country designator, and how they interact in this context.  For instance, the following language identifiers could all be used together:

- fr
- fr-CA
- en
- en-NZ

selfissued:
As I described at https://github.com/openid/OpenID4VCI/pull/198#pullrequestreview-1800714880, this would lose the ability to use the array order to express preferences among languages, which other OpenID specs do.  We should not make this change.

Sakurann:
@selfissued As @jogu pointed out in the PR comment, there is no need for the Issuer to express a language preference in its metadata.

Also I am not convinced that the fact that OIDC has a list of preferred locales, which is a different structure in a different place is a good argument not to do this change in VCI.

[...]

PR #198, opened 2024-01-02
Use map for display objects

ntsbs:
This PR implements #147 and replaces display arrays with maps


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2107,9 +2104,8 @@ The following additional Credential Issuer metadata are defined for this Credent
...
       * `value_type`: OPTIONAL. String value determining the type of value of the claim. A non-exhaustive list of valid values defined by this specification are `string`, `number`, and image media types such as `image/jpeg` as defined in IANA media type registry for images (https://www.iana.org/assignments/media-types/media-types.xhtml#image).
-      * `display`: OPTIONAL. Array of objects, where each object contains display properties of a certain claim in the Credential for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
+      * `display`: OPTIONAL. Object containing a list of name/value pairs, where each name identifies the language as language tag values defined in BCP47 [@!RFC5646] and where each value represents the following localized display object:
```
tlodderstedt, line 2107: same here

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2198,9 +2194,8 @@ The following additional Credential Issuer metadata are defined for this Credent
...
     * `value_type`: OPTIONAL. String value determining the type of value of the claim. A non-exhaustive list of valid values defined by this specification are `string`, `number`, and image media types such as `image/jpeg` as defined in IANA media type registry for images (https://www.iana.org/assignments/media-types/media-types.xhtml#image).
-    * `display`: OPTIONAL. Array of objects, where each object contains display properties of a certain claim in the Credential for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
+    * `display`: OPTIONAL. Object containing a list of name/value pairs, where each name identifies the language as language tag values defined in BCP47 [@!RFC5646] and where each value represents the following localized display object:
```
tlodderstedt, line 2197: and here

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2251,9 +2246,8 @@ The following additional Credential Issuer metadata parameters are defined for t
...
     * `value_type`: OPTIONAL. String value determining type of value of the claim. A non-exhaustive list of valid values defined by this specification are `string`, `number`, and image media types such as `image/jpeg` as defined in IANA media type registry for images (https://www.iana.org/assignments/media-types/media-types.xhtml#image).
-    * `display`: OPTIONAL. Array of objects, where each object contains display properties of a certain claim in the Credential for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
-        * `name`: OPTIONAL. String value of a display name for the claim.
-        * `locale`: OPTIONAL. String value that identifies language of this object represented as language tag values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
+    * `display`: OPTIONAL. Object containing a list of name/value pairs, where each name identifies the language as language tag values defined in BCP47 [@!RFC5646] and where each value represents the following localized display object for a claim.
```
tlodderstedt, line 2249: and finally here as well

tlodderstedt (changes requested):

tlodderstedt (approved):

selfissued (changes requested):
I've thought about this some more and this loses a capability that we may want to have - specifically, the ability to express the preference order among the languages.  (Arrays are ordered, objects are not.)

This ability is used in OpenID Connect at https://openid.net/specs/openid-connect-core-1_0.html#ClaimsLanguagesAndScripts, among other places.  There, the specification says:
> claims_locales
OPTIONAL. End-User's preferred languages and scripts for Claims being returned, represented as a space-separated list of [BCP47](https://openid.net/specs/openid-connect-core-1_0.html#RFC5646) [RFC5646] language tag values, **_ordered by preference_**. An error SHOULD NOT result if some or all of the requested locales are not supported by the OpenID Provider.

No, we're not expressing preferences yet, but if we make this change, we are foreclosing the possibility of doing so in the future without changing the data structure.

jogu:
> I've thought about this some more and this loses a capability that we may want to have - specifically, the ability to express the preference order among the languages. (Arrays are ordered, objects are not.)
> 
> This ability is used in OpenID Connect at https://openid.net/specs/openid-connect-core-1_0.html#ClaimsLanguagesAndScripts, among other places. There, the specification says:
> 
> > claims_locales
> > OPTIONAL. End-User's preferred languages and scripts for Claims being returned, represented as a space-separated list of [BCP47](https://openid.net/specs/openid-connect-core-1_0.html#RFC5646) [RFC5646] language tag values, **_ordered by preference_**. An error SHOULD NOT result if some or all of the requested locales are not supported by the OpenID Provider.
> 
> No, we're not expressing preferences yet, but if we make this change, we are foreclosing the possibility of doing so in the future without changing the data structure.

The parameter you mention @selfissued is a request parameter.

The parameter being changed is one that's in the metadata response. I am unclear in what situations a VC issuer would need to express a language preference in its metadata. The equivalent of when localised claims are returned in id_tokens ( https://openid.net/specs/openid-connect-core-1_0.html#ClaimsLanguagesAndScripts ) is not an array and does not allow the IDP to express a preference.

jogu (changes requested):
Note that (as I already commented on the issue) the PR does not deliver the original claimed benefits (as expressed in the issue) of allowing the lookup code to be simplified to:

`result = display["en-US"]`

This code does not work for two reasons:

1. It fails if the issuer only returns a display value for 'en'
2. I believe in most programming languages it fails if the issuer returns a display value of 'EN-US', which is legal as per https://datatracker.ietf.org/doc/html/rfc5646#section-2.1.1 and the matching is required to be case insensitive.

In my opinion, we should not continue with a PR that does not deliver the lookup simplification that was the justification for making the change.

Sakurann:
@jogu I think we need to discuss and agree what is the assumption and a desired behavior here. It might be that some of us made an assumption that  the Wallet and the Issuer know what language identifier to use to look up. in which case, this new structure helps with simplification. but if as you point out, the issuer might be using "en-US", "en" or "EN-US" as the same meaning (english), than this new structure does not really make life of the wallet easier.

jogu:
>  It might be that some of us made an assumption that the Wallet and the Issuer know what language identifier to use to look up. in which case, this new structure helps with simplification. but if as you point out, the issuer might be using "en-US", "en" or "EN-US" as the same meaning (English)

Yeah, it would be good to write down the assumptions. I don't think there's any language in the specification that says you can't use 'en' or 'EN-US' (the spec currently just says "String value that identifies the language of this object represented as a language tag taken from values defined in BCP47 RFC5646" , so adding text that restricts the values that can be used is one way to go (not sure it's a good way though, as it may be difficult to word without breaking some of the more complex to deal with countries).

> "en-US", "en" or "EN-US" as the same meaning (English)

To be pedantic, en and en-US have different meanings - "English" vs "English (USA)", and also distinct from 'en-GB' ("British English"). [i.e. "en" would be used when an issuer decides to provide the same translations for both British and USA users.]

Sakurann:
> To be pedantic, en and en-US have different meanings - "English" vs "English (USA)", and also distinct from 'en-GB' ("British English"). [i.e. "en" would be used when an issuer decides to provide the same translations for both British and USA users.]

right. i don't disagree. and if this is true, the issuer will need to have separate entries for "en-US", "en" and "EN-US", so there won't be a case when the wallet can use either "en-US", "en" or "EN-US" to get the same object, in which case, the new structure proposed in this PR would probably be better.

peppelinux:
this change improves the development experience using the lookup values in a smart way, having them as member name and not in the corresponding value, the json object with the "locale" member.

The discussion in this thread is moving to the concerns about how the i18n values can be effective or not. From my perspective this concern can be applied in the current text as well, using the `locale` member.

regarding the valuable comments made by @jogu I might assume that a fallback strategy could be suggested in the current PR, by defining the "default" member name. This default would bring all the constant values to be reused in each localization, reducing the size of the metadata, as the following example shows. I also assume that the default value could be very often set to english.

then

````
            "display": {
                "default": {
                    "logo": {
                        "url": "https://exampleuniversity.com/public/logo.png",
                        "alt_text": "a square logo of a university"
                    },
                    "background_color": "#12107c",
                    "text_color": "#FFFFFF",
                    "name": "University Credential"
                },
                "en-us": {
                    "name": "University Credential",
                    "alt_text": "a square logo of a university"
                },
                "it-it": {
                    "name": "Credenziali d'Accesso Campus Universitario",
                    "alt_text": "Logo della Credenziale Universitaria corrispondente al logo dell'Università che l'ha emessa."
                }
            ]
            }

````

regarding the case (in)sensitiveness, according to RFC 5646, language tags are case-insensitive. However, it is recommended to use certain casing conventions for different parts of the tag for better readability. For instance, language subtags are typically written in lowercase, script subtags in title case, and region subtags in uppercase. But in terms of functionality, they can be used in any case and for the scopes of the specs, with the will to make impl's life easier, I would suggest to mandate lowercased member value identifiers, as I made in the example above.

**Comment by Sakurann** [comment]
discussed at the Jan-04-2024 WG call. I think the current summary is that 
1. when the wallet knows the exact locale the issuer uses, the proposed can simplify the look up for the wallet, for example when there is a set of locales defined in the ecosystem.
2. there are cases when the current approach does not give any look-up benefits because the issuer and the wallet might be using different case while locale is case-sensitive.

Some workarounds suggested were to define a fallback to be `en`, `fr` without country code, so that the call does not fail. or recommend using a certain case over the other (i.e. use en-US over en-us or EN-US).

the question, I think, is, whether the benefit for in use-cases 1. is worth making this breaking change

**Comment by sudeshrshetty** [comment]
agree with @peppelinux , enforcing lower-cased locale constants and defining default (fallback) values should solve the problem.

**Comment by bc-pi** [comment]
The current format/spec text allows for a display that does not have a local. Which can simplify things, for example, when there is one locale used in the ecosystem. Or as a default when an appropriate more specific locale isn't present. 

For example, "gpa" in the spec example here https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-12.html#section-10.2.3-6 has a display object with name but no local.

```
    ...
                    "gpa": {
                        "display": [
                            {
                                "name": "GPA"
                            }
                        ]
                    }
     ...

```

The proposed format doesn't allow for the same (or would make it very awkward using empty string as the map key).

**Comment by jogu** [comment]
Brian's right there, I think it's an important point that the proposed changes currently in this PR are actually removing functionality we currently have.

**Comment by peppelinux** [comment]
I can only say that during the write of my proposal I though about the default member name set to `""` and then I simply realized that I would prefer a "default" value as an identifier just to avoid json serializers that may have problem with "" (void) member names.

In my understanding of what direction the discussion is taking, I bring the evidence that in my proposal the default value is not removed but explicitly defined. I feel optimistic about a positive resolution of this discussion and the update of the content brought in this PR.

**Comment by jogu** [comment]
I don't mind either way, but in my opinion this PR can't be merged as is, so it's in danger of missing the ID1 cut off, and if that happens it's highly unlikely this change will ever happen.

**Comment by ntsbs** [comment]
Following the proposal of @peppelinux, how about the following definition:

* `display`: OPTIONAL. Object containing a list of name/value pairs, where each name identifies the language as language tag values defined in BCP47 [@!RFC5646]. All components of a language tag must be lowercase (e.g. "en-us"). As key to define default localized values `default` must be used. If this key is present, it should contain a localized display object with the same parameters as the other keys in the display object. If a preferred language tag is not available in the display object, the value of the corresponding key in the `default` object should be used instead. A localized display object contains the following non-exhaustive list of parameters that MAY be included:
  * `name`: OPTIONAL. String value of a display name for the Credential Issuer.

**Comment by paulbastian** [comment]
I think by mandating the lowercase of the `locale` we solve most of the issues and actually improve over the current solution.

Concerning the default: I think the current solution that `locale` is optional is rather odd. Wallets can have their own mechanisms of defining default values (e.g. based on the languages supported/downloaded in the phone), so we might not even require the `default`. Otherwise there is also the possibility to set a specific language as default/fallback instead of replicating the data. 

Additionally, we have the HTTP Content-Language Header that could be used to signalize the default/fallback.

In general, I'm leaning to approve if these points are addressed

**Comment by selfissued** [comment]
As Brian pointed out, this change means that we lose the ability to have an entry without a locale.  That sounds like a meaningful loss.

To the thread about lowercasing the locale values, even if we do that (which means we'd be using non-standard spellings), implementations interested in French would still have to be prepared to accept and prioritize all of "fr", "fr-fr", "fr-ca", "fr-ne", etc.

I still think that this is a step backwards, which we should not take.

**Comment by tplooker** [comment]
I think for us to accept this proposal we need to be

1. Confident that locale is the most important piece of information to index "display" information against, because if we are not, which im not necessarily saying, this proposal risks becoming less efficient then what we currently have.
2. To @selfissued's point, we'd need to believe that wallets and verifiers will often only have one preferred locale which is consistently supported by issuers, otherwise the sort of `result = display["en-US"]` simple look up doesn't work because the wallet has to repeat the query for a bunch of different locales trying to find a match. For the record I don't think this assumption is likely reliable given there are at least the following variations for the english language alone.

```
en-AU
en-BZ
en-CA
en-cb
en-GB
en-IE
en-IN
en-JM
en-MT
en-MY
en-NZ
en-PH
en-SG
en-TT
en-US
en-ZA
en-ZW
```

It feels far more likely that the sort of query a wallet or verifier will want to do when selecting display information is more along the lines of

```javascript
const supported_locales = [ "en", "en-AU", "en-NZ", "en-US" ];

const display = [
  {
      "name": "Example University",
      "locale": "en-US"
  },
  {
      "name": "Example Université",
      "locale": "fr-FR"
  }
];

const match = display.find(item => supported_locales.includes(item.locale));

console.log(match.name);
```

OR

```javascript
const supported_base_locale = "en";

const display = [
  {
      "name": "Example University",
      "locale": "en-US"
  },
  {
      "name": "Example Université",
      "locale": "fr-FR"
  }
];

const match = display.find(item => item.locale.startsWith(supported_base_locale));

console.log(match.name);
```

Both of the above sorts of queries get a bit more awkward in certain programming languages when you shift to a key value dictionary structure which is what this PR would do. For those reasons I'm not in favour of this change.

tplooker (changes requested):
See above for my comments

paulbastian:
> To the thread about lowercasing the locale values, even if we do that (which means we'd be using non-standard spellings), implementations interested in French would still have to be prepared to accept and prioritize all of "fr", "fr-fr", "fr-ca", "fr-ne", etc.

How is that different to the current solution?

paulbastian:
probably the discussion is difficult because no solution is significantly better. therefore it's probably okay to not merge this, as it still is a breaking change.

ntsbs:
After the discussion here I also think that this not mandatory to have

Sakurann:
Jan-11-2024 WG call agreed to close this PR and the issue and if there is an appetite to discuss this further, a new issue would be needed.

PR #208, opened 2024-01-11
remove `format` from the Credential Response

### LINKED ISSUE #204: Remove format from Credential Response
The format of the credential at this point is clear to both Issuer and Wallet. The parameter does not serve any purpose.

Proposal: delete it

paulbastian:
Closes #204

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -926,14 +926,13 @@ If the Credential Response is not encrypted, the media type of the response MUST
 
 The following claims are used in the JSON-encoded Credential Response body:
 
-* `format`: REQUIRED. String denoting the format of the issued Credential.
 * `credential`: OPTIONAL. Contains issued Credential. MUST be present when `transaction_id` is not returned. MAY be a string or an object, depending on the Credential format. See (#format_profiles) for the Credential format specific encoding requirements.
 * `transaction_id`: OPTIONAL. String identifying a Deferred Issuance transaction. This claim is contained in the response if the Credential Issuer was unable to immediately issue the Credential. The value is subsequently used to obtain the respective Credential with the Deferred Credential Endpoint (see (#deferred-credential-issuance)). It MUST be present when the `credential` parameter is not returned. It MUST be invalidated after the Credential for which it was meant has been obtained by the Wallet.
 * `c_nonce`: OPTIONAL. String containing a nonce to be used to create a proof of possession of key material when requesting a Credential (see (#credential_request)). When received, the Wallet MUST use this nonce value for its subsequent Credential Requests until the Credential Issuer provides a fresh nonce.
 * `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`.
 * `notification_id`: OPTIONAL. String identifying an issued Credential that the Wallet includes in the Notification Request as defined in (#notification ). This parameter MUST NOT be present if `credential` parameter is not present.
 
-The `format` key determines the Credential format and encoding of the credential in the Credential Response. Details are defined in the Credential Format Profiles in (#format_profiles). 
+The Credential format and encoding of the `credential` MUST match the Credential format specified in the Credential Request. Details are defined in the Credential Format Profiles in (#format_profiles).
 
 Credential formats expressed as binary data MUST be base64url-encoded and returned as a string.
 
@@ -945,7 +944,6 @@ Content-Type: application/json
 Cache-Control: no-store
 
 {
-  "format": "jwt_vc_json",
   "credential": "LUpixVCWJk0eOt4CXQe1NXK....WZwmhmn9OQp6YxX0a2L",
   "c_nonce": "fGFF7UkhLa",
   "c_nonce_expires_in": 86400  
@@ -2296,7 +2294,8 @@ Wallet Providers may also provide a marketplace where Issuers can register to be
    [[ To be removed from the final specification ]]
    
    -13
-  
+
+   * remove `format` from the Credential Response
    * added `signed_metadata` parameter
    * clarified that logo can is a uri and not a url only
    * moved the annex with Credential format profiles to the top of all annexes
--- a/examples/credential_response_jwt_vc_json.txt
+++ b/examples/credential_response_jwt_vc_json.txt
@@ -4,7 +4,6 @@ Content-Type: application/json
 Cache-Control: no-store
   
 {
-    "format": "jwt_vc_json",
     "credential": "eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJ2YyI6eyJAY29udGV4dCI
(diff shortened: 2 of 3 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -926,14 +926,13 @@ If the Credential Response is not encrypted, the media type of the response MUST
...
 * `c_nonce`: OPTIONAL. String containing a nonce to be used to create a proof of possession of key material when requesting a Credential (see (#credential_request)). When received, the Wallet MUST use this nonce value for its subsequent Credential Requests until the Credential Issuer provides a fresh nonce.
 * `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`.
 * `notification_id`: OPTIONAL. String identifying an issued Credential that the Wallet includes in the Notification Request as defined in (#notification ). This parameter MUST NOT be present if `credential` parameter is not present.
 
-The `format` key determines the Credential format and encoding of the credential in the Credential Response. Details are defined in the Credential Format Profiles in (#format_profiles). 
+The Credential format and encoding of the `credential` MUST match the Credential format specified in the Credential Request. Details are defined in the Credential Format Profiles in (#format_profiles).
```
awoie, line 935: Encoding is not used anywhere else in this spec like this. I suggest to remove it.
```suggestion
The Credential format of the `credential` MUST match the Credential format specified in the Credential Request. Details are defined in the Credential Format Profiles in (#format_profiles).
```

awoie (approved):

bc-pi (approved):
"approving" but suggest that @awoie's suggestion be incorporated

tplooker (approved):

c2bo:
Batch and Deferred Credential Endpoint seem to be referencing the `format` defined in 7.2 (that is removed here). So I guess we would need to remove format references/examples for all types of credential responses (Batch / Deferred)?

jogu (approved):
Approving, but agree Oliver's suggested change should be applied before merging.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: we need to put somewhere in the spec that "The `format` parameter value determines the Credential format and encoding of the credential in the Credential Response.". maybe it does not belong here anymore, but it should not be just omitted like this.

Sakurann:
```suggestion
The Credential format and encoding of the `credential` is determined by the Credential format specified in the Credential Request. Details are defined in the Credential Format Profiles in (#format_profiles).
```

Sakurann (changes requested):
per suggested changes.

paulbastian:
@c2bo good catch and @jogu thanks for doing the changes!

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: I commited @Sakurann comment as it does not affect @awoie proposal to remove encoding. I think we need more input for keeping or removing encoding

jogu: I'm good with @Sakurann's new wording; it is accurate, the encoding of the credential does vary depending on the credential format (e.g. VCI requires that mdoc is base64url encoded in the response).

awoie: IMO, it is an unnecessary repetition but I'm ok with either way. To me that sentence is just a bit confusing since encoding is always implied by the format. There is no format identified by its format identifier with multiple encodings. If we have such a case, we should fix that.

paulbastian: I actually agree with Oliver and only kept encoding in there to keep the changes as small is possible for this late PR

Sakurann: From the conversations I am having, it is not straightforward to the implementers that in the VP Token, sometimes credential is a string and sometimes it is an object. so we should be clear even if it feels redundant.

Sakurann (approved):

c2bo (approved):

openid-4-verifiable-credential-issuance-1_0.md
awoie: It is similar to the EdDSA problem in JOSE where it could mean Ed443 or Ed25519. We should define format identifiers unambiguously.

awoie: Also encoding is a characteristic of specific formats. Usually a format has one encoding. If there are legacy cases such as VCDM 1.1/1.0 where there are more encodings, we define already format identifiers like `jwt_vc_json`, and so on, that includes the encoding.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -926,14 +926,13 @@ If the Credential Response is not encrypted, the media type of the response MUST
...
 * `c_nonce`: OPTIONAL. String containing a nonce to be used to create a proof of possession of key material when requesting a Credential (see (#credential_request)). When received, the Wallet MUST use this nonce value for its subsequent Credential Requests until the Credential Issuer provides a fresh nonce.
 * `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`.
 * `notification_id`: OPTIONAL. String identifying an issued Credential that the Wallet includes in the Notification Request as defined in (#notification). This parameter MUST NOT be present if `credential` parameter is not present.
 
-The value of the `format` parameter determines the Credential format and encoding of the credential in the Credential Response. Details are defined in the Credential Format Profiles in (#format_profiles).
+The format and encoding of the `credential` value is determined by the Credential format specified in the Credential Request. Details are defined in the Credential Format Profiles in (#format_profiles).
 
 Credential formats expressed as binary data MUST be base64url-encoded and returned as a string.
```
paulbastian, line 937:
```suggestion
The format of the `credential` value is determined by the Credential format specified in the Credential Request. 

The encoding of the Credential in the `credential` value depends on the Credential Format Profiles defined in (#format_profiles). Credential formats expressed as binary data MUST be base64url-encoded and returned as a string.
```

paulbastian: Is this an acceptable compromise? @Sakurann @awoie

awoie: To me that works. 

I think your suggestion makes it more clear that encoding refers to the encoding of the `credential` value and does not refer to the encoding of the credential itself. For instance, mdocs have a native CBOR encoding while being encoded as base64url in the `credential` value.

Sakurann: here is my suggestion

```suggestion
The format of the Credential in the Credential Response is determined by the value of the `format` parameter specified in the Credential Request.

The encoding of the Credential returned in the `credential` parameter depends on the Credential Format. Credential formats expressed as binary data MUST be base64url-encoded and returned as a string.

More details such as Credential Format identifiers are defined in the Credential Format Profiles in (#format_profiles). 
```

Issue #206, opened 2024-01-11
mention OAuth Server Metadata in Section 11.2.3.


tlodderstedt:
The text currently refers to RFC 8414 but does not mention .well-known/oauth-authorization-server. Implementers typically seem to assume .well-known/openid-configuration shall be used for that purpose. I suggest adding a line mentioning .well-known/oauth-authorization-server.

Sakurann:
Jan-11-2024 WG call - agreed this is editorial and worth doing.

Sakurann:
PR merged

PR #209, opened 2024-01-11
added text on oauth-authorization-server metadata location

tlodderstedt:
solves #206


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1330,7 +1330,7 @@ Accept-Language: fr-ch, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
...
+* `authorization_servers`: OPTIONAL. Array of strings, where each string is an identifier of the OAuth 2.0 Authorization Server (as defined in [@!RFC8414]) the Credential Issuer relies on for authorization. If this parameter is omitted, the entity providing the Credential Issuer is also acting as the AS, i.e., the Credential Issuer's identifier is used to obtain the the Authorization Server metadata. The actual OAuth 2.0 Authorization Server metadata is obtained from the `oauth-authorization-server` well-known location as defined in Section 3 of [@!RFC8414]. When there are multiple entries in the array, the Wallet may be able to determine which AS to use by querying the metadata; for example, by examining the `grant_types_supported` values, the Wallet can filter the server to use based on the grant type it plans to use. When the Wallet is using `authorization_server` parameter in the Credential Offer as a hint to determine which AS to use out of multiple, the Wallet MUST NOT proceed with the flow if the `authorization_server` Credential Offer parameter value does not match any of the entries in the `authorization_servers` array.
```
selfissued, line 1333:
```suggestion
* `authorization_servers`: OPTIONAL. Array of strings, where each string is an identifier of the OAuth 2.0 Authorization Server (as defined in [@!RFC8414]) the Credential Issuer relies on for authorization. If this parameter is omitted, the entity providing the Credential Issuer is also acting as the Authorization Server, i.e., the Credential Issuer's identifier is used to obtain the Authorization Server metadata. The actual OAuth 2.0 Authorization Server metadata is obtained from the `oauth-authorization-server` well-known location as defined in Section 3 of [@!RFC8414]. When there are multiple entries in the array, the Wallet may be able to determine which AS to use by querying the metadata; for example, by examining the `grant_types_supported` values, the Wallet can filter the server to use based on the grant type it plans to use. When the Wallet is using `authorization_server` parameter in the Credential Offer as a hint to determine which AS to use out of multiple, the Wallet MUST NOT proceed with the flow if the `authorization_server` Credential Offer parameter value does not match any of the entries in the `authorization_servers` array.
```

jogu (approved):
LGTM. Agree with @selfissued's suggestion that fixes the "the the".

tplooker (approved):

Sakurann (approved):

PR #211, opened 2024-01-17
Change 'claimed url' to 'custom URL scheme'

### LINKED ISSUE #200: That's not a claimed URL
Section 11.1.1: "the following claimed URL is used: openid-credential-offer://."

That's not a claimed URL, but a custom scheme.

jogu:
This is the same term used in the OAuth native apps BCP:

https://datatracker.ietf.org/doc/html/rfc8252

fixes #200


bc-pi (approved):
10/10. No notes.

danielfett (approved):

paulbastian (approved):

PR #216, opened 2024-01-18
Fix Japanese text in examples

jogu:
Thanks to Kosuke Koiwai from KDDI for help with the translations.

I also updated the English alt text as it is not considered good practice to include 'figure of' in the alt text as screen readers for the blind etc already know it is an image.

Fixes part of #172


jogu:
Taka from Authlete kindly reviewed this offline and confirmed it looks good.

tlodderstedt (approved):

Sakurann (approved):

PR #219, opened 2024-01-19
added format to authorization details and respective format specific additional claims

### LINKED ISSUE #217: re-add an option to use format/type in `authorization_details`
documenting for visibility.

On Jan-18-2024 WG call, @tlodderstedt brought up an issue that the change made in #137 mandates the wallet to know `credential_configuration_id`, which takes away the advantage of RAR over `scopes` approach.

Discussed in the WG and agreed to re-add the option to use format/type in `authorization_details` in addition to an option to use `credential_configuration_id for the following reasons:
- there are implementers that are using out-of-band mechanisms instead of the credential issuer metadata as defined in the spec. or implementers that want to use metadata only to communicate the endpoint and display and not the credential format/type. (needs another issue if credential_configurations_supported should be optional)
- we may end up keeping only `authorization_details`, but right not there is not enough implementation experience and keeping only one option would limit flexibility
- current implementations that implement `credential_configuration_id` in RAR would not be affected.
- AS that supports RAR will have to make a decision whether to support both format and id or one of those approaches (cc @pmhsfelix )

tlodderstedt:
resolves #217.


Sakurann:
resolves #217.

paulbastian (changes requested):
As I stated in the WG Call, I believe that `credential_configuration_id` is the better choice. In general, I have trouble understanding how OpenID4VCI works well without metadata, as the Wallet needs to known:

- the credential endpoint
- supported proof types
- supported credentials and formats
- display data for the issuer and the credential

Therefore I believe in productive environment it will be very common that the Issuer has the ability to host metadata.

Moreover there are two ways to start Auth Code flow:

1. Credential Offer -> we need the `credential_configuration_id` anyway
2. The Wallet has implicit knowledge/is pre-configured, in this case it needs to know the `credential_issuer` anyway and it is not a big difference between storing format/type and configuration_id

In the end, my argument is that its less options to support, which is generally better. However I see that this is becoming very similar to scope, which is an indicator that the concept is not ideal yet and we are missing more implementation experience.

Therefore, feel free to override my request for change, I just wanted to share my opinion.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -414,12 +414,19 @@ There are two possible ways to request issuance of a specific Credential type in
...
 
 * `type`: REQUIRED. String that determines the authorization details type. It MUST be set to `openid_credential` for the purpose of this specification.
-* `credential_configuration_id`: REQUIRED. String specifying a unique identifier of the Credential being described in the `credential_configurations_supported` map in the Credential Issuer Metadata as defined in (#credential-issuer-parameters). The referenced object in the `credential_configurations_supported` map conveys the details, such as the format, for issuance of the requested Credential. This specification defines Credential Format specific Issuer Metadata in (#format_profiles).
+* `credential_configuration_id`: String specifying a unique identifier of the Credential being described in the `credential_configurations_supported` map in the Credential Issuer Metadata as defined in (#credential-issuer-parameters). The referenced object in the `credential_configurations_supported` map conveys the details, such as the format, for issuance of the requested Credential. This specification defines Credential Format specific Issuer Metadata in (#format_profiles).
```
Sakurann, line 417:
```suggestion
* `credential_configuration_id`: OPTIONAL. String specifying a unique identifier of the Credential being described in the `credential_configurations_supported` map in the Credential Issuer Metadata as defined in (#credential-issuer-parameters). The referenced object in the `credential_configurations_supported` map conveys the details, such as the format, for issuance of the requested Credential. This specification defines Credential Format specific Issuer Metadata in (#format_profiles). This parameter MUST NOT be present if `format` parameter is present.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -414,12 +414,19 @@ There are two possible ways to request issuance of a specific Credential type in
...
+* `format`: String identifying the format of the Credential the Wallet needs. This Credential format identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format_profiles).
```
Sakurann, line 418:
```suggestion
* `credential_configuration_id`: OPTIONAL. String specifying a unique identifier of the Credential being described in the `credential_configurations_supported` map in the Credential Issuer Metadata as defined in (#credential-issuer-parameters). The referenced object in the `credential_configurations_supported` map conveys the details, such as the format, for issuance of the requested Credential. This specification defines Credential Format specific Issuer Metadata in (#format_profiles). This parameter MUST NOT be present if `format` parameter is present.
* `format`: OPTIONAL. String identifying the format of the Credential the Wallet needs. This Credential format identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format_profiles). This parameter MUST NOT be present if `credential_configuration_id` parameter is present.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -414,12 +414,19 @@ There are two possible ways to request issuance of a specific Credential type in
...
 
-The following is a non-normative example of an `authorization_details` object:
+Either `credential_configuration_id` or `format` MUST be present in an authorization details object of `type`  `openid_credential`.
```
Sakurann, line 420:
```suggestion
Either `credential_configuration_id` or `format` MUST be present in an authorization details object of type `openid_credential`. 
```

Sakurann (approved):
i think we need to figure out optional/must in the authorization request, but looks good otherwise.

jogu:
> As I stated in the WG Call, I believe that `credential_configuration_id` is the better choice. In general, I have trouble understanding how OpenID4VCI works well without metadata, as the Wallet needs to known:
> 
> * the credential endpoint
> * supported proof types
> * supported credentials and formats
> * display data for the issuer and the credential
> 
> Therefore I believe in productive environment it will be very common that the Issuer has the ability to host metadata.

We had a fairly long conversation about this in one of last year's issues or PRs (likely the that one that first added configuration ids). As I understood it, the issue was not so much that people didn't want to have metadata at all - it was that they didn't want to be required to list every supported credential in the metadata. (There's parallels for this in OAuth, e.g. https://datatracker.ietf.org/doc/html/rfc8414#section-2 the AS isn't required to list every supported scope in `scopes_supported`.)

jogu (approved):

openid-4-verifiable-credential-issuance-1_0.md
c2bo: Would this work for you? Sounds kind of weird to me but I'd definitely prefer some text with "REQUIRED when <..>"
```suggestion
* `credential_configuration_id`: REQUIRED when `format` is not present and MUST NOT be present if `format` parameter is present. String specifying a unique identifier of the Credential being described in the `credential_configurations_supported` map in the Credential Issuer Metadata as defined in (#credential-issuer-parameters). The referenced object in the `credential_configurations_supported` map conveys the details, such as the format, for issuance of the requested Credential. This specification defines Credential Format specific Issuer Metadata in (#format_profiles).
* `format`: REQUIRED when `credential_configuration_id` is not present and MUST NOT be present if `credential_configuration_id` parameter is present. String identifying the format of the Credential the Wallet needs. This Credential format identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format_profiles).
```

Sakurann: we looked at OID4VP and in the similar situations, such as `presentation_definition` and `presentation_definition_uri`, we have not put neither REQUIRED nor OPTIONAL.
https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#section-5-8.2

awoie, line 418: I think we should add REQUIRED/OPTIONAL etc. in front of them. Readers would get confused since we use those WORDS in other places of the spec too.

awoie (commented):
I'm not excited about this change. I agree with @paulbastian that metadata should be required, even if it was disseminated out-of-band. It would make things so much simpler. But I understand that people might have overlooked the implications in the previous PR that changed that assumption. I'm not blocking this PR.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* `credential_configuration_id`: Either `credential_configuration_id` or `format` MUST be present in an authorization details object of type `openid_credential`. String specifying a unique identifier of the Credential being described in the `credential_configurations_supported` map in the Credential Issuer Metadata as defined in (#credential-issuer-parameters). The referenced object in the `credential_configurations_supported` map conveys the details, such as the format, for issuance of the requested Credential. This specification defines Credential Format specific Issuer Metadata in (#format_profiles).
* `format`: Either `credential_configuration_id` or `format` MUST be present in an authorization details object of type `openid_credential`. String identifying the format of the Credential the Wallet needs. This Credential format identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format_profiles).
```

Sakurann:
```suggestion
* `credential_configuration_id`: REQUIRED when `format` parameter is not present. It MUST NOT be present if `format` parameter is present. String specifying a unique identifier of the Credential being described in the `credential_configurations_supported` map in the Credential Issuer Metadata as defined in (#credential-issuer-parameters). The referenced object in the `credential_configurations_supported` map conveys the details, such as the format, for issuance of the requested Credential. This specification defines Credential Format specific Issuer Metadata in (#format_profiles).
* `format`: REQUIRED when `credential_configuration_id` parameter is not present. It MUST NOT be present if `credential_configuration_id` parameter is present. String identifying the format of the Credential the Wallet needs. This Credential format identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format_profiles).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -414,12 +414,19 @@ There are two possible ways to request issuance of a specific Credential type in
...
+* `credential_configuration_id`: String specifying a unique identifier of the Credential being described in the `credential_configurations_supported` map in the Credential Issuer Metadata as defined in (#credential-issuer-parameters). The referenced object in the `credential_configurations_supported` map conveys the details, such as the format, for issuance of the requested Credential. This specification defines Credential Format specific Issuer Metadata in (#format_profiles).
+* `format`: String identifying the format of the Credential the Wallet needs. This Credential format identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format_profiles).
 
-The following is a non-normative example of an `authorization_details` object:
+Either `credential_configuration_id` or `format` MUST be present in an authorization details object of type `openid_credential`. 
+
```
Sakurann, line 421:
```suggestion
```

selfissued (approved):

c2bo (approved):
I share some of the concerns of Paul (and the benefit of a reduction of complexity), but I agree that right now we probably have too little experience/feedback. So I guess the best approach at the current state of the draft is to keep it in and revisit later.
That being said, I do believe we should definitely revisit this sometime this year.

PR #220, opened 2024-01-19
rename credential_configurations to credential_configuration_ids

### LINKED ISSUE #221: rename `credential_configurations` to `credential_configuration_ids`
I was explaining the concept of the naming of the parameters and the concept of three different layers to the engineering team (in the spirit of issue #91 and recent parameter name changes) and it was pointed out that there is the discrepancy in the naming of the `credential_configurations` parameter in the Credential offer and `credential_configuration_id` in the authorization details, even though they point to the same identifier.

I think it makes sense to align the Credential Offer parameter name to `credential_configuration_ids`. For the sake of preventing breaking changes, if there is enough agreement this should probably go into the spec before ID-1.

Sakurann:
aligning with `credential_configuration_id` in authorization details. resolves #221

### DIFF
--- a/examples/credential_offer_authz_code.txt
+++ b/examples/credential_offer_authz_code.txt
@@ -3,7 +3,7 @@ Content-Type: application/json
 
 {
     "credential_issuer": "https://credential-issuer.example.com",
-    "credential_configurations": [
+    "credential_configuration_ids": [
         "UniversityDegreeCredential"
     ],
     "grants": {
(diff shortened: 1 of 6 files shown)


paulbastian (approved):

c2bo (approved):
Only thing that I wondered about is that `examples/credential_offer_sd_jwt_vc.json` is not referenced in the spec at all right now? Is that intentional or is that missing in the draft right now?

jogu (approved):

jogu (approved):

tplooker (approved):

awoie (approved):

PR #232, opened 2024-01-24
Fix layout in "VC Secured using Data Integrity" issuer metadata

jogu:
Indentation of 'mandatory' was incorrect making it appear at the wrong level in the bulleted list.


tlodderstedt (approved):

c2bo (approved):

Sakurann (changes requested):
the current indentation is correct because `mandatory` is one of the parameters that  can be included inside `credentialSubject`.

Sakurann (approved):

PR #235, opened 2024-01-26
Changing the structure of `proof_types` from an array to a `proofs_supported` map

### LINKED ISSUE #37: Missing link between `cryptographic_suites_supported` and `format` of credential
### Original Reporter: TimoAnimo

The credential issuer metadata defines the `cryptographic_suites_supported` property, which indicates to the wallet which cryptographic suites an issuer supports for a specific credential type.

However I feel like there’s a missing link between `cryptographic_suites_supported`, the algorithm used in the proof in the credential request and the cryptographic suite that is used for signing the credential.

‌

When the issuer offers the wallet a credential of type `ldp_vc` and the `cryptographic_suites_supported` includes `Ed25519Signature2018`, however the proof in the credential request includes a proof with an `ES256` JWS for `did:example:123`, is this valid? 

The credential is issued in the `ldp_vc` format and the spec mentions:

> Cryptosuites for Credentials in `jwt_vc` format should use algorithm names defined in [IANA JOSE Algorithms Registry](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms). Cryptosuites for Credentials in `ldp_vc` format should use signature suites names defined in [Linked Data Cryptographic Suite Registry](https://w3c-ccg.github.io/ld-cryptosuite-registry/). 

However this doesn’t mention how the cryptosuites should impact the proof in the credential request.  Could the `cryptographic_suites_supported` also contain `EdDSA` / `ES256` even though the credential is issued in `ldp_vc` format? Or should you infer the alg to use for the proof type based on the crypto suites supported \(`Ed25519Signature2018` maps to `EdDSA`, etc..\)?

‌

In some API’s I’ve seen where we used `ES256` for the credential request proof, but the credential was always issued in `Ed25519Signature2018` format. Should the wallet have control over this?

### LINKED ISSUE #214: Separation of cryptographic_binding_methods_supported and cryptographic_suites_supported parameters may cause problems
In the credential issuer metadata in `credential_configurations_supported` the parameters `cryptographic_binding_methods_supported` and `cryptographic_suites_supported` are defined as a format unspecific parameter and can be configured separately.

Several problems exist with this approach

1. When using a single value in `cryptographic_binding_methods_supported` the meaning of `cryptographic_suites_supported` is obvious, all included values refer to the one binding method. But if several binding methods are used, it is not that easy. Some of the suites may only be valid for some of the binding methods. Different binding methods may require different values and formats for the same crypto suites (like JOSE and COSE).
2. Currently the values for `cryptographic_suites_supported` are tied to the credential format. For jwt_vc IANA.JOSE.ALGS should be used and for ldp_vc LD_SUITES_REGISTRY. In fact the valid values depend more on the cryptographic_binding_method used and not the credential format.
3. It is unclear which binding method will be used in the credential when several are supported.

I suggest

1. to remove `cryptographic_suites_supported` and use objects as values in `cryptographic_binding_methods_supported`. This way the suites valid for a specific binding method can be mentioned there,
2. to remove [the suggestion to use specific values depending on the credential format](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-11.2.3-2.11.2.4) and suggest specific values depending on the binding method,
3. to leave it up to credential format profiles to further constrain the usage of specific binding methods and suites with specific formats and
4. to discuss problem 3 raised above.

Another option could be to remove both parameters completely as format unspecific parameters and let the credential format profiles define custom parameters if required.

Example:

```json
"credential_configurations_supported": {
        "UniversityDegreeCredential": {
            "format": "jwt_vc_json",
            "scope": "UniversityDegree",
            "cryptographic_binding_methods_supported": {
                "jwk": {
                  "cryptographic_suites_supported": [
                     "ES256"
                  ]
                },
                "cose_key": {
                  "cryptographic_suites_supported": [
                     {
                       "cose_algorithm": -7,  // COSE Algorithms value for ES256
                       "curve": 1 // COSE Curve P-256
                     }
                  ]
                }
            },
            ...
}
```

Sakurann:
resolves #214. resolves #37

* changes the structure of `proof_types` from an array to a `proofs_supported` map that contains a required `proof_signing_alg_values_supported` parameter (required to avoid having an empty object...).
* renamed `cryptographic_suites_supported` to `credential_signing_alg_values_supported` to clarify the purpose of the parameter.


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1352,9 +1352,10 @@ This specification defines the following Credential Issuer Metadata parameters:
...
+  * `proofs_supported`: OPTIONAL. Object that describes specifics of the key proof(s) that the Credential Issuer supports. This object contains a list of name/value pairs, where each name is a unique identifier of the supported proof type being described. This identifier is used in the Credential Request defined in (#credential_request) to communicate to the Wallet specifics about the proof of possession of the cryptographic key material. A list of valid key proof type values defined by this specification can be found in (#proof_types). The value in the name/value pair is an object that contains metadata about the key proof and contains the following parameters defined by this specification:
+    * `proof_signing_alg_values_supported`: REQUIRED. Array of case sensitive strings that identify the algorithms that the Issuer uses to sign the proof. Algorithm names used SHOULD be one of those defined in [@IANA.JOSE.ALGS], [@LD_Suite_Registry] or [@!ISO.18013-5].
```
TimoGlastra, line 1358: If we're already making this an object per proof_type, i think it might be useful to define the structure of the object in the proof key types section. For `cwt` and `jwt` ,`proof_signing_alg_values_supported` makes sense. However for data integrity proofs, you can have different types (`DataIntegrityProof`, `Ed25519Signature2020`), but also for `DataIntegrityProof` you have `cryptosuites`.

E.g. for `credential_signing_alg_values_supported` it mentions "Algorithm names used are determined by the Credential format and are defined in (#format_profiles)", so maybe we can make this:

```suggestion
  * `proofs_supported`: OPTIONAL. Object that describes specifics of the key proof(s) that the Credential Issuer supports. This object contains a list of name/value pairs, where each name is a unique identifier of the supported proof type being described. This identifier is used in the Credential Request defined in (#credential_request) to communicate to the Wallet specifics about the proof of possession of the cryptographic key material. A list of valid key proof type values defined by this specification can be found in (#proof_types). The value in the name/value pair is an object of which the value is determined by the Key Proof Type and are defined in (#key-proof-types).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1870,6 +1871,8 @@ When the `format` value is `jwt_vc_json`, the entire Credential Offer, Authoriza
 
 #### Credential Issuer Metadata {#server_metadata_jwt_vc_json}
 
+Cryptographic algorithm names used in the `cryptographic_suites_supported` parameter SHOULD be one of those defined in [@IANA.JOSE.ALGS].
```
TimoGlastra, line 1874:
```suggestion
Cryptographic algorithm names used in the `credential_signing_alg_values_supported` parameter SHOULD be one of those defined in [@IANA.JOSE.ALGS].
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1932,6 +1935,8 @@ Note: Data Integrity used to be called Linked Data Proofs, hence the "ldp" in th
 
 #### Credential Issuer Metadata {#server_metadata_ldp_vc}
 
+Cryptographic algorithm names used in the `cryptographic_suites_supported` parameter SHOULD be one of those defined in [@LD_Suite_Registry].
```
TimoGlastra, line 1938:
```suggestion
Cryptographic algorithm names used in the `credential_signing_alg_values_supported` parameter SHOULD be one of those defined in [@LD_Suite_Registry].
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2025,6 +2030,8 @@ The Credential format identifier is `mso_mdoc`.
 
 ### Credential Issuer Metadata {#server_metadata_mso_mdoc}
 
+Cryptographic algorithm names used in the `cryptographic_suites_supported` parameter SHOULD be one of those defined in [@!ISO.18013-5].
```
TimoGlastra, line 2033:
```suggestion
Cryptographic algorithm names used in the `credential_signing_alg_values_supported` parameter SHOULD be one of those defined in [@!ISO.18013-5].
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2076,6 +2083,8 @@ The Credential format identifier is `vc+sd-jwt`.
 
 ### Credential Issuer Metadata {#server_metadata_sd_jwt_vc}
 
+Cryptographic algorithm names used in the `cryptographic_suites_supported` parameter SHOULD be one of those defined in [@IANA.JOSE.ALGS].
```
TimoGlastra, line 2086:
```suggestion
Cryptographic algorithm names used in the `credential_signing_alg_values_supported` parameter SHOULD be one of those defined in [@IANA.JOSE.ALGS].
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2285,6 +2294,9 @@ Wallet Providers may also provide a marketplace where Issuers can register to be
    
    -13
 
+   * change the structure of `proof_types` from an array to a `proofs_supported` map that contains a required `proof_signing_alg_values_supported` parameter
+   * renamed `credential_configurations_supported` to `credential_signing_alg_values_supported` to clarify the purpose of the parameter
```
TimoGlastra, line 2298:
```suggestion
   * renamed `cryptographic_suites_supported` to `credential_signing_alg_values_supported` to clarify the purpose of the parameter
```

openid-4-verifiable-credential-issuance-1_0.md
TimoGlastra: If we do want to keep `proof_signing_alg_values_supported`, I think it at least makes sense to change

> Algorithm names used SHOULD be one of those defined in [@IANA.JOSE.ALGS], [@LD_Suite_Registry] or [@!ISO.18013-5].

to 

> Algorithm names used are determined by the Key Proof Type and are defined in (#key_proof_types).

And add the following sentence to each of the key proof types (with the correct spec):

> Cryptographic algorithm names used in the `proof_signing_alg_values_supported` parameter SHOULD be one of those defined in [@IANA.JOSE.ALGS].


This makes it consistent with `credential_signing_alg_values_supported`

Sakurann: this differentiation is on purpose because the algs in the proof types are dependent on the proof type (jwt, jdp, cwt) ad not the credential typ, so it is not accurate to point to the credential profiles for the algs used in the proofs

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: great catch - thank you so much!!

openid-4-verifiable-credential-issuance-1_0.md
TimoGlastra: > so it is not accurate to point to the credential profiles for the algs

I don't mean for them to point to the credential profiles, but to the Key Proof Types section where the actual proof types are defined: https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-7.2.1

So Section 7.2.1.1 will mention:

> Cryptographic algorithm names used in the `proof_signing_alg_values_supported` parameter for this supported proof type SHOULD be one of those defined in [@IANA.JOSE.ALGS].

And Section 7.2.1.2 will mention:

> Cryptographic algorithm names used in the `proof_signing_alg_values_supported` parameter for this supported proof type SHOULD be one of those defined in [@LD_Suite_Registry].

decentralgabe (approved):
lgtm!

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1352,9 +1352,10 @@ This specification defines the following Credential Issuer Metadata parameters:
...
+  * `credential_signing_alg_values_supported`: OPTIONAL. Array of case sensitive strings that identify the algorithms that the Issuer uses to sign the issued Credential. Algorithm names used are determined by the Credential format and are defined in (#format_profiles).
+  * `proofs_supported`: OPTIONAL. Object that describes specifics of the key proof(s) that the Credential Issuer supports. This object contains a list of name/value pairs, where each name is a unique identifier of the supported proof type being described. This identifier is used in the Credential Request defined in (#credential_request) to communicate to the Wallet specifics about the proof of possession of the cryptographic key material. A list of valid key proof type values defined by this specification can be found in (#proof_types). The value in the name/value pair is an object that contains metadata about the key proof and contains the following parameters defined by this specification:
```
tlodderstedt, line 1357: why `proofs_supported` and not just keep `proof_types`?

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt, line 1358:
```suggestion
    * `proof_signing_alg_values_supported`: REQUIRED. Array of case sensitive strings that identify the algorithms that the Issuer supports to sign the respective proof type. Algorithm names used SHOULD be one of those defined in [@IANA.JOSE.ALGS], [@LD_Suite_Registry] or [@!ISO.18013-5].
```
It's not the Issuer who signs the proofs, it's the wallet

jogu: I agree with Torsten's suggestion and suggest being even more explicit (also a slight simplification/improvement to the grammar):

```suggestion
    * `proof_signing_alg_values_supported`: REQUIRED. Array of case sensitive strings that identify the algorithms supported by the Issuer for this proof type, one of which the Wallet can select to sign the proof. Algorithm names used SHOULD be those defined in [@IANA.JOSE.ALGS], [@LD_Suite_Registry] or [@!ISO.18013-5].
```

jogu:
I can't figure out how to make a suggestion but this text on line 713 needs to fixed as it still mentions `proof_types`:

> The `proof` object is REQUIRED if the `proof_types` parameter is non-empty and present in the `credential_configurations_supported` parameter of the Issuer metadata for the requested Credential.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: The object includes not just information about the proof type but also the algs so more general name seemed more accurate.

jogu (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
Sakurann: `It is used to communicate to the Wallet the details of the key proofs the Issuer accepts.` this sentence seems to say the same thing as the first sentence `Object that describes specifics of the key proof(s) that the Credential Issuer supports.` so I took it out.

also type values defined in #proof_types section are technically, `non-exhaustive`.

```suggestion
  * `proofs_supported`: OPTIONAL. Object that describes specifics of the key proof(s) that the Credential Issuer supports. This object contains a list of name/value pairs, where each name is a unique identifier of the supported proof type(s). Valid values are defined in (#proof_types), other values MAY be used. This identifier is also used by the Wallet in the Credential Request as defined in (#credential_request). The value in the name/value pair is an object that contains metadata about the key proof and contains the following parameters defined by this specification:
```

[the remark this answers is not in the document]
jogu: @Sakurann Your suggestion looks good to me.

tlodderstedt: Aren't these signature algs nevertheless applicable for a certain type of proofs? I mean, that's why we moved them underneath the proof_type.

Sakurann: that is true. Would `proof_types_supported` work?

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: ok. that makes sense - thank you for clarifying. Happy to make this change if there is an agreement on the WG call (just to prevent the double work of potentially reverting.)

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: This one is not applied yet. The history text sends wrong?

openid-4-verifiable-credential-issuance-1_0.md
jogu: I'm fine with `proof_types_supported`

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1352,9 +1352,10 @@ This specification defines the following Credential Issuer Metadata parameters:
...
+  * `credential_signing_alg_values_supported`: OPTIONAL. Array of case sensitive strings that identify the algorithms that the Issuer uses to sign the issued Credential. Algorithm names used are determined by the Credential format and are defined in (#format_profiles).
+  * `proofs_supported`: OPTIONAL. Object that describes specifics of the key proof(s) that the Credential Issuer supports. This object contains a list of name/value pairs, where each name is a unique identifier of the supported proof type(s). Valid values are defined in (#proof_types), other values MAY be used. This identifier is also used by the Wallet in the Credential Request as defined in (#credential_request). The value in the name/value pair is an object that contains metadata about the key proof and contains the following parameters defined by this specification:
```
bc-pi, line 1357: `proof_types_supported` (here and in the examples) I think was the agreed on name from the call

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: Discussed in WG call on 31 Jan (NZ). People in the call leaned towards moving the proof type specific algs to the key proof type section.

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: sorry if this is redundant w/ other comments/threads

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: WG call on 31st Of Jan (NZ): consensus for `proof_types_supported`. Additionally, section 7.2.1's title is changed to "proof types".

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: changes applied

tlodderstedt (approved):

TimoGlastra (approved):

bc-pi (approved):

jogu (approved):

PR #243, opened 2024-01-30
Editorial

### LINKED ISSUE #230: Confusing sentence about @context in "VC Secured using Data Integrity" section
In the "VC Secured using Data Integrity" section (Appendix A.1.2.2) it currently says:

> It is recommended to define an `@context` value to communicate additional information such as which claims are mandatory-to-be-issued, type of claim value (i.e., string, number, etc.), the display properties of a Credential, and the order of the claim values when displayed as in (#vc-jwt).

I don't really know what this sentence means, but the "as in (#vc-jwt)" part particularly confuses me as that section doesn't mention `@context` at all.

### LINKED ISSUE #231: 'claims' definition for mdl authorization_details looks wrong
The spec currently has this in https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#appendix-A.2.3  :

> * `claims`: OPTIONAL. Object as defined in (#server_metadata_mso_mdoc).

I don't believe this is correct as for example I don't think you should include the 'display' property in authorization_details:

```
[
    {
        "type":"openid_credential",
        "credential_configuration_id": "org.iso.18013.5.1.mDL",
        "claims": {
            "org.iso.18013.5.1": {
                "given_name": {
                      "display": [
                            {
                                "name": "Given Name",
                                "locale": "en-US"
                            },
                }
            },
        }
    }
]
```

The sections for other credential formats that have a `claims` parameter look similarly wrong.

### LINKED ISSUE #237: cryptographic_binding_methods_supported value "mso" is used in mdl issuer metadata example but not defined in spec
https://github.com/openid/OpenID4VCI/blob/main/examples/credential_metadata_mso_mdoc.json uses `cryptographic_binding_methods_supported` value `mso` - however the VCI specification only defines the value `cose_key`:

> Support for keys expressed as a COSE Key object [[RFC8152](file:///Users/joseph/Downloads/openid-4-verifiable-credential-issuance-wg-draft.html#RFC8152)] (for example, used in [[ISO.18013-5](file:///Users/joseph/Downloads/openid-4-verifiable-credential-issuance-wg-draft.html#ISO.18013-5)]) is indicated by the value cose_key

### LINKED ISSUE #239: cryptographic_binding_methods_supported text about supporting all DID methods is unclear
Spec currently sayings this when defined `cryptographic_binding_methods_supported`:

> Support for all DID methods listed in Section 13 of [@DID_Specification_Registries] is indicated by sending a DID without any method-name

I don't understand what 'sending a DID without any method-name' means. My guess would be:

> by sending the string `did:`.

However the DID Specification Registries is I think a living document, so as there is no reference to a particular dated version of the registry I think it is never clear what "all DID methods" means and hence I'd suggest just deleting this sentence entirely.

### LINKED ISSUE #240: Unreferenced examples
As discussed in last DCP call, there are some examples in the examples folder of this repository that are currently not referenced in the spec.

This is the list of currently not referenced examples I found:
- credential_request_jwt_vc_json-ld.json
- credential_request_iso_mdl.json
- credential_offer_pre-authz_code.json
- credential_offer_sd_jwt_vc.json
- credential_request_jwt_vc_json.json

Should I create a PR removing these or are there some of these we want to reference in the spec / want them to stay in the repo?

### LINKED ISSUE #241: Inexact sentence on invoking the Token Endpoint
In https://github.com/openid/OpenID4VCI/blob/f9a59ce20ef00de031b2e7e4801b239e5723b8e7/openid-4-verifiable-credential-issuance-1_0.md?plain=1#L564 it is assumed that the `Token Endpoint` is invoked only after a successful Authorization Response, while if using the Pre-Authorized Code Flow, the endpoint is invoked right after the Credential Offer.

It is a minor thing, but it may help clarifying this.

### LINKED ISSUE #247: Possible typo in Annex A
The Annex A, part A1.1.4 (and others) contains this two sentences:

"The following additional parameters are defined for Credential Requests and this Credential format.

credential_definition: REQUIRED when the credential_identifier was not present in the Credential Request. "

I guess the last term shall be "Token Response".

Sakurann:
fixes #241 and addresses #240, resolves #237, resolves #239, fixes #231, fixes #230. resolves #247 

one thing worth pointing out is that the PR removes an option to indicate support for all DID methods.


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1945,7 +1945,7 @@ The following additional Credential Issuer metadata parameters are defined for t
           * `locale`: OPTIONAL. String value that identifies language of this object represented as language tag values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
 * `order`: OPTIONAL. Array of the claim name values that lists them in the order they should be displayed by the Wallet.
 
-It is recommended to define an `@context` value to communicate additional information such as which claims are mandatory-to-be-issued, type of claim value (i.e., string, number, etc.), the display properties of a Credential, and the order of the claim values when displayed as in (#jwt-vc-json).
+It is recommended to define an `@context` value to communicate additional information such as which claims are mandatory-to-be-issued, type of claim value (i.e., string, number, etc.), the display properties of a Credential, and the order of the claim values when displayed.
```
jogu, line 1948: I don't understand the reason for this recommendation or I don't understand what this recommendation is trying to tell the reader to do; it says the purpose of defining a `@context` value is "to communicate additional information" but all the information listed can, as far as I can see, be communicated in the issuer metadata. Is it saying you should define which claims are mandatory both in an @context and in the issuer metadata?

I'm not really familiar with `@context`, does the example in this section follow that recommendation? It doesn't appear to as https://www.w3.org/2018/credentials/examples/v1 doesn't define any display properties I think.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1352,7 +1352,7 @@ This specification defines the following Credential Issuer Metadata parameters:
...
+  * `cryptographic_binding_methods_supported`: OPTIONAL. Array of case sensitive strings that identify how the Credential is bound to the identifier of the End-User who possesses the Credential, as defined in (#credential-binding). Support for keys in JWK format [@!RFC7517] is indicated by the value `jwk`. Support for keys expressed as a COSE Key object [@!RFC8152] (for example, used in [@!ISO.18013-5]) is indicated by the value `cose_key`. When the Cryptographic Binding Method is a DID, valid values are a `did:` prefix followed by a method-name using a syntax as defined in Section 3.1 of [@!DID-Core], but without a `:`and method-specific-id. For example, support for the DID method with a method-name "example" would be represented by `did:example`. Support for all DID methods listed in Section 13 of [@DID_Specification_Registries] at that moment in time is indicated by sending a value `did` without any method-name.
```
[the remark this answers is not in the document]
Sakurann: yes, this was added because there were implementers who asked for it. For example, those who integrate universal resolver code into their application can instantly handle all the DID methods supported by the universal resolver. If an implementer implements 1-5 DID methods, they can list all those, but if there are 20+, they don't want to.

Honestly, I don't think anyone wants/will to support all of the DID methods in the DID registry, but there is a need for a parameter for those who support "all of the most used DID methods" and we can ask people to define alias values that stand for a specific combination of certain DID methods, but that is not very realistic, so `did` is probably most realistic, usable thing.

cc @nklomp

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: when @context is used, I would imagine that would take precedence over issuer metadata regarding type of claim value, mandatory or not, etc. not sure about display and order. but honestly, I am ok deleting it.

```suggestion
```

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
nklomp: IMO the generic `did` support will remain problematic, even now it is restricted to the spec registry. You will always have parties on any side that do not implement everything like @jogu mentioned, or which think they implement all of it, but the spec registry changed in the mean time.
It would be very wise to only allow specific did methods. We have seen in the wild, that implementers just choose  `did` because the spec allows for it and it is easy, leading to interop issues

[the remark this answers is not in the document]
nklomp: Also people using the Universal Resolver, should be running their own instance anyway, and they really should restrict it to a few DID methods, as many DID specs/implementations are pretty poor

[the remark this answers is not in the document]
Sakurann:
```suggestion
  * `cryptographic_binding_methods_supported`: OPTIONAL. Array of case sensitive strings that identify how the Credential is bound to the identifier of the End-User who possesses the Credential, as defined in (#credential-binding). Support for keys in JWK format [@!RFC7517] is indicated by the value `jwk`. Support for keys expressed as a COSE Key object [@!RFC8152] (for example, used in [@!ISO.18013-5]) is indicated by the value `cose_key`. When the Cryptographic Binding Method is a DID, valid values are a `did:` prefix followed by a method-name using a syntax as defined in Section 3.1 of [@!DID-Core], but without a `:`and method-specific-id. For example, support for the DID method with a method-name "example" would be represented by `did:example`.
```

tlodderstedt (approved):
lgtm - please note I would prefer PRs focusing on a single topic. That makes reviewing them easier.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2097,7 +2096,7 @@ The following is a non-normative example of an object comprising the `credential
 The following additional claims are defined for authorization details of type `openid_credential` and this Credential format.
 
 * `vct`: REQUIRED. String as defined in (#server-metadata-sd-jwt-vc). This claim contains the type values the Wallet requests authorization for at the Credential Issuer. It MUST only be present if the `format` claim is present. It MUST not be present otherwise.
-* `claims`: OPTIONAL. An object as defined in (#server-metadata-sd-jwt-vc).
+* `claims`: OPTIONAL. An object as defined in (#server-metadata-sd-jwt-vc) excluding the `display` parameter.
```
jogu, line 2099: As we're only excluding `display`, did we conclude that the `mandatory` and `value_type` do have meaning if passed in the authorization details?

jogu:
#240 is not fully resolved as there are still two unreferenced examples (I think if they're important examples, we should be including them in the spec).

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* `claims`: OPTIONAL. An object as defined in (#server-metadata-sd-jwt-vc) excluding the `display` and `value_type` parameters. `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2045,7 +2044,7 @@ The following is a non-normative example of an object containing the `credential
 The following additional claims are defined for authorization details of type `openid_credential` and this Credential format.
 
 * `doctype`: OPTIONAL. String as defined in (#server-metadata-mso-mdoc). This claim contains the type value the Wallet requests authorization for at the Credential Issuer. It MUST only be present if the `format` claim is present. It MUST not be present otherwise. 
-* `claims`: OPTIONAL. Object as defined in (#server-metadata-mso-mdoc).
+* `claims`: OPTIONAL. Object as defined in (#server-metadata-mso-mdoc) excluding the `display` parameter.
```
Sakurann, line 2047:
```suggestion
* `claims`: OPTIONAL. Object as defined in (#server-metadata-sd-jwt-vc) excluding the `display` and `value_type` parameters. `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1958,7 +1957,7 @@ The following additional claims are defined for authorization details of type `o
 * `credential_definition`: OPTIONAL. Object containing the detailed description of the Credential consisting of the following parameter:
     * `@context`: OPTIONAL. Array as defined in (#server-metadata-ldp-vc). It MUST only be present if the `format` claim is present in the root of the authorization details object. It MUST not be present otherwise. 
     * `type`: OPTIONAL. Array as defined in (#server-metadata-ldp-vc).  This claim contains the type values the Wallet requests authorization for at the Credential Issuer. MUST only be present if the `@context` claim is present. 
-    * `credentialSubject`: OPTIONAL. Object as defined in (#authorization-jwt-vc-json).
+    * `credentialSubject`: OPTIONAL. Object as defined in (#authorization-jwt-vc-json) excluding the `display` parameter.
```
Sakurann, line 1960:
```suggestion
    * `credentialSubject`: OPTIONAL. Object as defined in (#server-metadata-sd-jwt-vc) excluding the `display` and `value_type` parameters. `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
```

jogu (approved):

PR #187, opened 2023-12-30
Wrote initial Privacy Considerations

selfissued:
TBD -> actual text!

PR #244, opened 2024-01-31
more detailed privacy considerations

Sakurann:
More detailed and thorough Privacy Considerations section for ID-1 inspired by implementation experience and a lot of related specifications. Not perfect, but should be good enough to convey how serious we are about privacy.
intended to replace #187.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1513,8 +1513,8 @@ The privacy principles of [@ISO.29100] should be adhered to.
 
 ## User Consent
 
-The Credential Issuer should obtain the End-User's consent before issuing Credential(s)
-to the Wallet. It should be made clear to the End-User what information is being included in the
+The Credential Issuer SHOULD obtain the End-User's consent before issuing Credential(s)
+to the Wallet. It SHOULD be made clear to the End-User what information is being included in the
 Credential(s) and for what purpose.
 
 ## Minimum Disclosure
@@ -1536,8 +1536,8 @@ for should be minimized.
 
 After Issuance, Credential Issuers SHOULD NOT store the Issuer-signed Credentials if they
 contain privacy-sensitive data. Wallets SHOULD store Credentials only in encrypted form, and,
-wherever possible, use hardware-backed encryption. Credentials that are no longer valid to present should be deleted as soon as
-possible.
+wherever possible, use hardware-backed encryption. Wallets SHOULD not store
+Credentials longer than needed.
 
 ## Correlation 
 
@@ -1552,6 +1552,8 @@ methods, including but not limited to the following ones:
 * Issue a batch of Credentials to enable the usage of a unique Credential per presentation or per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
 * Use cryptographic schemes that can provide non-correlation.
 
+Credential Issuers specifically SHOULD discard values that can be used in collusion with a Verifier to track a user, such as the Issuer's signature or cryptographic key material to which an issued credential was bound to.
+
 ### Credential Offer
 
 The Privacy Considerations in Section 11.2 of [@!RFC9101] apply to the `credential_offer` and
@@ -1564,9 +1566,10 @@ for example, by including clear-text session information as a `state` parameter 
 it in a `redirect_uri` parameter. A third party may observe such information through browser
 history, etc. and correlate the user's activity using it.
 
-## Credential Issuer Identifier
+## Identifier of the Credential Issuer
 
-The Credential Issuer Identifier alone may reveal information about the End-User.
+Information in the credential identifying a particular Credential Issuer, such as a Credential Issuer Identifier,
+issuer's certificate, or issuer's public key may reveal information about the End-User.
 
 For example, when a military organization or a drug rehabilitation center issues a vaccine
 credential, verifiers can deduce that the owner of the Wallet storing such Credential is a
@@ -1580,8 +1583,13 @@ For example, if the National Cancer Institute only issued Credentials with cance
 information, it is possible to deduce that the owner of the Wallet storing such Credential is a
 cancer patient.
 
-To mitigate this issue, a group of Credential Issuers may elect to use a common Credential Issuer
-Identifier. A group signature scheme may also be used, instead of an individual signature.
+To mitigate these issues, a group of organizations may elect to use a common Credential Issuer,
+such that any credentials issued by this Issuer cannot be attributed to a particular organization
+through identifiers of the Credential Issuers alone. A group signature scheme may also be used
+instead of an individual signature.
+
+When a common Credential Issuer is used, appropriate guardrails need to be in place to prevent
+one organization from issuing illegitimate credentials on behalf of other organizations.
 
 ## Custom URL Scheme
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1513,8 +1513,8 @@ The privacy principles of [@ISO.29100] should be adhered to.
 
 ## User Consent
 
-The Credential Issuer MUST obtain the End-User consent before issuing Credential(s) to the
-Wallet. It should be made clear to the End-User what information is being included in the
+The Credential Issuer should obtain the End-User's consent before issuing Credential(s)
+to the Wallet. It should be made clear to the End-User what information is being included in the
 Credential(s), for what purpose, and to which Wallet it is being issued to.
 
 ## Minimum Disclosure
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1549,7 +1549,7 @@ unique values encoded in the Credential (End-User claims, identifiers, Issuer si
 To prevent these types of correlation, Credential Issuers and Wallets SHOULD use
 methods, including but not limited to the following ones:
 
-* Issue a batch of Credentials to enable the usage a unique Credential per presentation or per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
+* Issue a batch of Credentials to enable the usage of a unique Credential per presentation or per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
 * Use cryptographic schemes that can provide non-correlation.
 
 ### Credential Offer
@@ -1560,7 +1560,7 @@ The Privacy Considerations in Section 11.2 of [@!RFC9101] apply to the `credenti
 ### Authorization Request
 
 The Wallet SHOULD NOT include potentially sensitive information in the Authorization Request,
-for example, that including clear-text session information as a `state` parameter value, or encoding
+for example, by including clear-text session information as a `state` parameter value or encoding
 it in a `redirect_uri` parameter. A third party may observe such information through browser
 history, etc. and correlate the user's activity using it.
 
@@ -1585,12 +1585,12 @@ Identifier. A group signature scheme may also be used, instead of an individual 
 
 ## Custom URL Scheme
 
-There is a potential for leaking information about the Wallet to the third parties when the
-Wallet reacts to a Credential Offer that was sent to the custom URL scheme that the Wallet
-supports. An attacker may send Credential Offers to different custom URL schemes, see if the
+There is a potential for leaking information about the Wallet to third parties when the
+Wallet reacts to a Credential Offer that was sent to a custom URL scheme that the Wallet
+supports. An attacker may send Credential Offers using different custom URL schemes, see if the
 Wallet reacts (e.g., whether the wallet retrieves Credential Issuer metadata hosted by an
 attacker's server), and, therefore, learn which Wallet is installed. To avoid this, the Wallet
-SHOULD require user interaction (or establish trust in the Issuer) before fetching the `credential_offer_uri ` or acting on the received Credential Offer.
+SHOULD require user interaction (or establish trust in the Issuer) before fetching any `credential_offer_uri ` or acting on the received Credential Offer.
 
 ## Untrusted Wallets
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1526,13 +1526,13 @@ Credential formats that support selective disclosure, or consider issuing a sepa
 for each user claim.
 
 If supported by the Credential format, the Credential Issuer SHOULD provide the End-User with
-the options to choose which claims are selectively disclosable.
+the option to choose which claims are selectively disclosable.
 
 ## Storage of the Credentials
 
 To prevent a leak of signed End-User data, which risks revealing private data of End-Users to
-third parties Systems implementing this specifications SHOULD be designed to minimize the amount
-of End-User data that is stored. All involved parties SHOULD store Verifiable Credentias
+third parties, systems implementing this specifications SHOULD be designed to minimize the amount
+of End-User data that is stored. All involved parties SHOULD store Verifiable Credentials
 containing privacy-sensitive data only for as long as needed, including in log files.
 
 After Issuance, Credential Issuers SHOULD NOT store the Issuer-signed Credentials if they
@@ -1550,11 +1550,11 @@ Colluding Issuer/Verifier or Verifier/Verifier pairs could link issuance/present
 presentation sessions to the same End-User on the basis of unique values encoded in the
 Credential (End-User claims, identifiers, Issuer signature, etc.).
 
-To prevent these types of correlation, Credential Issuers and the Wallets SHOULD use the
+To prevent these types of correlation, Credential Issuers and the Wallets SHOULD use these
 methods, including but not limited to the following ones:
 
 * Issue a batch of Credentials to enable the usage a unique Credential per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
-* Use advanced cryptographic schemes
+* Use cryptographic schemes that can provide non-correlation.
 
 ### Credential Offer
 
@@ -1566,7 +1566,7 @@ The Privacy Considerations in Section 11.2 of [@!RFC9101] apply to the `credenti
 The Wallet SHOULD NOT include potentially sensitive information in the Authorization Request,
 such as that including clear-text session information as a `state` parameter value, or encoding
 it in a `redirect_uri` parameter. A third party may observe such information through browser
-history, etc. and start correlating the user's activity using it.
+history, etc. and correlate the user's activity using it.
 
 ## Credential Issuer Identifier
 
@@ -1600,7 +1600,7 @@ SHOULD require user interaction before acting on the received Credential Offer.
 
 The Wallet transmits and stores sensitive information about the End-User. To ensure that the
 Wallet can handle those appropriately (i.e., according to a certain trust framework or a
-regulation), the Credential Issuer SHOULD request an Attestation from the Wallet using the
+regulation), the Credential Issuer SHOULD request an Attestation from the Wallet using
 methods, including but not limited to those defined in [@!I-D.ietf-oauth-attestation-based-client-auth]. 
 
 {backmatter}
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1515,7 +1515,7 @@ The privacy principles of [@ISO.29100] should be adhered to.
 
 The Credential Issuer should obtain the End-User's consent before issuing Credential(s)
 to the Wallet. It should be made clear to the End-User what information is being included in the
-Credential(s), for what purpose, and to which Wallet it is being issued to.
+Credential(s) and for what purpose.
 
 ## Minimum Disclosure
 
@@ -1525,35 +1525,31 @@ number of times by the End-User, the Credential Issuers and the Wallets SHOULD i
 Credential formats that support selective disclosure, or consider issuing a separate Credential
 for each user claim.
 
-If supported by the Credential format, the Credential Issuer SHOULD provide the End-User with
-the option to choose which claims are selectively disclosable.
-
 ## Storage of the Credentials
 
-To prevent a leak of signed End-User data, which risks revealing private data of End-Users to
+To prevent a leak of End-User data, especially when it is signed, which risks revealing private data of End-Users to
 third parties, systems implementing this specification SHOULD be designed to minimize the amount
 of End-User data that is stored. All involved parties SHOULD store Verifiable Credentials
-containing privacy-sensitive data only for as long as needed, including in log files.
+containing privacy-sensitive data only for as long as needed, including in log files. Any logging of
+End-User data should be carefully considered as to whether it is necessary at all. The time logs are retained
+for should be minimized.
 
 After Issuance, Credential Issuers SHOULD NOT store the Issuer-signed Credentials if they
 contain privacy-sensitive data. Wallets SHOULD store Credentials only in encrypted form, and,
-wherever possible, use hardware-backed encryption in particular for the private key used for Key
-Binding. Decentralized storage of data, e.g., on End-User devices, SHOULD be preferred for
-End-User Credentials over centralized storage. Expired Credentials SHOULD be deleted as soon as
+wherever possible, use hardware-backed encryption. Credentials that are no longer valid to present should be deleted as soon as
 possible.
 
 ## Correlation 
 
 ### Unique Values Encoded in the Credential
 
-Colluding Issuer/Verifier or Verifier/Verifier pairs could link issuance/presentation or two
-presentation sessions to the same End-User on the basis of unique values encoded in the
-Credential (End-User claims, identifiers, Issuer signature, etc.).
+Issuance/presentation or two presentation sessions by the same End-User can be linked on the basis of
+unique values encoded in the Credential (End-User claims, identifiers, Issuer signature, etc.) either by colluding Issuer/Verifier or Verifier/Verifier pairs, or by the same Verifier.
 
 To prevent these types of correlation, Credential Issuers and Wallets SHOULD use
 methods, including but not limited to the following ones:
 
-* Issue a batch of Credentials to enable the usage a unique Credential per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
+* Issue a batch of Credentials to enable the usage a unique Credential per presentation or per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
 * Use cryptographic schemes that can provide non-correlation.
 
 ### Credential Offer
@@ -1594,14 +1590,13 @@ Wallet reacts to a Credential Offer that was sent to the custom URL scheme that 
 supports. An attacker may send Credential Offers to different custom URL schemes, see if the
 Wallet reacts (e.g., whether the wallet retrieves Credential Issuer metadata hosted by an
 attacker's server), and, therefore, learn which Wallet is installed. To avoid this, the Wallet
-SHOULD require user interaction before acting on the received Credential Offer.
+SHOULD require user interaction (or establish trust in the Issuer) before fetching the `credential_offer_uri ` or acting on the received Credential Offer.
 
 ## Untrusted Wallets
 
 The Wallet transmits and stores sensitive information about the End-User. To ensure that the
 Wallet can handle those appropriately (i.e., according to a certain trust framework or a
-regulation), the Credential Issuer SHOULD request an Attestation from the Wallet using
-methods, including but not limited to those defined in [@!I-D.ietf-oauth-attestation-based-client-auth]. 
+regulation), the Credential Issuer should properly authenticate the Wallet and ensure it is a trusted entity. For more details, see (#trust-between-wallet-and-issuer).
 
 {backmatter}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1932,6 +1932,13 @@ methods, including but not limited to those defined in [@!I-D.ietf-oauth-attesta
   </front>
 </reference>
 
+<reference anchor="ISO.29100" target="https://standards.iso.org/ittf/PubliclyAvailableStandards/index.html">
+  <front>
+    <author fullname="ISO"></author>
+    <title>ISO/IEC 29100:2011 Information technology — Security techniques — Privacy framework</title>
+  </front>
+</reference>
+
 # Credential Format Profiles {#format-profiles}
 
 This specification defines several extension points to accommodate the differences across Credential formats. Sets of Credential format specific parameters or claims referred to as Credential format identifiers are identified by the Credential format identifier and used at these extension points.
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1591,7 +1591,7 @@ instead of an individual signature.
 When a common Credential Issuer is used, appropriate guardrails need to be in place to prevent
 one organization from issuing illegitimate credentials on behalf of other organizations.
 
-## Processing of Credential Offers by Wallet prior to User Interaction
+## Identifying the Wallet
 
 There is a potential for leaking information about the Wallet to third parties when the
 Wallet reacts to a Credential Offer. An attacker may send Credential Offers using different
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1566,7 +1566,7 @@ for example, by including clear-text session information as a `state` parameter 
 it in a `redirect_uri` parameter. A third party may observe such information through browser
 history, etc. and correlate the user's activity using it.
 
-## Identifier of the Credential Issuer
+## Identifying the Credential Issuer
 
 Information in the credential identifying a particular Credential Issuer, such as a Credential Issuer Identifier,
 issuer's certificate, or issuer's public key may reveal information about the End-User.
@@ -1591,14 +1591,15 @@ instead of an individual signature.
 When a common Credential Issuer is used, appropriate guardrails need to be in place to prevent
 one organization from issuing illegitimate credentials on behalf of other organizations.
 
-## Custom URL Scheme
+## Processing of Credential Offers by Wallet prior to User Interaction
 
 There is a potential for leaking information about the Wallet to third parties when the
-Wallet reacts to a Credential Offer that was sent to a custom URL scheme that the Wallet
-supports. An attacker may send Credential Offers using different custom URL schemes, see if the
+Wallet reacts to a Credential Offer. An attacker may send Credential Offers using different
+custom URL schemes or claimed https urls, see if the
 Wallet reacts (e.g., whether the wallet retrieves Credential Issuer metadata hosted by an
-attacker's server), and, therefore, learn which Wallet is installed. To avoid this, the Wallet
-SHOULD require user interaction (or establish trust in the Issuer) before fetching any `credential_offer_uri ` or acting on the received Credential Offer.
+attacker's server), and, therefore, learn which Wallet is installed. To avoid this, the Wallet SHOULD
+require user interaction or establish trust in the Issuer before fetching any `credential_offer_uri `
+or acting on the received Credential Offer.
 
 ## Untrusted Wallets
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1507,7 +1507,7 @@ It is up to the Credential Issuer whether to update both the signature and the c
 
 # Privacy Considerations
 
-When [@!RFC9396] is used, the Privacy Considerations of the specification apply.
+When [@!RFC9396] is used, the Privacy Considerations of that specification also apply.
 
 The privacy principles of [@ISO.29100] should be adhered to.
 
@@ -1519,8 +1519,8 @@ Credential(s), for what purpose, and to which Wallet it is being issued to.
 
 ## Minimum Disclosure
 
-To ensure minimum disclosure and prevent Verifiers from obtaining claims irrelevant for the
-transaction at hand, when issuing Credentials that are meant to be created once and then used a
+To ensure minimum disclosure and prevent Verifiers from obtaining claims unnecessary for the
+transaction at hand, when issuing Credentials that are intended to be created once and then used a
 number of times by the End-User, the Credential Issuers and the Wallets SHOULD implement
 Credential formats that support selective disclosure, or consider issuing a separate Credential
 for each user claim.
@@ -1531,7 +1531,7 @@ the option to choose which claims are selectively disclosable.
 ## Storage of the Credentials
 
 To prevent a leak of signed End-User data, which risks revealing private data of End-Users to
-third parties, systems implementing this specifications SHOULD be designed to minimize the amount
+third parties, systems implementing this specification SHOULD be designed to minimize the amount
 of End-User data that is stored. All involved parties SHOULD store Verifiable Credentials
 containing privacy-sensitive data only for as long as needed, including in log files.
 
@@ -1550,7 +1550,7 @@ Colluding Issuer/Verifier or Verifier/Verifier pairs could link issuance/present
 presentation sessions to the same End-User on the basis of unique values encoded in the
 Credential (End-User claims, identifiers, Issuer signature, etc.).
 
-To prevent these types of correlation, Credential Issuers and the Wallets SHOULD use these
+To prevent these types of correlation, Credential Issuers and Wallets SHOULD use
 methods, including but not limited to the following ones:
 
 * Issue a batch of Credentials to enable the usage a unique Credential per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
@@ -1564,7 +1564,7 @@ The Privacy Considerations in Section 11.2 of [@!RFC9101] apply to the `credenti
 ### Authorization Request
 
 The Wallet SHOULD NOT include potentially sensitive information in the Authorization Request,
-such as that including clear-text session information as a `state` parameter value, or encoding
+for example, that including clear-text session information as a `state` parameter value, or encoding
 it in a `redirect_uri` parameter. A third party may observe such information through browser
 history, etc. and correlate the user's activity using it.
 
@@ -1576,7 +1576,7 @@ For example, when a military organization or a drug rehabilitation center issues
 credential, verifiers can deduce that the owner of the Wallet storing such Credential is a
 military member or may have a substance use disorder.
 
-In addition, when Credential Issuer issues only one type of Credential, it might have privacy implications,
+In addition, when a Credential Issuer issues only one type of Credential, it might have privacy implications,
 because if the Wallet has a Credential issued by that Issuer, its type and claim names can be
 determined.
 
@@ -1596,7 +1596,7 @@ Wallet reacts (e.g., whether the wallet retrieves Credential Issuer metadata hos
 attacker's server), and, therefore, learn which Wallet is installed. To avoid this, the Wallet
 SHOULD require user interaction before acting on the received Credential Offer.
 
-## Wallet Attestation
+## Untrusted Wallets
 
 The Wallet transmits and stores sensitive information about the End-User. To ensure that the
 Wallet can handle those appropriately (i.e., according to a certain trust framework or a
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
 
 # Privacy Considerations
 
-TBD
+When [@!RFC9396] is used, the Privacy Considerations of the specification apply.
+
+The privacy principles of [@ISO.29100] should be adhered to.
+
+## User Consent
+
+The Credential Issuer MUST obtain the End-User consent before issuing Credential(s) to the
+Wallet. It should be made clear to the End-User what information is being included in the
+Credential(s), for what purpose, and to which Wallet it is being issued to.
+
+## Minimum Disclosure
+
+To ensure minimum disclosure and prevent Verifiers from obtaining claims irrelevant for the
+transaction at hand, when issuing Credentials that are meant to be created once and then used a
+number of times by the End-User, the Credential Issuers and the Wallets SHOULD implement
+Credential formats that support selective disclosure, or consider issuing a separate Credential
+for each user claim.
+
+If supported by the Credential format, the Credential Issuer SHOULD provide the End-User with
+the options to choose which claims are selectively disclosable.
+
+## Storage of the Credentials
+
+To prevent a leak of signed End-User data, which risks revealing private data of End-Users to
+third parties Systems implementing this specifications SHOULD be designed to minimize the amount
+of End-User data that is stored. All involved parties SHOULD store Verifiable Credentias
+containing privacy-sensitive data only for as long as needed, including in log files.
+
+After Issuance, Credential Issuers SHOULD NOT store the Issuer-signed Credentials if they
+contain privacy-sensitive data. Wallets SHOULD store Credentials only in encrypted form, and,
+wherever possible, use hardware-backed encryption in particular for the private key used for Key
+Binding. Decentralized storage of data, e.g., on End-User devices, SHOULD be preferred for
+End-User Credentials over centralized storage. Expired Credentials SHOULD be deleted as soon as
+possible.
+
+## Correlation 
+
+### Unique Values Encoded in the Credential
+
+Colluding Issuer/Verifier or Verifier/Verifier pairs could link issuance/presentation or two
+presentation sessions to the same End-User on the basis of unique values encoded in the
+Credential (End-User claims, identifiers, Issuer signature, etc.).
+
+To prevent these types of correlation, Credential Issuers and the Wallets SHOULD use the
+methods, including but not limited to the following ones:
+
+* Issue a batch of Credentials to enable the usage a unique Credential per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+## Minimum Disclosure
+
+To ensure minimum disclosure and prevent Verifiers from obtaining claims irrelevant for the
+transaction at hand, when issuing Credentials that are meant to be created once and then used a
+number of times by the End-User, the Credential Issuers and the Wallets SHOULD implement
+Credential formats that support selective disclosure, or consider issuing a separate Credential
+for each user claim.
+
+If supported by the Credential format, the Credential Issuer SHOULD provide the End-User with
+the options to choose which claims are selectively disclosable.
```
selfissued, line 1529:
```suggestion
the option to choose which claims are selectively disclosable.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+## Storage of the Credentials
+
+To prevent a leak of signed End-User data, which risks revealing private data of End-Users to
+third parties Systems implementing this specifications SHOULD be designed to minimize the amount
```
selfissued, line 1534:
```suggestion
third parties, Systems implementing this specifications SHOULD be designed to minimize the amount
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+of End-User data that is stored. All involved parties SHOULD store Verifiable Credentias
```
selfissued, line 1535:
```suggestion
of End-User data that is stored. All involved parties SHOULD store Verifiable Credentials
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+## Correlation 
+
+### Unique Values Encoded in the Credential
+
+Colluding Issuer/Verifier or Verifier/Verifier pairs could link issuance/presentation or two
+presentation sessions to the same End-User on the basis of unique values encoded in the
+Credential (End-User claims, identifiers, Issuer signature, etc.).
+
+To prevent these types of correlation, Credential Issuers and the Wallets SHOULD use the
```
selfissued, line 1553:
```suggestion
To prevent these types of correlation, Credential Issuers and the Wallets SHOULD use these
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+methods, including but not limited to the following ones:
+
+* Issue a batch of Credentials to enable the usage a unique Credential per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
+* Use advanced cryptographic schemes
```
selfissued, line 1557:
```suggestion
* Use cryptographic schemes providing non-correlation
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+The Privacy Considerations in Section 11.2 of [@!RFC9101] apply to the `credential_offer` and
+`credential_offer_uri` parameters defined in (#credential-offer).
+
+### Authorization Request
+
+The Wallet SHOULD NOT include potentially sensitive information in the Authorization Request,
+such as that including clear-text session information as a `state` parameter value, or encoding
+it in a `redirect_uri` parameter. A third party may observe such information through browser
+history, etc. and start correlating the user's activity using it.
```
selfissued, line 1569:
```suggestion
history, etc. and correlate the user's activity using it.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+supports. An attacker may send Credential Offers to different custom URL schemes, see if the
+Wallet reacts (e.g., whether the wallet retrieves Credential Issuer metadata hosted by an
+attacker's server), and, therefore, learn which Wallet is installed. To avoid this, the Wallet
+SHOULD require user interaction before acting on the received Credential Offer.
+
+## Wallet Attestation
+
+The Wallet transmits and stores sensitive information about the End-User. To ensure that the
+Wallet can handle those appropriately (i.e., according to a certain trust framework or a
+regulation), the Credential Issuer SHOULD request an Attestation from the Wallet using the
```
selfissued, line 1603:
```suggestion
regulation), the Credential Issuer SHOULD request an Attestation from the Wallet using
```

selfissued (approved):
Approved, pending my suggestions being accepted.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
third parties, systems implementing this specifications SHOULD be designed to minimize the amount
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* Use cryptographic schemes that can provide non-correlation.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
 # Privacy Considerations
 
-TBD
+When [@!RFC9396] is used, the Privacy Considerations of the specification apply.
+
+The privacy principles of [@ISO.29100] should be adhered to.
+
+## User Consent
+
+The Credential Issuer MUST obtain the End-User consent before issuing Credential(s) to the
```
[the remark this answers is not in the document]
tlodderstedt: Whether a user consent is required depends on legal requirements. So far we (OIDF) have tried to stay out of that field. So I'm not sure this should be a MUST.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+### Unique Values Encoded in the Credential
+
+Colluding Issuer/Verifier or Verifier/Verifier pairs could link issuance/presentation or two
+presentation sessions to the same End-User on the basis of unique values encoded in the
+Credential (End-User claims, identifiers, Issuer signature, etc.).
+
+To prevent these types of correlation, Credential Issuers and the Wallets SHOULD use these
+methods, including but not limited to the following ones:
+
+* Issue a batch of Credentials to enable the usage a unique Credential per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
```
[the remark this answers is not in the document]
Sakurann: I think it's both.

```suggestion
* Issue a batch of Credentials to enable the usage a unique Credential per presentation or per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+### Credential Offer
+
+The Privacy Considerations in Section 11.2 of [@!RFC9101] apply to the `credential_offer` and
+`credential_offer_uri` parameters defined in (#credential-offer).
+
+### Authorization Request
+
+The Wallet SHOULD NOT include potentially sensitive information in the Authorization Request,
+such as that including clear-text session information as a `state` parameter value, or encoding
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
for example, that including clear-text session information as a `state` parameter value, or encoding
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
-TBD
+When [@!RFC9396] is used, the Privacy Considerations of the specification apply.
+
+The privacy principles of [@ISO.29100] should be adhered to.
+
+## User Consent
+
+The Credential Issuer MUST obtain the End-User consent before issuing Credential(s) to the
+Wallet. It should be made clear to the End-User what information is being included in the
+Credential(s), for what purpose, and to which Wallet it is being issued to.
```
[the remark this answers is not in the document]
Sakurann: umm, I think knowing that the issuer is issuing to an unknown wallet is part of knowing `to which wallet it is being issued to`.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
 
-TBD
+When [@!RFC9396] is used, the Privacy Considerations of that specification also apply.
+
+The privacy principles of [@ISO.29100] should be adhered to.
+
+## User Consent
+
+The Credential Issuer MUST obtain the End-User consent before issuing Credential(s) to the
+Wallet. It should be made clear to the End-User what information is being included in the
```
Sakurann, line 1517:
```suggestion
The Credential Issuer should obtain the End-User consent before issuing Credential(s) to the End-User. It should be made clear to the End-User what information is being included in the
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+## Minimum Disclosure
+
+To ensure minimum disclosure and prevent Verifiers from obtaining claims unnecessary for the
+transaction at hand, when issuing Credentials that are intended to be created once and then used a
+number of times by the End-User, the Credential Issuers and the Wallets SHOULD implement
+Credential formats that support selective disclosure, or consider issuing a separate Credential
+for each user claim.
+
+If supported by the Credential format, the Credential Issuer SHOULD provide the End-User with
+the option to choose which claims are selectively disclosable.
```
Sakurann, line 1529:
```suggestion
If supported by the Credential format, the Credential Issuer should provide the End-User with
the option to choose which claims are selectively disclosable.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+## Storage of the Credentials
+
+To prevent a leak of signed End-User data, which risks revealing private data of End-Users to
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
To prevent a leak of End-User data, especially when it is signed, which risks revealing private data of End-Users to
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+third parties, systems implementing this specifications SHOULD be designed to minimize the amount
+of End-User data that is stored. All involved parties SHOULD store Verifiable Credentials
+containing privacy-sensitive data only for as long as needed, including in log files.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
containing privacy-sensitive data only for as long as needed, including in log files. Any logging of
End-User data should be carefully considered, whether it is necessary at all. The time logs are retained
for should be minimized.
```

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
jogu: I think we should be explicit whether this means the Issuer has to tell the user that they don't know which Wallet the credential will be sent to then, or if in that case they should not say anything.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+After Issuance, Credential Issuers SHOULD NOT store the Issuer-signed Credentials if they
+contain privacy-sensitive data. Wallets SHOULD store Credentials only in encrypted form, and,
+wherever possible, use hardware-backed encryption in particular for the private key used for Key
+Binding. Decentralized storage of data, e.g., on End-User devices, SHOULD be preferred for
+End-User Credentials over centralized storage. Expired Credentials SHOULD be deleted as soon as
```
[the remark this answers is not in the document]
Sakurann: + replace `expired` with `credentials that are no longer valid to present`.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
-TBD
+When [@!RFC9396] is used, the Privacy Considerations of that specification also apply.
+
+The privacy principles of [@ISO.29100] should be adhered to.
+
+## User Consent
+
+The Credential Issuer should obtain the End-User's consent before issuing Credential(s)
+to the Wallet. It should be made clear to the End-User what information is being included in the
+Credential(s), for what purpose, and to which Wallet it is being issued to.
```
Sakurann, line 1518:
```suggestion
to the Wallet. It should be made clear to the End-User what information is being included in the
Credential(s) and for what purpose.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+## Minimum Disclosure
+
+To ensure minimum disclosure and prevent Verifiers from obtaining claims irrelevant for the
+transaction at hand, when issuing Credentials that are meant to be created once and then used a
+number of times by the End-User, the Credential Issuers and the Wallets SHOULD implement
+Credential formats that support selective disclosure, or consider issuing a separate Credential
+for each user claim.
+
+If supported by the Credential format, the Credential Issuer SHOULD provide the End-User with
+the option to choose which claims are selectively disclosable.
```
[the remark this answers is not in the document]
Sakurann: @danielfett I think this is the sentence you have proposed (sorry if I am wrong). removing for now. please open an issue or PR if you care.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+To ensure minimum disclosure and prevent Verifiers from obtaining claims unnecessary for the
+transaction at hand, when issuing Credentials that are intended to be created once and then used a
+number of times by the End-User, the Credential Issuers and the Wallets SHOULD implement
+Credential formats that support selective disclosure, or consider issuing a separate Credential
+for each user claim.
+
+If supported by the Credential format, the Credential Issuer SHOULD provide the End-User with
+the option to choose which claims are selectively disclosable.
+
```
Sakurann, line 1530:
```suggestion
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+To prevent a leak of signed End-User data, which risks revealing private data of End-Users to
+third parties, systems implementing this specification SHOULD be designed to minimize the amount
+of End-User data that is stored. All involved parties SHOULD store Verifiable Credentials
+containing privacy-sensitive data only for as long as needed, including in log files.
+
+After Issuance, Credential Issuers SHOULD NOT store the Issuer-signed Credentials if they
+contain privacy-sensitive data. Wallets SHOULD store Credentials only in encrypted form, and,
+wherever possible, use hardware-backed encryption in particular for the private key used for Key
+Binding. Decentralized storage of data, e.g., on End-User devices, SHOULD be preferred for
+End-User Credentials over centralized storage. Expired Credentials SHOULD be deleted as soon as
```
Sakurann, line 1542:
```suggestion
wherever possible, use hardware-backed encryption. Credentials that are no longer valid to present should be deleted as soon as
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+possible.
+
+## Correlation 
+
+### Unique Values Encoded in the Credential
+
+Colluding Issuer/Verifier or Verifier/Verifier pairs could link issuance/presentation or two
+presentation sessions to the same End-User on the basis of unique values encoded in the
+Credential (End-User claims, identifiers, Issuer signature, etc.).
```
Sakurann, line 1551:
```suggestion
Issuance/presentation or two presentation sessions by the same End-User can be linked on the basis of
unique values encoded in the Credential (End-User claims, identifiers, Issuer signature, etc.) either by colluding Issuer/Verifier or Verifier/Verifier pairs, or by the same Verifier.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,101 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+Wallet reacts (e.g., whether the wallet retrieves Credential Issuer metadata hosted by an
+attacker's server), and, therefore, learn which Wallet is installed. To avoid this, the Wallet
+SHOULD require user interaction before acting on the received Credential Offer.
+
+## Wallet Attestation
+
+The Wallet transmits and stores sensitive information about the End-User. To ensure that the
+Wallet can handle those appropriately (i.e., according to a certain trust framework or a
+regulation), the Credential Issuer SHOULD request an Attestation from the Wallet using
+methods, including but not limited to those defined in [@!I-D.ietf-oauth-attestation-based-client-auth].
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
regulation), the Credential Issuer should properly authenticate the Wallet and ensure it is a trusted entity. For more details, see (#trust-between-wallet-and-issuer).
```

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
Sakurann: Discussed during the WG call, agreed to keep it a should for VCI use-cases. (OIDC core says "MUST obtain user consent"..)

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
Sakurann: agreed to remove this sentence about the wallet as there are multiple scenarios possible.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,96 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+### Unique Values Encoded in the Credential
+
+Issuance/presentation or two presentation sessions by the same End-User can be linked on the basis of
+unique values encoded in the Credential (End-User claims, identifiers, Issuer signature, etc.) either by colluding Issuer/Verifier or Verifier/Verifier pairs, or by the same Verifier.
+
+To prevent these types of correlation, Credential Issuers and Wallets SHOULD use
+methods, including but not limited to the following ones:
+
+* Issue a batch of Credentials to enable the usage a unique Credential per presentation or per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
```
bc-pi, line 1552: just a typo missed word 

```suggestion
* Issue a batch of Credentials to enable the usage of a unique Credential per presentation or per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,96 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+### Credential Offer
+
+The Privacy Considerations in Section 11.2 of [@!RFC9101] apply to the `credential_offer` and
+`credential_offer_uri` parameters defined in (#credential-offer).
+
+### Authorization Request
+
+The Wallet SHOULD NOT include potentially sensitive information in the Authorization Request,
+for example, that including clear-text session information as a `state` parameter value, or encoding
```
bc-pi, line 1563: This sentence doesn't quite read right to me. Here's an attempt to fix while keeping what I think was the intent:

```suggestion
for example, by including clear-text session information as a `state` parameter value or encoding
```

bc-pi (approved):
Two minor nits on the latest text that I assume will be incorporated so approving.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,96 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+For example, if the National Cancer Institute only issued Credentials with cancer registry
+information, it is possible to deduce that the owner of the Wallet storing such Credential is a
+cancer patient.
+
+To mitigate this issue, a group of Credential Issuers may elect to use a common Credential Issuer
+Identifier. A group signature scheme may also be used, instead of an individual signature.
+
+## Custom URL Scheme
+
+There is a potential for leaking information about the Wallet to the third parties when the
```
jogu, line 1588:
```suggestion
There is a potential for leaking information about the Wallet to third parties when the
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,96 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+Wallet reacts to a Credential Offer that was sent to the custom URL scheme that the Wallet
```
jogu, line 1589:
```suggestion
Wallet reacts to a Credential Offer that was sent to a custom URL scheme that the Wallet
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,96 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+supports. An attacker may send Credential Offers to different custom URL schemes, see if the
+Wallet reacts (e.g., whether the wallet retrieves Credential Issuer metadata hosted by an
+attacker's server), and, therefore, learn which Wallet is installed. To avoid this, the Wallet
+SHOULD require user interaction (or establish trust in the Issuer) before fetching the `credential_offer_uri ` or acting on the received Credential Offer.
```
jogu, line 1593:
```suggestion
SHOULD require user interaction (or establish trust in the Issuer) before fetching any `credential_offer_uri ` or acting on the received Credential Offer.
```

jogu (approved):
Looks good to me! I agree with Brian's suggestions, and also suggested a few trivial grammar improvements.

c2bo:
The line breaks are somehow weird for this PR? Shouldn't for example

> For example, when a military organization or a drug rehabilitation center issues a vaccine
> credential, verifiers can deduce that the owner of the Wallet storing such Credential is a
> military member or may have a substance use disorder.

be in one line like this?

> For example, when a military organization or a drug rehabilitation center issues a vaccine credential, verifiers can deduce that the owner of the Wallet storing such Credential is a military member or may have a substance use disorder.

jogu:
> The line breaks are somehow weird for this PR?

The markdown has been wrapped, which is inconsistent with the rest of the spec - albeit that doesn't make any difference to the generated HTML.

selfissued:
> The markdown has been wrapped, which is inconsistent with the rest of the spec - albeit that doesn't make any difference to the generated HTML.

The practice of putting an entire paragraph into a single source line results in MANY unnecessary merge conflicts.  I support putting each sentence or long phrase into its own line - at least as we add new text.

That said, line breaks should be at the end of sentences and/or complete clauses so that odd line breaks don't make it hard to search for multiple-word terms such as "authorization server".

peppelinux (approved):
Approved considering Brian's suggestions taken into consideration

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,96 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+cancer patient.
+
+To mitigate this issue, a group of Credential Issuers may elect to use a common Credential Issuer
+Identifier. A group signature scheme may also be used, instead of an individual signature.
+
+## Custom URL Scheme
+
+There is a potential for leaking information about the Wallet to the third parties when the
+Wallet reacts to a Credential Offer that was sent to the custom URL scheme that the Wallet
+supports. An attacker may send Credential Offers to different custom URL schemes, see if the
```
c2bo, line 1590: Not sure about this one, but sending to custom URL schemes sounds weird to me
```suggestion
supports. An attacker may send Credential Offers using different custom URL schemes, see if the
```

c2bo (approved):
I agree with the suggestions of joseph & brian. Looks good to me apart from those minor changes

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,96 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+third parties, systems implementing this specification SHOULD be designed to minimize the amount
+of End-User data that is stored. All involved parties SHOULD store Verifiable Credentials
+containing privacy-sensitive data only for as long as needed, including in log files. Any logging of
+End-User data should be carefully considered as to whether it is necessary at all. The time logs are retained
+for should be minimized.
+
+After Issuance, Credential Issuers SHOULD NOT store the Issuer-signed Credentials if they
+contain privacy-sensitive data. Wallets SHOULD store Credentials only in encrypted form, and,
+wherever possible, use hardware-backed encryption. Credentials that are no longer valid to present should be deleted as soon as
+possible.
```
cobward, line 1540:
```suggestion
wherever possible, use hardware-backed encryption.
```

This might be undesirable. I imagine there may be use cases where presenting an expired credential may still be helpful, for example presenting an expired credential to the issuer as an authentication factor for provisioning a fresh credential or simply as a convenient way to share data about oneself (without the guarantee of the validity of the data).

IMO, it is a policy decision of the relying party whether or not they wish to accept expired credentials, and the credential holder should have ultimate control over their credentials. This suggestion would be equivalent to a physical ID card (safely) incinerating itself upon reaching the expiry date.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,96 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+### Unique Values Encoded in the Credential
+
+Issuance/presentation or two presentation sessions by the same End-User can be linked on the basis of
+unique values encoded in the Credential (End-User claims, identifiers, Issuer signature, etc.) either by colluding Issuer/Verifier or Verifier/Verifier pairs, or by the same Verifier.
+
+To prevent these types of correlation, Credential Issuers and Wallets SHOULD use
+methods, including but not limited to the following ones:
+
+* Issue a batch of Credentials to enable the usage a unique Credential per presentation or per Verifier using Batch Credential Endpoint defined in (#batch-credential-endpoint). This only helps with Verifier/Verifier unlinkability.
+* Use cryptographic schemes that can provide non-correlation.
```
cobward, line 1553:
```suggestion
* Use cryptographic schemes that can provide non-correlation.
* Issuers should discard credential identifiers, such as the signature or subject key, after issuance to prevent Issuer/Verifier collusion.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,96 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+### Authorization Request
+
+The Wallet SHOULD NOT include potentially sensitive information in the Authorization Request,
+for example, that including clear-text session information as a `state` parameter value, or encoding
+it in a `redirect_uri` parameter. A third party may observe such information through browser
+history, etc. and correlate the user's activity using it.
+
+## Credential Issuer Identifier
+
+The Credential Issuer Identifier alone may reveal information about the End-User.
```
cobward, line 1569: To whom would this information be revealed?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,96 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+In addition, when a Credential Issuer issues only one type of Credential, it might have privacy implications,
+because if the Wallet has a Credential issued by that Issuer, its type and claim names can be
+determined.
+
+For example, if the National Cancer Institute only issued Credentials with cancer registry
+information, it is possible to deduce that the owner of the Wallet storing such Credential is a
+cancer patient.
+
+To mitigate this issue, a group of Credential Issuers may elect to use a common Credential Issuer
+Identifier. A group signature scheme may also be used, instead of an individual signature.
```
cobward, line 1584: Is this about the `Credential Issuer Identifier`, a metadata field in OpenID4VCI, or about generally linking credentials to issuers using identifiers in the credential?

If it is the former, how does the `Credential Issuer Identifier` become shared with a verifier?

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: what's the definition of "credential identifiers"

this is needed for a better understanding of your proposed change

GDPR and national regulation prevents this, since the signed documents must be retained for years since are citizen's personal data handled by the issuers (for legacy infrastructurs id token are retained from 2 to 5 years). There might be other requirements about national regulation where the storage of a signed artifact is needed to resolve any disputes in the future

openid-4-verifiable-credential-issuance-1_0.md
jogu: > The most important rule to keep in mind is that anything must be under the user's control, even the deletion.

I don't think this is anything we should be including in the VCI spec; there are cases (e.g. fraud) where it would be completely valid for the wallet to purge a credential without the user agreeing to it.

openid-4-verifiable-credential-issuance-1_0.md
jogu: The wallet, any verifier, and anyone else that is able to access the credential.

openid-4-verifiable-credential-issuance-1_0.md
jogu: > If it is the former, how does the Credential Issuer Identifier become shared with a verifier?

It is the former. In an SD-JWT VC for example, the credential issuer identifier is stored in the `iss` claim in the credential send to the verifier.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: text ` Credentials that are no longer valid to present should be deleted as soon as possible.` has been deleted already, so I think your comment has been addressed, @cobward ?

jogu: @Sakurann That text is still there. (We did remove the mention of 'expired' after last week's WG call but kept the text, I think Jacob is the first person I'm aware of that has suggested removed that text entirely.)

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: I think this is what Jacob meant..?
```suggestion
* Use cryptographic schemes that can provide non-correlation.
* Discard values that can be used in Issuer/Verifier collusion, such as Issuer's signature or cryptographic key material to which an issued credential was bound to.
```

cobward: Yep exactly

openid-4-verifiable-credential-issuance-1_0.md
cobward: `mdoc` does not use `Credential Issuer Identifier`, but it does have a certificate that identifies the issuer. So I think this section needs to be more general, and talk about any field in the credential that could identify a specific issuer, and therefore disclose more information than intended.

openid-4-verifiable-credential-issuance-1_0.md
cobward: AFAIK the OID4VCI spec does not explicitly state that the Credential Issuer Identifier must be stored in a credential, this is what confused me as I thought only the issuer and wallet would be the only parties privy to this information. See my below comment.

openid-4-verifiable-credential-issuance-1_0.md
cobward: Thanks all for the clarification, and I'm sorry that I missed this context from the WG calls; they almost always clash with other immovable meetings for me. Are there minutes for these calls that I can use to keep up with the conversations?

Back on the topic:
> The word 'expired' was deliberately avoided for exactly the reasons you say

Ah right. For this reason, I think the fact that I read it this way could mean that the text is ambiguous and at the very least needs to clarify what `valid to present` actually means.

> I would expect ecosystems and/or issuers will probably provide a definition for 'valid to present'.

If this is what we mean then we should be specific, and say something to the effect of:
_Credentials that are **deemed** no longer valid to present **by the given credential ecosystem** should be deleted as soon as possible._

I still think it should be up to the user to decide whether or not their credential gets deleted, as they may have uses for the credential that the wallet developer is not aware of. I don't think it's necessarily true that credentials always exist in closed and unchanging ecosystems.

>  there are cases (e.g. fraud)

We're talking about validity here though, not fraud. Would you have the same problem with this modification of Giuseppe's sentence? _The most important rule to keep in mind is that anything must be under **a legitimate** user's control, even the deletion._

openid-4-verifiable-credential-issuance-1_0.md
cobward: > GDPR and national regulation prevents this, since the signed documents must be retained for years since are citizen's personal data handled by the issuers (for legacy infrastructurs id token are retained from 2 to 5 years). There might be other requirements about national regulation where the storage of a signed artifact is needed to resolve any disputes in the future

@peppelinux is this true for all kinds of credentials or only "identity" credentials? If this is only for identity credentials then I think my point is still valid.

Besides, if we believe a mechanism is good for privacy then I think it is worthy of being in a privacy considerations section, even if it is ruled out by regulations. Regulations change over time, so this consideration could become more relevant.

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: my UX is that nobody can put his hands in my pocket, open my wallet, and remove stuffs. As in the real world, as in the digital one :-)

jogu: >  Are there minutes for these calls that I can use to keep up with the conversations?

Yes, they get sent out to the mailing list - you can subscribe or view the archives here: https://lists.openid.net/mailman/listinfo/openid-specs-digital-credentials-protocols

> Credentials that are deemed no longer valid to present by the given credential ecosystem should be deleted as soon as possible.

I think this presupposes ecosystems etc which aren't currently ever mentioned in the VCI spec; these are privacy recommendations so should be kept general.

> We're talking about validity here though, not fraud.

I think they're related though; my presumption would be that a credential that is discovered to have been fraudulently obtained would likely become invalid to present. (Is it the word 'valid' that's causing confusion - if so any suggestions for alternative words are welcome!)

> The most important rule to keep in mind is that anything must be under a legitimate user's control, even the deletion.

I do appreciate the comparison removing a physical credential from a user's physical wallet but I am not convinced that a Wallet/Issuer should need to obtain the legitimate user's permission to delete a credential that subsequently turns out to have been fraudulently issued.

The discussion on the WG call did seek to draw a line between 'credential content has expired' (so a physical passport that has reached it's "valid until" date is potentially no longer valid for travel but might still be valid for identity proofing)  with (e.g. in the case of an SD-JWT) 'credential format cryptographic wrapping `exp` value' - the consensus on the call seemed to be towards the latter being something that was generally expected to be honoured, I'm interested to know if there is actually consensus on that.

openid-4-verifiable-credential-issuance-1_0.md
cobward: Suggested replacement text for this section:

> ## Credential Issuer identifiers
>
> Information in the credential identifying a particular Credential Issuer, such as an issuer identifier or certificate, may reveal information about the End-User.
> 
> For example, when a military organization or a drug rehabilitation center issues a vaccine credential, verifiers can deduce that the owner of the Wallet storing such a credential is a military member or may have a substance use disorder.
>   
> In addition, when a Credential Issuer issues only one type of credential, it might have privacy implications because if the Wallet has a credential issued by that Issuer, its type and claim names can be determined.
> 
> For example, if the National Cancer Institute only issued credentials with cancer registry information, it is possible to deduce that the owner of the Wallet storing such a credential is a cancer patient.
> 
> To mitigate these issues, a group of organizations may elect to share a single Credential Issuer, such that any credentials issued by this Issuer cannot be attributed to a particular organization through issuer identifiers alone. A group signature scheme may also be used instead of an individual signature.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,96 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
-TBD
+When [@!RFC9396] is used, the Privacy Considerations of that specification also apply.
+
+The privacy principles of [@ISO.29100] should be adhered to.
+
+## User Consent
+
+The Credential Issuer should obtain the End-User's consent before issuing Credential(s)
+to the Wallet. It should be made clear to the End-User what information is being included in the
+Credential(s) and for what purpose.
```
danielfett, line 1518: Why lower-case should here?

openid-4-verifiable-credential-issuance-1_0.md
yaromin: > 
> While validity and fraud are similar we should explicitly talk about each separately rather than conflating them into a single problem, since they should have different outcomes.
> 

I agree. For clarity it should be addressed in a separate line or paragraph. 

I also wanted to point out that in the same paragraph we have the first line that defines the intended behavior of the Issuer, the second that explicitly refers to the Wallet and the third one without an explicit subject. I would explictly mention a Wallet there for clarity.

>Credentials that are deemed no longer valid to present by all concerned parties (the holder and the verifiers >that the holder can reasonably be expected to interact with) SHOULD be deleted as soon as possible.

When it comes to the above sentence, it became a little verbose. In addition to that it should take into account UX which was raised by @peppelinux too. 
>SHOULD be deleted as soon as possible
The deletion does not have to happen automagically. I imagine that it could be indicated to a user in some way that such credentials are in their wallet, suggest deletion and let them do it.

openid-4-verifiable-credential-issuance-1_0.md
yaromin: > Suggested replacement text for this section:
> 
> > ## Credential Issuer identifiers
> > Information in the credential identifying a particular Credential Issuer, such as an issuer identifier or certificate, may reveal information about the End-User.
+1

> > To mitigate these issues, a group of organizations may elect to share a single Credential Issuer, such that any credentials issued by this Issuer cannot be attributed to a particular organization through issuer identifiers alone. A group signature scheme may also be used instead of an individual signature.

This is a more elegant approach but I would not rule out use of a technique described in the original sentence i.e. single issuer identifier shared across multiple Issuers.

jogu: @cobward 's suggestion seems sensible to me, though I think I would avoid using the term 'Credential Issuer identifiers' or 'issuer identifiers' as this seems highly likely to be confused with the defined term 'Credential Issuer Identifier'.

Perhaps something like "User-group specific Credential Issuers" for the title?

I think we should mention 'public keys' in the "such as" in the first paragraph too.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: It used to be a `MUST`, we had a passionate discussion during the WG call based on [a comment here](https://github.com/openid/OpenID4VCI/pull/244#discussion_r1473908607) and there were concerns with having a normative statements regarding consent.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann, line 1540:
```suggestion
After Issuance, Credential Issuers SHOULD NOT store the Issuer-signed Credentials if they
contain privacy-sensitive data. Wallets SHOULD store Credentials only in encrypted form, and,
wherever possible, use hardware-backed encryption. Wallet should delete Credentials that are
no longer valid to present as soon as possible.
```

Sakurann: > Credentials that are deemed no longer valid to present by all concerned parties (the holder and the verifiers that the holder can reasonably be expected to interact with) SHOULD be deleted as soon as possible.

I am not sure this is very actionable. it assumes there is always an agreement between verifier and wallet regarding whether the wallet can present an "expired" credential.

current sentence `no longer valid to be presented` implicitly includes making the determination what is `valid` which I think was the main point? we could also consider changing to `Wallet should delete Credentials that are no longer _verifiable_ to be presented as soon as possible.`

openid-4-verifiable-credential-issuance-1_0.md
Sakurann, line 1518:
```suggestion
The Credential Issuer SHOULD obtain the End-User's consent before issuing Credential(s)
to the Wallet. It SHOULD be made clear to the End-User what information is being included in the
Credential(s) and for what purpose.
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: > Yes, except it only applies to issuers. I don't think we want the wallet discarding those values :)

I added `(only applies to Issuers)`.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,96 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+because if the Wallet has a Credential issued by that Issuer, its type and claim names can be
+determined.
+
+For example, if the National Cancer Institute only issued Credentials with cancer registry
+information, it is possible to deduce that the owner of the Wallet storing such Credential is a
+cancer patient.
+
+To mitigate this issue, a group of Credential Issuers may elect to use a common Credential Issuer
+Identifier. A group signature scheme may also be used, instead of an individual signature.
+
```
Sakurann, line 1585:
```suggestion
## Identifier of the Credential Issuer

Information in the credential identifying a particular Credential Issuer, such as an issuer identifier or certificate, may reveal information about the End-User.

For example, when a military organization or a drug rehabilitation center issues a vaccine
credential, verifiers can deduce that the owner of the Wallet storing such Credential is a
military member or may have a substance use disorder.

In addition, when a Credential Issuer issues only one type of Credential, it might have privacy implications,
because if the Wallet has a Credential issued by that Issuer, its type and claim names can be
determined.

For example, if the National Cancer Institute only issued Credentials with cancer registry
information, it is possible to deduce that the owner of the Wallet storing such Credential is a
cancer patient.

To mitigate these issues, a group of organizations may elect to use a common Credential Issuer, such that any credentials issued by this Issuer cannot be attributed to a particular organization through issuer identifiers alone. A group signature scheme may also be used instead of an individual signature.

```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: @cobward thank you for the suggestion - I made the suggestion here: https://github.com/openid/OpenID4VCI/pull/244/files#r1480497886

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: should all `should`s in this text be upper case `SHOULD`?

openid-4-verifiable-credential-issuance-1_0.md
tplooker: One minor nit, is there any reason to caveat these normative requirements as only applying to credentials with "privacy-sensitve data" or could we generalise these statements a little more by removing that part

e.g

```suggestion
After Issuance, Credential Issuers SHOULD NOT store the Issuer-signed Credentials. Wallets SHOULD store Credentials only in encrypted form, and,
wherever possible, use hardware-backed encryption. Credentials that are no longer valid to present should be deleted as soon as
possible.
```

openid-4-verifiable-credential-issuance-1_0.md
tplooker:
```suggestion
* Use cryptographic schemes that can provide non-correlation.
* Credential Issuers specifically SHOULD discard values that can be used in collusion with a Verifier to track a user, such as the Issuer's signature or cryptographic key material to which an issued credential was bound to.
```

Alternative suggestion to the language above, has the same affect IMO.

openid-4-verifiable-credential-issuance-1_0.md
tplooker: @cobward typically a verifier requires knowing who the issuer is to some extend to establish trust, do you mean the concrete way the issuer is identified is different across credential formats?

tplooker (approved):
Minor nits but in general big supporter of this PR, thanks!

openid-4-verifiable-credential-issuance-1_0.md
cobward: > do you mean the concrete way the issuer is identified is different across credential formats?

Yes my point was that the Credential Issuer Identifier isn't a concrete way that is used across all formats.

Sakurann: that has been applied here; https://github.com/openid/OpenID4VCI/pull/244/files#r1480497886

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: based on the WG call, we changed the language to what is in the thread below: https://github.com/openid/OpenID4VCI/pull/244/files#r1480442717
moving away from valid/verifiable terms

cobward: > I am not sure this is very actionable. it assumes there is always an agreement between verifier and wallet regarding whether the wallet can present an "expired" credential.
> 
> current sentence no longer valid to be presented implicitly includes making the determination what is valid which I think was the main point?

Does that imply that it is up to the wallet alone to determine the validity of the credential?

> we could also consider changing to Wallet should delete Credentials that are no longer _verifiable_ to be presented as soon as possible.

Without a definition of "verifiable" I don't think it makes much difference. For example, if I can verify the signature, is that enough to be considered verifiable?

I think we need to take one of two paths:
1. Include or reference a definition of "valid to present"/"verifiable".
2. Explicitly state which party(s) can determine what "valid to present"/"verifiable" means.

openid-4-verifiable-credential-issuance-1_0.md
cobward: > I think we should mention 'public keys' in the "such as" in the first paragraph too.

+1

> This is a more elegant approach but I would not rule out use of a technique described in the original sentence i.e. single issuer identifier shared across multiple Issuers.

@yaromin I believe a shared issuer identifier alone does not solve the problem. As @jogu highlighted, a public key alone is enough to identify an issuer.

openid-4-verifiable-credential-issuance-1_0.md
cobward: I wonder if we should also mention the risks of this approach?

> Without sufficient guardrails, this shared issuer could be used by one organization to issue illegitimate credentials that another organization in the group is responsible for issuing.

cobward: And as @jogu suggested, could you add `public key` to the list of examples in the first sentence?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,104 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+`credential_offer_uri` parameters defined in (#credential-offer).
+
+### Authorization Request
+
+The Wallet SHOULD NOT include potentially sensitive information in the Authorization Request,
+for example, by including clear-text session information as a `state` parameter value or encoding
+it in a `redirect_uri` parameter. A third party may observe such information through browser
+history, etc. and correlate the user's activity using it.
+
+## Identifier of the Credential Issuer
```
jogu, line 1569: I think this might read better?

```suggestion
## Identity of the Credential Issuer
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,104 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+instead of an individual signature.
+
+When a common Credential Issuer is used, appropriate guardrails need to be in place to prevent
+one organization from issuing illegitimate credentials on behalf of other organizations.
+
+## Custom URL Scheme
+
+There is a potential for leaking information about the Wallet to third parties when the
+Wallet reacts to a Credential Offer that was sent to a custom URL scheme that the Wallet
+supports. An attacker may send Credential Offers using different custom URL schemes, see if the
```
jogu, line 1598: Reading this again (sorry for not spotting it the first time) I don't think this concern is limited to custom url schemes, I think similar concerns apply to claimed https urls?

```suggestion
## Processing of Credential Offers by Wallet prior to User Interaction

There is a potential for leaking information about the Wallet to third parties when the
Wallet reacts to a Credential Offer. An attacker may send Credential Offers using different
custom URL schemes or claimed https urls, see if the
```

openid-4-verifiable-credential-issuance-1_0.md
cobward: Ah sorry @Sakurann, not sure why I didn't see your comment when I responded. That text works for me, so will resolve this thread.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1507,7 +1507,104 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+Wallet reacts (e.g., whether the wallet retrieves Credential Issuer metadata hosted by an
+attacker's server), and, therefore, learn which Wallet is installed. To avoid this, the Wallet
+SHOULD require user interaction (or establish trust in the Issuer) before fetching any `credential_offer_uri ` or acting on the received Credential Offer.
```
Sakurann, line 1601:
```suggestion
attacker's server), and, therefore, learn which Wallet is installed. To avoid this, the Wallet MAY
require user interaction (or establish trust in the Issuer) before fetching any `credential_offer_uri `
or acting on the received Credential Offer.
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
## Identifying the Credential Issuer
```

Issue #201, opened 2024-01-04
Add a section of Wallet invocation


paulbastian:
OpenID4VP has a (small) section on how the Wallet shall be launched/invoked/started.
At least copying the section would not seem like a bad idea to me.

Especially section at https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-4.1-1
> The Credential Issuer sends Credential Offer using an HTTP GET request or an HTTP redirect to the Wallet's Credential Offer Endpoint defined in [Section 11.1](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#client_metadata).

does not match what many wallets have implemented today. Rather referencing to a Wallet Invocation Section may be a better solution.

peppelinux:
I agree, Wallet Invocation is something very particular in this field and it worth of a dedicated section (security considerations included I would add)

Sakurann:
I agree. redirect and GET are definitely not the only options. I tried clarify this in PR #246. please review and comment if it meets your expectations.

jogu:
Some or maybe all of this is probably resolved by the fix for #13, PR https://github.com/openid/OpenID4VCI/pull/380

PR #246, opened 2024-01-31
clarify wallet invocation when credential offer is sent

Sakurann:
resolves issue #201.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -302,11 +302,24 @@ For more details and concrete mitigations, see (#security-considerations-pre-aut
 
 # Credential Offer Endpoint {#credential-offer-endpoint}
 
-This endpoint is used by a Credential Issuer that is already interacting with an End-User who wishes to initiate a Credential issuance. It is used to pass available information relevant for the Credential issuance to ensure a convenient and secure process.
+Credential Offer Endpoint is used by a Credential Issuer that is already interacting with an End-User who wishes to initiate a Credential issuance. It is used to pass available information relevant for the Credential issuance to ensure a convenient and secure process as defined in (#credential-offer).
+
+Credential Offer Endpoint can take one of the following values:
+
+- A case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components.
+- Custom URL scheme. For example, `openid-credential-offer://` as defined in (#client-metadata-retrieval).
+- Domain-bound Universal Links/App link.
+- No specific value, user scanning a QR code with Credential Offer using a manually opened Wallet, instead of an arbitrary camera application on a user-device (neither of other options are used).
+
+How to obtain Credential Offer Endpoint is described in (#client-metadata-retrieval).
 
 ## Credential Offer {#credential-offer}
 
-The Credential Issuer sends Credential Offer using an HTTP GET request or an HTTP redirect to the Wallet's Credential Offer Endpoint defined in (#client-metadata).
+The Credential Issuer sends Credential Offer using one of the following mechanisms:
+
+- Render a link that the End-User can click resulting in an HTTP GET request to the Credential Offer Endpoint.
+- Render a QR code that can be scanned by the End-User using a Wallet.
+- An HTTP redirect to the Credential Offer Endpoint.
 
 The Credential Offer object, which is a JSON-encoded object with the Credential Offer parameters, can be sent by value or by reference.
 
@@ -314,8 +327,6 @@ The Credential Offer contains a single URI query parameter, either `credential_o
 
 * `credential_offer`: Object with the Credential Offer parameters. This MUST NOT be present when the `credential_offer_uri` parameter is present.
 * `credential_offer_uri`: String that is a URL using the `https` scheme referencing a resource containing a JSON object with the Credential Offer parameters. This MUST NOT be present when the `credential_offer` parameter is present.
-
-The Credential Issuer MAY render a QR code containing the Credential Offer that can be scanned by the End-User using a Wallet, or a link that the End-User can click.
 
 For security considerations, see (#credential-offer-security).
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -309,16 +309,16 @@ Credential Offer Endpoint can take one of the following values:
 - A case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components.
 - Custom URL scheme. For example, `openid-credential-offer://` as defined in (#client-metadata-retrieval).
 - Domain-bound Universal Links/App link.
-- No specific value, user scanning a QR code with Credential Offer using a manually opened Wallet, instead of an arbitrary camera application on a user-device (neither of other options are used).
+- No specific value. The issuer displays a QR code containing the Credential Offer, which the user can scan with a manually opened Wallet, or an arbitrary camera application on the user-device.
 
 How to obtain Credential Offer Endpoint is described in (#client-metadata-retrieval).
 
 ## Credential Offer {#credential-offer}
 
-The Credential Issuer sends Credential Offer using one of the following mechanisms:
+The Credential Issuer sends the Credential Offer using one of the following mechanisms:
 
 - Render a link that the End-User can click resulting in an HTTP GET request to the Credential Offer Endpoint.
-- Render a QR code that can be scanned by the End-User using a Wallet.
+- Render a QR code that can be scanned by the End-User using a Wallet or camera on the user's device.
 - An HTTP redirect to the Credential Offer Endpoint.
 
 The Credential Offer object, which is a JSON-encoded object with the Credential Offer parameters, can be sent by value or by reference.


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -302,11 +302,24 @@ For more details and concrete mitigations, see (#security-considerations-pre-aut
...
 
-This endpoint is used by a Credential Issuer that is already interacting with an End-User who wishes to initiate a Credential issuance. It is used to pass available information relevant for the Credential issuance to ensure a convenient and secure process.
+Credential Offer Endpoint is used by a Credential Issuer that is already interacting with an End-User who wishes to initiate a Credential issuance. It is used to pass available information relevant for the Credential issuance to ensure a convenient and secure process as defined in (#credential-offer).
+
+Credential Offer Endpoint can take one of the following values:
+
+- A case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components.
+- Custom URL scheme. For example, `openid-credential-offer://` as defined in (#client-metadata-retrieval).
+- Domain-bound Universal Links/App link.
+- No specific value, user scanning a QR code with Credential Offer using a manually opened Wallet, instead of an arbitrary camera application on a user-device (neither of other options are used).
```
David-Chadwick, line 312: No specific value. The issuer displays a QR code containing the Credential Offer, which the user can scan with a manually opened Wallet, or an arbitrary camera application on the user-device.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -302,11 +302,24 @@ For more details and concrete mitigations, see (#security-considerations-pre-aut
...
+
+How to obtain Credential Offer Endpoint is described in (#client-metadata-retrieval).
 
 ## Credential Offer {#credential-offer}
 
-The Credential Issuer sends Credential Offer using an HTTP GET request or an HTTP redirect to the Wallet's Credential Offer Endpoint defined in (#client-metadata).
+The Credential Issuer sends Credential Offer using one of the following mechanisms:
```
David-Chadwick, line 318: sends the Credential Offer

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -302,11 +302,24 @@ For more details and concrete mitigations, see (#security-considerations-pre-aut
...
+
+- Render a link that the End-User can click resulting in an HTTP GET request to the Credential Offer Endpoint.
+- Render a QR code that can be scanned by the End-User using a Wallet.
```
David-Chadwick, line 321: using a Wallet or camera on the user's device.

David-Chadwick (changes requested):

jogu (changes requested):
I think if we're going to make changes to this section then it needs a bigger rework.

The credential offer endpoint is a specific thing - we are defining the properties of the `credential_offer_endpoint` entry in the client meta data. In my opinion, the metadata for a client probably shouldn't ever be `openid-credential-offer://` as this value is only expected to be used when there isn't client meta data available for the wallet.

Section 4 should probably be renamed from 'Credential Offer Endpoint' to just 'Credential Offer', and a new 4.1 added called something like 'Credential Offer Endpoint' that explicitly defines `credential_offer_endpoint` metadata. Or the text that has been added to section 4 should instead be in section 11.1 where the `credential_offer_endpoint` is defined (and the IANA section should be updated to reference 11.1 instead of 4 as it does currently).

Finally, the 'no specific value' case should be better defined if we want interoperability. I think it is implying that the endpoint URL is treated as the empty string and hence the qr code would contain start with '?' but this isn't actually stated, i.e.

`?credential_offer=%7B%22credential_issuer%22:%22https://credential-issuer.example.com%22,%22credentials%22:%5B%22org.iso.18013.5.1.mDL%22%5D,%22grants%22:%7B%22urn:ietf:params:oauth:grant-type:pre-authorized_code%22:%7B%22pre-authorized_code%22:%22oaKazRN8I0IbtZ0C7JuMn5%22,%22tx_code%22:%7B%22input_mode%22:%22text%22,%22description%22:%22Please%20enter%20the%20serial%20number%20of%20your%20physical%20drivers%20license%22%7D%7D%7D%7D`

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: love the suggestion! thank you.

```suggestion
- No specific value. The issuer displays a QR code containing the Credential Offer, which the user can scan with a manually opened Wallet, or an arbitrary camera application on the user-device.
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
The Credential Issuer sends the Credential Offer using one of the following mechanisms:
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
- Render a QR code that can be scanned by the End-User using a Wallet or camera on the user's device.
```

Sakurann:
@jogu I am simply trying to resolve issue #201, because the issue rightly pointed out that HTTP GET and redirect is not what most implementers use - QR code is much more used, so it felt that the spec better reflect the reality.

you are making valid points, if you could open the issues, we can address them accordingly, but I am not convinced this PR cannot go in without larger changes you suggest.

Sakurann:
> I think it is implying that the endpoint URL is treated as the empty string and hence the qr code would contain start with '?' but this isn't actually stated, i.e.

sure, this can be clarified, but why use `?` if there is no URL...

jogu:
> @jogu I am simply trying to resolve issue #201, because the issue rightly pointed out that HTTP GET and redirect is not what most implementers use - QR code is much more used, so it felt that the spec better reflect the reality.
> 
> you are making valid points, if you could open the issues, we can address them accordingly, but I am not convinced this PR cannot go in without larger changes you suggest.

The minimum would be to move the newly added text  that is defining properties of credential_offer_endpoint to section 11.1 and to reference that section from section 4 but really all the points I make apart from the existing issue with the IANA considerations pointing to the wrong place are issues newly created by adding in the non specific url qr codes option.

jogu:
> > I think it is implying that the endpoint URL is treated as the empty string and hence the qr code would contain start with '?' but this isn't actually stated, i.e.
> 
> sure, this can be clarified, but why use `?` if there is no URL...

That was my attempt to interpet what the currently proposed text means.

I think the reason to use a '?' would be because we're still expecting everything in the offer parameters to be URL encoded as if it was present in the URL query, and included a leading '?' means that it is a valid relative URL.

It could work either way, and it would probably be good to know what implementers using this option actually do - but if we're adding it to the spec I believe we should be very clear about how it works otherwise implementations won't be interoperable.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -302,11 +302,24 @@ For more details and concrete mitigations, see (#security-considerations-pre-aut
...
+- No specific value. The issuer displays a QR code containing the Credential Offer, which the user can scan with a manually opened Wallet, or an arbitrary camera application on the user-device.
+
+How to obtain Credential Offer Endpoint is described in (#client-metadata-retrieval).
 
 ## Credential Offer {#credential-offer}
 
-The Credential Issuer sends Credential Offer using an HTTP GET request or an HTTP redirect to the Wallet's Credential Offer Endpoint defined in (#client-metadata).
+The Credential Issuer sends the Credential Offer using one of the following mechanisms:
+
+- Render a link that the End-User can click resulting in an HTTP GET request to the Credential Offer Endpoint.
```
jogu, line 320: I don't think there is any reason to restrict this to a link. (A button would be fine.)

I think something like "Sends the user agent to the Credential Offer Endpoint" (i.e. similar language to how OAuth 2 talks about the redirect url) is sufficient, which covers all possible cases include a HTTP redirect. (It'd be fine to include "For example, a link ... or a HTTP redirect...".)

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -302,11 +302,24 @@ For more details and concrete mitigations, see (#security-considerations-pre-aut
...
 
-This endpoint is used by a Credential Issuer that is already interacting with an End-User who wishes to initiate a Credential issuance. It is used to pass available information relevant for the Credential issuance to ensure a convenient and secure process.
+Credential Offer Endpoint is used by a Credential Issuer that is already interacting with an End-User who wishes to initiate a Credential issuance. It is used to pass available information relevant for the Credential issuance to ensure a convenient and secure process as defined in (#credential-offer).
+
+Credential Offer Endpoint can take one of the following values:
+
+- A case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components.
+- Custom URL scheme. For example, `openid-credential-offer://` as defined in (#client-metadata-retrieval).
+- Domain-bound Universal Links/App link.
+- No specific value. The issuer displays a QR code containing the Credential Offer, which the user can scan with a manually opened Wallet, or an arbitrary camera application on the user-device.
```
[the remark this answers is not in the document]
David-Chadwick: I know that our Android wallet app has a "Scan QR code" option inside it, so this definitely works. I "think" that our IoS phone allowed the camera to scan the QR code and then it automatically launched the wallet app, but I am no longer able to test this.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -302,11 +302,24 @@ For more details and concrete mitigations, see (#security-considerations-pre-aut
 
 # Credential Offer Endpoint {#credential-offer-endpoint}
 
-This endpoint is used by a Credential Issuer that is already interacting with an End-User who wishes to initiate a Credential issuance. It is used to pass available information relevant for the Credential issuance to ensure a convenient and secure process.
+Credential Offer Endpoint is used by a Credential Issuer that is already interacting with an End-User who wishes to initiate a Credential issuance. It is used to pass available information relevant for the Credential issuance to ensure a convenient and secure process as defined in (#credential-offer).
```
yaromin, line 305:
```suggestion
Wallet exposes the Credential Offer Endpoint, enabling the Credential Issuer to pass the Credential Offer to the Wallet, thereby initiating the Credential Issuance process. In this way, the Credential Issuer, which an End-User interacts with, uses Credential Offer to pass to the Wallet information necessary for convenient and secure Credential Issuance process defined in (#credential-offer) .
```

I would apply this change and in the line below indicated that there are several strategies Credential Offer Endpoint strategies like link, custom scheme, qr code  and so on.

Sakurann:
i need to update this PR - will do next week, sorry.

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
jogu: I think my conclusion (after asking on the WG calls) was that this 'No specific value' option isn't used by anyone so I think we should omit it (both here and in VP).

I think the only thing people might be doing is potentially ignoring the custom scheme when scanning a QR code in a wallet. Seems reasonable to mention that wallets might choose to do that.

openid-4-verifiable-credential-issuance-1_0.md
awoie: I agree that this has to changed. But opening a native app is technically not sending the user agent to the credential offer endpoint. It is also not following a redirect nor a HTTP GET request. 

What about:

```suggestion
- Render a link that the End-User can click resulting in following the link to the Credential Offer Endpoint.
```

jogu:
I believe this has been superseded by https://github.com/openid/OpenID4VCI/pull/380 - closing on that basis to keep the PR list to items we're actively working on, please feel free to reopen if I'm mistaken.

PR #253, opened 2024-02-06
Remove the duplicate requirement about not including query parameters to Credential Issuer Identifier

adeinega:
A Credential Issuer URL should not already include the query component per 11.2.2.

If an intent for the wallet was to not add any query parameters to the Credential Issuer Identifier when it fetches the metadata then this sentence should be changed to better reflect this.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1316,7 +1316,7 @@ Credential Issuers publishing metadata MUST make a JSON document available at th
 
 Communication with the Credential Issuer Metadata Endpoint MUST utilize TLS.
 
-To fetch the Credential Issuer Metadata, the requester MUST send an HTTP request using the GET method and the path formed following the steps above. The URL SHOULD NOT contain any query parameters. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and the HTTP Status Code 200.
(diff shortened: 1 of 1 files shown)


tplooker (approved):
I agree with @adeinega, I believe the normative requirement that this PR removes is needless duplication

Sakurann (commented):
if you are removing a normative statement, it is not an editorial change. please change the PR title accordingly

pmhsfelix:
Agree with the removal given both:
- the restriction that the issuer ID cannot have query string.
- the documented URI creation procedure for the metadata GET from the issuer ID.

pmhsfelix (approved):

paulbastian (changes requested):
The existing query statement is only refering to the the [Credential Issuer Identifier](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-credential-issuer-identifie).

I would propose to move this statement one paragraph earlier as:
"Credential Issuers publishing metadata MUST make a JSON document available at the path formed by concatenating the string /.well-known/openid-credential-issuer to the Credential Issuer Identifier. **The resulting URL SHOULD NOT contain any query parameters**"

selfissued (approved):

paulbastian (approved):
agreed afet wg call

Issue #263, opened 2024-02-11
Improve wording of explanation of credential_definition for "VC Signed as a JWT, Not Using JSON-LD"


hesusruiz:
In Section [A.1.1. ](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#appendix-A.1.1)[VC Signed as a JWT, Not Using JSON-LD](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-vc-signed-as-a-jwt-not-usin) the description for `credential_definition` states:

> Object containing the detailed description of the Credential type. It consists of **at least the following two parameters**:

But only the `type` parameter is REQUIRED, as `credentialSubject` is OPTIONAL.

For me the meaning is clear, but some readers may find it is somewhat inconsistent.

PR #269, opened 2024-02-16
ed: clarify "consists at least of" text

Sakurann:
fixes issue #263.

happy to remove "defined by this specification" if people think it's redundant.


tplooker (approved):
LGTM, in general I like this pattern of "defined by this specification", it leaves forward extensibility permitted but makes it clear what normative requirements this spec has for the effected data structure.

paulbastian (approved):

jogu (approved):

Issue #262, opened 2024-02-11
Authorization Details claims


c2bo:
The Appendix introduces specific claims for Authorization Details for mdoc and sd-jwt vc type credentials.
For both of them a new claim `claims` is introduced as

> claims: OPTIONAL. Object as defined in [Appendix A.3.2](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#server-metadata-sd-jwt-vc) excluding the display and value_type parameters. mandatory parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).

I was wondering if this should be extended to also allow expected values inside the claim. This would allow the Wallet to signal expected claims which might help for cases where there are several options of the same type of credential to be issued.

A good example would be a bank where one user might be the owner of different bank accounts within that one bank and would like to get a credential for a specific one. This way `authorization_details` could be leveraged to signal the expected one.

We could introduce a new optional claim `expected_value` like this:
```
[
    {
        "type": "openid_credential",
        "format": "vc+sd-jwt",
        "vct": "SD_JWT_VC_example_in_OpenID4VCI",
        "claims": {
            "given_name": {
                "expected_value": "SomeName"
            }
        }
    }
]
```

Sakurann:
> A good example would be a bank where one user might be the owner of different bank accounts within that one bank and would like to get a credential for a specific one.

Why would a wallet have information which bank account the user have in mind? It should be the issuer who provides user with an interface to specify which claims should go into the credential if there could be multiple options.

David-Chadwick:
`Why would a wallet have information which bank account the user have in mind?`
Ans. Because the user is in charge of the wallet and can type the information into it. The user surely knows which bank accounts she has.

c2bo:
Depends on how you start the flows. If you start from a website, then yes the user should be guided through some form of selection and choose for example the correct banking account. If you trigger the process from for example another App that has relevant information, it might make sense to provide that directly without user interaction.

The issuer would still check if the provided information is correct and matches its internal data but it would provide an alternative way of selection/interaction with the system.

jogu:
Perhaps a better example would be the wallet wanting to obtain, say, a university degree that matches the user it already has an identity credential for.

Sakurann:
I am much clearer on the use-case, thank you for clarifying. 

I think it was the following suggestion that confused me: `I was wondering if this should be extended to also allow expected values inside the claim. `

[...]

PR #276, opened 2024-02-20
Define claims display description and claims path query

### LINKED ISSUE #266: Credential format profiles: 'claims' definitions a̶r̶e̶ ̶b̶r̶o̶k̶e̶n̶ should be improved
Right now, the definitions (e.g., [this](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-credential-issuer-metadata-6)) for `claims`/`credentialSubject` in the credential format profiles are broken (as [pointed out](https://github.com/openid/OpenID4VCI/issues/57#issuecomment-1723533366) earlier).

At least the following problems exist:

 * There is no way to distinguish between a claim named `mandatory` and the syntax element `mandatory` (same for `display` etc.).
 * There is no way to describe, e.g., the `address` claim and the `address/region` sub-claim.
 * The implementer has to parse the structure in order to figure out whether a certain object defines properties of the claim or contains nested claims. While this is possible, it is hard to implement.
 * Handling of arrays of objects is undefined.

This is also related to [this issue](https://github.com/openid/OpenID4VCI/issues/74), both go back to not properly accounting for nested structures.

The problem exists in all three profiles defined in Appendix A.

### LINKED ISSUE #271: Variance in definitions for claim display metadata
In the three format profiles described in Appendix A, the definitions for the `claims`/`credentialSubject` contents do not match (ignoring inherent differences due to one of them describing the ISO mdoc format):

#jwt-vc-json says:

> ...MAY be an object that includes the following parameters defined by this specification (other parameters MAY also be included)

while the other two profiles omit the addition in parentheses. I presume that this is not intended.

### LINKED ISSUE #272: Incomplete definition of `mandatory` in SD-JWT VC claim metadata
The definition of the `mandatory` key in the claim display metadata in the format profiles in Appendix A seems to be incomplete for SD-JWT VC.

The first two read:

> `mandatory`: OPTIONAL. Boolean which, when set to `true`, indicates that the Credential Issuer will always include this claim in the issued Credential. If set to `false`, the claim is not included in the issued Credential if the wallet did not request the inclusion of the claim, and/or if the Credential Issuer chose to not include the claim. If the `mandatory` parameter is omitted, the default value is `false`.

The one for SD-JWT VC reads:

> `mandatory`: OPTIONAL. Boolean which when set to `true` indicates the claim MUST be present in the issued Credential. If the `mandatory` property is omitted its default should be assumed to be `false`.

Additional question: Why is there a normative keyword in the latter but not the previous two?

This is the PR introducing this change, but I don't have the time for archeology right now: https://github.com/openid/OpenID4VCI/pull/128

danielfett:
Fixes #272
Fixes #271 
Fixes #266

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1955,6 +1955,8 @@ This specification defines several extension points to accommodate the differenc
 
 This section defines Credential Format Profiles for a few of the commonly used Credential formats. Other specifications or deployments can define their own Credential Format Profiles.
 
+
+
 ## W3C Verifiable Credentials
 
 Sections 6.1 and 6.2 of [@VC_DATA] define how Verifiable Credentials MAY or MAY NOT use JSON-LD [@JSON-LD]. As acknowledged in Section 4.1 of [@VC_DATA], implementations can behave differently regarding processing of the `@context` property whether JSON-LD is used or not.
@@ -1987,13 +1989,7 @@ The following additional Credential Issuer metadata parameters are defined for t
 
 * `credential_definition`: REQUIRED. Object containing the detailed description of the Credential type. It consists of at least the following two parameters:
   * `type`: REQUIRED. Array designating the types a certain Credential type supports, according to [@VC_DATA], Section 4.3.
-  * `credentialSubject`: OPTIONAL. Object containing a list of name/value pairs, where each name identifies a claim offered in the Credential. The value can be another such object (nested data structures), or an array of such objects. To express the specifics about the claim, the most deeply nested value MAY be an object that includes the following parameters defined by this specification (other parameters MAY also be used):
-      * `mandatory`: OPTIONAL. Boolean which, when set to `true`, indicates that the Credential Issuer will always include this claim in the issued Credential. If set to `false`, the claim is not included in the issued Credential if the wallet did not request the inclusion of the claim, and/or if the Credential Issuer chose to not include the claim. If the `mandatory` parameter is omitted, the default value is `false`.
-      * `value_type`: OPTIONAL. String value determining the type of value of the claim. Valid values defined by this specification are `string`, `number`, and image media types such as `image/jpeg` as defined in IANA media type registry for images (https://www.iana.org/assignments/media-types/media-types.xhtml#image). Other values MAY also be used.
-      * `display`: OPTIONAL. Array of objects, where each object contains display properties of a certain claim in the Credential for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
-          * `name`: OPTIONAL. String value of a display name for the claim.
-          * `locale`: OPTIONAL. String value that identifies language of this object represented as language tag values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
-* `order`: OPTIONAL. Array of the claim name values that lists them in the order they should be displayed by the Wallet.
+  * `credentialSubject`: OPTIONAL. A claims description as defined in (#claims-description).
 
 The following is a non-normative example of an object containing the `credential_configurations_supported` parameter for Credential format `jwt_vc_json`:
 
@@ -2005,7 +2001,7 @@ The following additional claims are defined for authorization details of type `o
 
 * `credential_definition`: OPTIONAL. Object containing a detailed description of the Credential consisting of the following parameter:
   * `type`: OPTIONAL. Array as defined in (#server-metadata-jwt-vc-json). This claim contains the type values the Wallet requests authorization for at the Credential Issuer. It MUST be present if the claim `format` is present in the root of the authorization details object. It MUST not be present otherwise. 
-  * `credentialSubject`: OPTIONAL. Object as defined in (#server-metadata-sd-jwt-vc) excluding the `display` and `value_type` parameters. `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
+  * `credentialSubject`: OPTIONAL. A claims description as defined in (#claims-description) excluding the `display` and `value_type` parameters. The `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
 
 The following is a non-normative example of an authorization details object with Credential format `jwt_vc_json`:
 
@@ -2017,7 +2013,7 @@ The following additional parameters are defined for Credential Requests and this
 
 * `credential_definition`: REQUIRED when the `format` parameter is present in the Credential Request. It MUST NOT be used otherwise. It is an object containing the detailed description of the Credential type. It consists of at least the following parameters:
   * `type`: REQUIRED. Array as defined in (#server-metadata-jwt-vc-json). The credential issued by the Credential Issuer MUST contain at least the values listed in this claim.
-  * `credentialSubject`: OPTIONAL. Object as defined in (#authorization-jwt-vc-json).
+  * `credentialSubject`: OPTIONAL. A claims description as defined in (#claims-description).
 
 The following is a non-normative example of a Credential Request with Credential format `jwt_vc_json`:
 
@@ -2052,13 +2048,7 @@ The following additional Credential Issuer metadata parameters are defined for t
 * `credential_definition`: REQUIRED. Object containing the detailed description of the Credential type. It consists of at least the following three parameters:
   * `@context`: REQUIRED. Array as defined in [@VC_DATA], Section 4.1.
   * `type`: REQUIRED. Array designating the types a certain credential type supports, according to [@VC_DATA], Section 4.3.
-  * `credentialSubject`: OPTIONAL. Object containing a list of name/value pairs, where each name identifies a claim offered in the Credential. The value can be another such object (nested data structures), or an array of such objects. To express the specifics about the claim, the most deeply nested value MAY be an object that includes the following parameters defined by this specification (other parameters MAY also be included):
-      * `mandatory`: OPTIONAL. Boolean which, when set to `true`, indicates that the Credential Issuer will always include this claim in the issued Credential. If set to `false`, the claim is not included in the issued Credential if the wallet did not request the inclusion of the claim, and/or if the Credential Issuer chose to not include the claim. If the `mandatory` parameter is omitted, the default value is `false`.
-      * `value_type`: OPTIONAL. String value determining the type of value of the claim. Valid values defined by this specification are `string`, `number`, and image media types such as `image/jpeg` as defined in IANA media type registry for images (https://www.iana.org/assignments/media-types/media-types.xhtml#image). Other values MAY also be used.
-      * `display`: OPTIONAL. Array of objects, where each object contains display properties of a certain claim in the Credential for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
-          * `name`: OPTIONAL. String value of a display name for the claim.
-          * `locale`: OPTIONAL. String value that identifies language of this object represented as language tag values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
-* `order`: OPTIONAL. Array of the claim name values that lists them in the order they should be displayed by the Wallet.
+  * `credentialSubject`: OPTIONAL. A claims description as defined in (#claims-description).
 
 
 The following is a non-normative example of an object containing the `credential_configurations_supported` parameter for Credential format `ldp_vc`:
@@ -2072,7 +2062,7 @@ The following additional claims are defined for authorization details of type `o
 * `credential_definition`: OPTIONAL. Object containing the detailed description of the Credential consisting of the following parameter:
     * `@context`: OPTIONAL. Array as defined in (#server-metadata-ldp-vc). It MUST only be present if the `format` claim is present in the root of the authorization details object. It MUST not be present otherwise. 
     * `type`: OPTIONAL. Array as defined in (#server-metadata-ldp-vc).  This claim contains the type values the Wallet requests authorization for at the Credential Issuer. MUST only be present if the `@context` claim is present. 
-    * `credentialSubject`: OPTIONAL. Object as defined in (#server-metadata-sd-jwt-vc) excluding the `display` and `value_type` parameters. `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
+    * `credentialSubject`: OPTIONAL. A claims description as defined in (#claims-description) excluding the `display` and `value_type` parameters. The `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
 
 The following is a non-normative example of an authorization details object with Credential format `ldp_vc`:
 
@@ -2144,12 +2134,7 @@ Cryptographic algorithm names used in the `credential_signing_alg_values_support
 The following additional Credential Issuer metadata parameters are defined for this Credential format for use in the `credential_configurations_supported` parameter, in addition to those defined in (#credential-issuer-parameters).
 
 * `doctype`: REQUIRED. String identifying the Credential type, as defined in [@!ISO.18013-5].
-* `claims`: OPTIONAL. Object containing a list of name/value pairs, where the name is a certain `namespace` as defined in [@!ISO.18013-5] (or any profile of it), and the value is an object. This object also contains a list of name/value pairs, where the name is a claim name value that is defined in the respective namespace and is offered in the Credential. The value is an object detailing the specifics of the claim with the following non-exhaustive list of parameters that MAY be included:
-    * `mandatory`: OPTIONAL. Boolean which, when set to `true`, indicates that the Credential Issuer will always include this claim in the issued Credential. If set to `false`, the claim is not included in the issued Credential if the wallet did not request the inclusion of the claim, and/or if the Credential Issuer chose to not include the claim. If the `mandatory` parameter is omitted, the default value is `false`.
-    * `value_type`: OPTIONAL. String value determining the type of value of the claim. Valid values defined by this specification are `string`, `number`, and image media types such as `image/jpeg` as defined in IANA media type registry for images (https://www.iana.org/assignments/media-types/media-types.xhtml#image). Other values MAY also be used.
-    * `display`: OPTIONAL. Array of objects, where each object contains display properties of a certain claim in the Credential for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
-        * `name`: OPTIONAL. String value of a display name for the claim.
-        * `locale`: OPTIONAL. String value that identifies language of this object represented as language tag values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
+* `claims`: OPTIONAL. Object containing a list of name/value pairs, where the name is a certain `namespace` as defined in [@!ISO.18013-5] (or any profile of it), and the value is a claims description as defined in (#claims-description). The namespace applies to the claims in the claims description.
 * `order`: OPTIONAL. Array of namespaced claim name values that lists them in the order they should be displayed by the Wallet. The values MUST be two strings separated by a tilde ('~') character, where the first string is a namespace value and a second is a claim name value. For example, `org.iso.18013.5.1~given_name".
 
 The following is a non-normative example of an object containing the `credential_configurations_supported` parameter for Credential format `mso_mdoc`:
@@ -2161,7 +2146,7 @@ The following is a non-normative example of an object containing the `credential
 The following additional claims are defined for authorization details of type `openid_credential` and this Credential format.
 
 * `doctype`: OPTIONAL. String as defined in (#server-metadata-mso-mdoc). This claim contains the type value the Wallet requests authorization for at the Credential Issuer. It MUST only be present if the `format` claim is present. It MUST not be present otherwise. 
-* `claims`: OPTIONAL. Object as defined in (#server-metadata-sd-jwt-vc) excluding the `display` and `value_type` parameters. `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
+* `claims`: OPTIONAL. Object as defined in (#server-metadata-mso-mdoc) excluding the `display` and `value_type` parameters. The `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
 
 The following is a non-normative example of an authorization details object with Credential format `mso_mdoc`:
 
@@ -2198,13 +2183,7 @@ The following additional Credential Issuer metadata parameters are defined for t
 
 
 * `vct`: REQUIRED. String designating the type of a Credential, as defined in [@!I-D.ietf-oauth-sd-jwt-vc].
-* `claims`: OPTIONAL. Object containing a list of name/value pairs, where each name identifies a claim about the subject offered in the Credential. The value can be another such object (nested data structures), or an array of such objects. To express the specifics about the claim, the most deeply nested value MAY be an object that includes the following parameters defined by this specification:
-    * `mandatory`: OPTIONAL. Boolean which when set to `true` indicates the claim MUST be present in the issued Credential. If the `mandatory` property is omitted its default should be assumed to be `false`.
-    * `value_type`: OPTIONAL. String value determining the type of value of the claim. Valid values defined by this specification are `string`, `number`, and image media types such as `image/jpeg` as defined in IANA media type registry for images (https://www.iana.org/assignments/media-types/media-types.xhtml#image). Other values MAY also be used.
-    * `display`: OPTIONAL. Array of objects, where each object contains display properties of a certain claim in the Credential for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
-        * `name`: OPTIONAL. String value of a display name for the claim.
-        * `locale`: OPTIONAL. String value that identifies language of this object represented as language tag values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
-* `order`: OPTIONAL. An array of the claim name values that lists them in the order they should be displayed by the Wallet.
+* `claims`: OPTIONAL. A claims description as defined in (#claims-description).
 
 The following is a non-normative example of an object comprising the `credential_configurations_supported` parameter for Credential format `vc+sd-jwt`.
 
@@ -2215,7 +2194,7 @@ The following is a non-normative example of an object comprising the `credential
 The following additional claims are defined for authorization details of type `openid_credential` and this Credential format.
 
 * `vct`: REQUIRED. String as defined in (#server-metadata-sd-jwt-vc). This claim contains the type values the Wallet requests authorization for at the Credential Issuer. It MUST only be present if the `format` claim is present. It MUST not be present otherwise.
-* `claims`: OPTIONAL. Object as defined in (#server-metadata-sd-jwt-vc) excluding the `display` and `value_type` parameters. `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
+* `claims`: OPTIONAL. A claims description as defined in (#claims-description). The `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
 
 The following is a non-normative example of an authorization details object with Credential format `vc+sd-jwt`.
 
@@ -2226,7 +2205,7 @@ The following is a non-normative example of an authorization details object with
 The following additional parameters are defined for Credential Requests and this Credential format.
 
 * `vct`: REQUIRED when the `format` parameter is present in the Credential Request. It MUST NOT be used otherwise. It is a string as defined in (#server-metadata-sd-jwt-vc). This claim contains the type value of the Credential that the Wallet requests the Credential Issuer to issue.
-* `claims`: OPTIONAL. An object as defined in (#server-metadata-sd-jwt-vc).
+* `claims`: OPTIONAL. A claims description as defined in (#claims-description).
 
 The following is a non-normative example of a Credential Request with Credential format `vc+sd-jwt`.
 
@@ -2235,6 +2214,116 @@ The following is a non-normative example of a Credential Request with Credential
 ### Credential Response {#credential-response-jwt-vc-json}
 
 The value of the `credential` claim in the Credential Response MUST be a string that is an SD-JWT VC. Credentials of this format are already suitable for transfer and, therefore, they need not and MUST NOT be re-encoded.
+
+# Claims Description {#claims-description}
+
+A claims description object is an object used to describe how a certain
+claim in the Credential should be displayed to the End-User. It is used in the
+`display` parameter of the `claims` or `credentialSubject` parameters in the
+Credential Issuer metadata defined in (#format-profiles).
+
+For a claims description object, the following keys defined by this
+specification MAY be used to describe the claim or claims (other keys MAY also
+be used):
+
+  * `path`: REQUIRED. Array. Claim path query as defined in (#claim-path-query)
+    to identify the claim(s) in the Credential.
+  * `mandatory`: OPTIONAL. Boolean which, when set to `true`, indicates that the
+    Credential Issuer will always include this claim in the issued Credential.
+    If set to `false`, the claim is not included in the issued Credential if the
+    wallet did not request the inclusion of the claim, and/or if the Credential
+    Issuer chose to not include the claim. If the `mandatory` parameter is
+    omitted, the default value is `false`.
+  * `value_type`: OPTIONAL. String value determining the type of value of the
+    claim. Valid values defined by this specification are `string`, `number`,
+    and image media types such as `image/jpeg` as defined in IANA media type
+    registry for images
+    (https://www.iana.org/assignments/media-types/media-types.xhtml#image).
+    Other values MAY also be used.
+  * `display`: OPTIONAL. Array of objects, where each object contains display
+    properties of a certain claim in the Credential for a certain language.
+    Below is a non-exhaustive list of valid parameters that MAY be included:
+      * `name`: OPTIONAL. String value of a display name for the claim.
+      * `locale`: OPTIONAL. String value that identifies language of this object
+        represented as language tag values defined in BCP47 [@!RFC5646]. There
+        MUST be only one object for each language identifier.
+
+# Claims Path Query {#claims-path-query}
+
+The metadata specifications above make use of claims path queries to select a
+particular claim in the credential or a set of such claims. This appendix
+defines the syntax and semantics of claims path queries.
+
+A claims path query MUST be a non-empty array of strings, `null` values, or
+non-negative integers. 
+
+A string indicates that the respective key is to be selected, a `null` value
+indicates that all elements of the currently selected array(s) are to be
+selected, and a non-negative integer indicates that the respective index in an
+array is to be selected.
+
+## Example
+
+The following shows a non-normative, simplified example of a credential:
+
+```json
+{
+  "name": "Arthur Dent",
+  "address": {
+    "street_address": "42 Market Street",
+    "city": "Milliways",
+    "postal_code": "12345"
+  },
+  "degrees": [
+    {
+      "type": "Bachelor of Science",
+      "university": "University of Betelgeuse"
+    },
+    {
+      "type": "Master of Science",
+      "university": "University of Betelgeuse"
+    }
+  ],
+  "nationalities": ["British", "Betelgeusian"]
+}
+```
+
+The following shows examples of claims path queries and the respective selected
+claims:
+
+- `["name"]`: The claim `name` with the value `Arthur Dent` is selected.
+- `["address"]`: The claim `address` with its sub-claims as the value is selected.
+- `["address", "street_address"]`: The claim `street_address` with the value
+  `42 Market Street` is selected.
+- `["degrees", null, "type"]`: All `type` claims in the `degrees` array are
+  selected.
+- `["nationalities", 1]`: The second nationality is selected.
+
+## Processing
+
+In detail, the array is processed from left to right as follows:
+
+ 1. Select the root element of the credential, i.e., the top-level JSON object.
+ 2. Process the query components from left to right:
+    1. If the query component is a string, select the element in the respective
+       key in the currently selected element(s). If any of the currently
+       selected element(s) is not an object, abort processing and return an
+       error. If the key does not exist in any element currently selected,
+       remove that element from the selection.
+    2. If the query component is `null`, select all elements of the currently
+       selected array(s). If any of the currently selected element(s) is not an
+       array, abort processing and return an error.
+    3. If the query component is a non-negative integer, select the element at
+       the respective index in the currently selected array(s). If any of the
+       currently selected element(s) is not an array, abort processing and
+       return an error. If the index does not exist in a selected array, remove
+       that array from the selection.
+  3. If the set of elements currently selected is empty, abort processing and
+     return an error.
+
+The result of the processing is the set of elements to which the respective
+claim metadata applies.
+
 
 # IANA Considerations
 


examples/credential_metadata_sd_jwt_vc.json
```diff
@@ -43,19 +45,17 @@
...
-                    "country": {}
-                },
-                "birthdate": {},
-                "is_over_18": {},
-                "is_over_21": {},
-                "is_over_65": {}
-            }
+                {"path": ["email"]},
+                {"path": ["phone_number"]},
+                {"path": ["address", "street_address"]},
```
mickrau, line 50: We would like to give display properties to an object of claims. The wallet could use this property when collapsing the object's claims in the UI. Is this possible and would the following be the right way?
If so, this should be part of an example.

```json
                  {
                    "path": ["address"],
                    "display": [
                        {
                            "name": "Address",
                            "locale": "en-US"
                        },
                        {
                            "name": "Adresse",
                            "locale": "de-DE"
                        }
                    ],
                    "value_type": "object"
                },
                {"path": ["address", "street_address"]},
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1987,13 +1989,7 @@ The following additional Credential Issuer metadata parameters are defined for t
...
-      * `mandatory`: OPTIONAL. Boolean which, when set to `true`, indicates that the Credential Issuer will always include this claim in the issued Credential. If set to `false`, the claim is not included in the issued Credential if the wallet did not request the inclusion of the claim, and/or if the Credential Issuer chose to not include the claim. If the `mandatory` parameter is omitted, the default value is `false`.
-      * `value_type`: OPTIONAL. String value determining the type of value of the claim. Valid values defined by this specification are `string`, `number`, and image media types such as `image/jpeg` as defined in IANA media type registry for images (https://www.iana.org/assignments/media-types/media-types.xhtml#image). Other values MAY also be used.
```
mickrau, line 1992: The types `array` and `object` should also be mentioned explicitly and used in examples.
How can the type of array elements be determined?
```json
{"path": ["nationalities"], "value_type": "array"}
{"path": ["nationalities", null], "value_type": "string"}
```

mickrau (changes requested):
After review in our team: In principle, we like the proposal and support it.
We left two comments regarding claims object display property and value_type.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2161,7 +2146,7 @@ The following is a non-normative example of an object containing the `credential
 The following additional claims are defined for authorization details of type `openid_credential` and this Credential format.
 
 * `doctype`: OPTIONAL. String as defined in (#server-metadata-mso-mdoc). This claim contains the type value the Wallet requests authorization for at the Credential Issuer. It MUST only be present if the `format` claim is present. It MUST not be present otherwise. 
-* `claims`: OPTIONAL. Object as defined in (#server-metadata-sd-jwt-vc) excluding the `display` and `value_type` parameters. `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
+* `claims`: OPTIONAL. Object as defined in (#server-metadata-mso-mdoc), excluding the `display` and `value_type` parameters. The `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
```
paulbastian, line 2149:
```suggestion
* `claims`: OPTIONAL. Object containing a list of name/value pairs, where the name is a `namespace` as defined in [@!ISO.18013-5] (or any profile of it), and the value is an array of claims path queries as defined in (#claims-path-query). The  parameter is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued containing exactly those claim(s). The namespace applies to the claims described in the array of claims description objects.
```

paulbastian (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2236,6 +2215,118 @@ The following is a non-normative example of a Credential Request with Credential
 
 The value of the `credential` claim in the Credential Response MUST be a string that is an SD-JWT VC. Credentials of this format are already suitable for transfer and, therefore, they need not and MUST NOT be re-encoded.
 
+# Claims Description {#claims-description}
+
```
pmhsfelix, line 2219: Currently the _claims description object_ described in this section is being used for both:
- the metadata (information produced by the issuer).
- `authorization_details` entries (RAR) and credential request (information produced by the wallet), where the `value_type` and `display` keys cannot be used.

Instead of having a single object model, I believe it would be preferable to have two distinct object model sections, both sharing the path definition. IMO, the advantages are:
- We can grow each object model independently. For instance, in the future we can add policy information to the metadata model, without needing to explicitly forbid those new elements on RAR and credential requests.
- Have better descriptions for each key. For instance, even if the `mandatory` key appears on both, we could have distinct fine-tuned descriptions for each case. Currently the `mandatory` description text only applies to its use on the metadata and the meaning of `mandatory` on RAR and credential requests needs to be inferred.

paulbastian: Hi Pedro, this is exactly what may suggestions do, please check

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2005,7 +2001,7 @@ The following additional claims are defined for authorization details of type `o
 
 * `credential_definition`: OPTIONAL. Object containing a detailed description of the Credential consisting of the following parameter:
   * `type`: OPTIONAL. Array as defined in (#server-metadata-jwt-vc-json). This claim contains the type values the Wallet requests authorization for at the Credential Issuer. It MUST be present if the claim `format` is present in the root of the authorization details object. It MUST not be present otherwise. 
-  * `credentialSubject`: OPTIONAL. Object as defined in (#server-metadata-sd-jwt-vc) excluding the `display` and `value_type` parameters. `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
+  * `credentialSubject`: OPTIONAL. An array of claims description objects as defined in (#claims-description), excluding the `display` and `value_type` parameters. The `mandatory` parameter here is used by the Wallet to indicate to the Issuer that it only accepts Credential(s) issued with those claim(s).
```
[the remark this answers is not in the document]
bc-pi: I don't think that referencing this way works very well given how the #claims-description and #claims-path-query sections are written currently.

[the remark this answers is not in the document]
paulbastian: Could we start with the #claims-path-query section defining only `path` and then follow up with the #claims-description building on top?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1987,13 +1989,7 @@ The following additional Credential Issuer metadata parameters are defined for t
...
-      * `value_type`: OPTIONAL. String value determining the type of value of the claim. Valid values defined by this specification are `string`, `number`, and image media types such as `image/jpeg` as defined in IANA media type registry for images (https://www.iana.org/assignments/media-types/media-types.xhtml#image). Other values MAY also be used.
-      * `display`: OPTIONAL. Array of objects, where each object contains display properties of a certain claim in the Credential for a certain language. Below is a non-exhaustive list of valid parameters that MAY be included:
-          * `name`: OPTIONAL. String value of a display name for the claim.
-          * `locale`: OPTIONAL. String value that identifies language of this object represented as language tag values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
-* `order`: OPTIONAL. Array of the claim name values that lists them in the order they should be displayed by the Wallet.
+  * `credentialSubject`: OPTIONAL. An array of claims description objects as defined in (#claims-description).
```
pmhsfelix, line 1992: The use of an array of claim description objects now allows for multiple contradictory descriptions of the same claim, which I believe was not possible before.
E.g.
```
[
   {"path": ["nationalities"], "mandatory": true},
   {"path": ["nationalities"], "mandatory": false}
] 
```
We should clarify how a consumer should react to those cases:
- Reject the array (e.g. reject the authorization request is this happens on inside `authorization_details`)?
- Only consider the first one entry?
- Merge the properties of both entries (which may only make sense for some cases)?
With arrays, these collisions may be a bit more subtle.
E.g. How should this be handled on an authorization request?
```
[
   {"path": ["nationalities", 0], "mandatory": true},
   {"path": ["nationalities", null], "mandatory": false}
] 
```
Does this mean the `nationalities` array can have one or more elements.

Finally, we could have entries suggesting different structure (and not only different properties), such as
```
[
    {"path": ["nationalities", 0], "display": {...}},
    {"path": ["nationalities", "country"], "display": {...}}
]
```
The first entry indicates that `nationalities` is an array. The second entry indicates `nationalities` is an object.

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
bc-pi: Probably, yeah. My comment here was just to say that this change (and others like it) doesn't work without changes to the referenced sections.

examples/authorization_details_ldp_vc.json
```diff
@@ -3,11 +3,11 @@
         "type": "openid_credential",
         "credential_configuration_id": "UniversityDegree_LDP_VC",
         "credential_definition": {
-            "credentialSubject": {
-                "given_name": {},
-                "family_name": {},
-                "degree": {}
-            }
+            "credentialSubject": [
```
peppelinux, line 6: `credentialSubject` -> `credential_subject`

If the current evolution of this project introduces breaking changes, this presents an opportune moment to standardize naming conventions across the board. This could mean adopting snake_case.

peppelinux (changes requested):

examples/authorization_details_ldp_vc.json
danielfett: True. I had the same thought. Although I'm inclined to go for `claims` instead.

Sakurann (changes requested):
I put a comment in issue #266, but after today's WG call where a concern that this is a very big breaking change was raised, we discussed this internally with Microsoft engineers. We do not currently have use-cases with the claims that have nested structure and array-based solution is too big of a breaking change since we just finished implementing OpenID4VCI based on the text that went to ID-1 voting. So our preference is actually defining the new property `nested` which may contain descriptions about nested objects (original approach 2).

As a compromise for the existing implementations, my suggestion would be fixing the text that becomes ID-1 with `nested` approach. and than after ID-1, incorporating the array approach defined in this PR.

examples/authorization_details_ldp_vc.json
Sakurann: Microsoft just finished implementing with 'credentialSubject'. Given this is more of an issue of preference, would appreciate if we could keep this camel case.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2236,6 +2215,118 @@ The following is a non-normative example of a Credential Request with Credential
...
+# Claims Path Query {#claims-path-query}
+
+The metadata specifications above make use of claims path queries to select a
+particular claim in the credential or a set of such claims. This appendix
+defines the syntax and semantics of claims path queries.
+
+A claims path query MUST be a non-empty array of strings, `null` values, or
+non-negative integers. 
+
+A string indicates that the respective key is to be selected, a `null` value
```
peppelinux, line 2264: `null` -> `*`

would be more understandable ... but, before delving into these details, please consider the gap we have in IETF regarding the JSONPath (or similar approaches) and the requirement to get this standardized before propagating it in the openid specifications, in an unconventional way or in such a way that each specification, for specific needs, produces many narrow implementations with obvious implementation costs for the openid/ietf ecosystem.

As we know, JSONPath outlines a syntax for defining paths within a JSON document to find elements, much like XPath does for XML. It allows for filtering elements based on their attributes, navigating through arrays, and performing operations like slicing arrays.

peppelinux (changes requested):

examples/authorization_details_ldp_vc.json
TakahikoKawasaki: It appears at times that, on behalf of your employer, you may be hindering the community from considering better specifications. This is a significant concern for the community.

openid-4-verifiable-credential-issuance-1_0.md
danielfett: But `"*"` would be a string value which would require special rules for claims called `*` (admittedly, those shouldn't be too common).

JSONPath is a **huge** dependency. Suddenly, you have regular expressions and other stuff in your credential issuance protocol. Also some security issues, when you implement it incorrectly. And it brings so many things we don't need and would have to define-away...

examples/authorization_details_ldp_vc.json
danielfett: They would need to change quite a lot around the processing anyways, I can't imagine how a single renamed field would add significant cost to that.

openid-4-verifiable-credential-issuance-1_0.md
danielfett: The first and second example show points we indeed need to address. The third one should be taken care of by the current processing rules already.

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: I don't consider myself a jsonpath lover, I'm wondering if there's the time to define a basic, common and general purpose approach for dealing with all these json handling. I agree to not use JsonPath. I think that we have to define an alternative way, not just tied in openid4vci, something to be reused, and expandend where necessarly.

Considering my concerns about the use of null in claims path queries being potentially misleading and carrying a negative connotation, I'm wondering on an alternative approach with a more intuitive and self-explanatory keyword that aligns with the intended functionality. 

we have "special" string like "*" or "".

Or using objects like `{"select": "all"}` that unfortunately carries more nesting and other concerns.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1987,13 +1989,7 @@ The following additional Credential Issuer metadata parameters are defined for t
 
 * `credential_definition`: REQUIRED. Object containing the detailed description of the Credential type. It consists of at least the following two parameters:
```
cryptphil, line 1990: Isn't the sentence "It consists of at least the following two parameters" misleading because `credentialSubject` is marked as optional, so a valid `credential_definition` may also only contain one parameter?

I would suggest something like "It consists of at least the type parameter and an optional credentialSubject".

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2236,6 +2215,118 @@ The following is a non-normative example of a Credential Request with Credential
...
+array is used by the Wallet to determine the order in which the claims are
+displayed to the End-User, unless another mechanism is defined by the profile.
+
+# Claims Path Query {#claims-path-query}
+
+The metadata specifications above make use of claims path queries to select a
+particular claim in the credential or a set of such claims. This appendix
+defines the syntax and semantics of claims path queries.
+
+A claims path query MUST be a non-empty array of strings, `null` values, or
```
awoie, line 2261: It would be nice to support negative integers as well since those are used by CWT/COSE claims.

awoie (changes requested):
Need a way to define negative integers for COSE.

openid-4-verifiable-credential-issuance-1_0.md
danielfett: For addressing claims identified by numbers in CWT/COSE, I suggest to use numbers in strings to distinguish them from array elements. WDYT?

examples/authorization_details_ldp_vc.json
vdzhuvinov: Having disparate casings in a standard is frankly infuriating. Don't do that :)

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: We don't yet have any Credential Format Profiles where claims identified by numbers in CWT/COSE.  But it might be good to have an approach for them anyway.  Using numbers in strings to distinguish them from array elements would work but then there's a potential (though unlikely) for ambiguity of a claim identified by the number vs. identified by the number as a string. Maybe the rules here could cover that rare case by just saying to select for both in the case that the credential format allows for it. 

Or use an object to represent a number claim name/key something like `{"claim-identified-by-number": -257}`

bc-pi: Another approach would be use numbers similar to strings as representing the literal key value to select for. And then a to use single element or empty arrays to convey selecting on array element(s). So something like `[1]` would be use where just `1` is being used now. And an empty `[]` array (or some variation) could be used in place of how `null` is being used now.

Something like this, for example: 

 `["degrees", [], "type"]`: All `type` claims in the `degrees` array are selected.
 `["nationalities", [1]]`: The second nationality is selected.

awoie: Another approach would be to use a different property instead of `path`. Perhaps the simplest option would be to use something like `path_cbor` that could then contain negative integer values?

openid-4-verifiable-credential-issuance-1_0.md
c2bo: I was thinking empty string as well instead of null (i think it's reasonable to not allow empty string names) but some libraries will probably also do weird stuff with an empty string.

TakahikoKawasaki:
>>As a compromise for the existing implementations, my suggestion would be fixing the text that becomes ID-1 with nested approach. and than after ID-1, incorporating the array approach defined in this PR.
>>
>That would mean two breaking changes.

Is it suggested to incorporate a temporary solution (the `nested` approach) into ID1, even though it may be replaced in the future with another solution (the "array" approach)? Is this a proposal to improve the specification, or a suggestion to ease the workload of developers working on ID1 in a specific company? If the answer is the latter, it is a significant concern for the community.

Sakurann:
It's a suggestion to improve interoperability among the companies that have already implemented the current specification text, of which there are quite a few

bc-pi (commented):
There's a lot of activity on this right now but I'm anyway leaving a review/comment to express (continued) support for the general concept and direction the PR.

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: I like this proposal, in particular using `[]` instead of `null` as a wildcard

I don't know it there might be issues with some implementations, as the ones mentioned for `""`. I can say that in my implementations I can handle void instanced of arrays or strings without any problem.

c2bo: Using an Array would also allow you to
- select everything via `[]` 
- select a specific entry `[1]`
- select several sepcific entries (which other approaches can't easyil do `[0,1]`

That sounds like the cleanest option to me so far imho.

Sakurann:
I think we should have the discussion about what we want to prioritize: having a cleanest solution possible, or having a solution that builds up on what has already been implemented.

awoie (commented):
Generally supportive of the direction of this draft.

tlodderstedt:
I think the PR would bring a significant improvement to the VCI spec. However, I would like to point that it is a significant breaking change. Merging this PR now will make existing implementations incompatible and require us to restart the ID1 review process. 

I would like to understand what existing VCI implementers think and whether they support the change being merged now. 

The alternative would be to fix the most important issues brought up in https://github.com/openid/OpenID4VCI/issues/266 by allowing the key words `mandatory`, `value_type`, and `display` to be differentiated from ordinary claim names by adding a prefix `_`, for example.  The text could also be changed to allow for the key words, at least `display`, in objects. 

That might not be as nice and clean as the change proposed in this PR, but it could be specified in a backward compatible manner and would allow us to continue the ID process.

paulbastian:
How about not replacing the `claims` object but instead having a different parameter name for the new object. That would allow for better transition as Issuers could support both old and new format.
We could leave a statement that old format will be deprecate with ID-2?

selfissued (changes requested):
It doesn't seem that the proposed solution solves the whole problem.  I'd suggest not trying to apply it before Implementer's Draft one is approved.

After that, we can look at a wholistic solution.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2236,6 +2215,118 @@ The following is a non-normative example of a Credential Request with Credential
...
+
+  * `path`: REQUIRED. Array. Claim path query as defined in (#claims-path-query)
+    to identify the claim(s) in the Credential.
+  * `mandatory`: OPTIONAL. Boolean which, when set to `true`, indicates that the
+    Credential Issuer will always include this claim in the issued Credential.
+    If set to `false`, the claim is not included in the issued Credential if the
+    wallet did not request the inclusion of the claim, and/or if the Credential
+    Issuer chose to not include the claim. If the `mandatory` parameter is
+    omitted, the default value is `false`.
+  * `value_type`: OPTIONAL. String value determining the type of value of the
```
sloops77, line 2237: I feel that `value_type` needs more description. 
Is the default `string`?
The following value_types that are needed by our wallets are
1. iso8601_date
2. iso8601_time
3. iso8601_datetime
4. iso8601_duration
5. uri
6. image/* where the credential has a generic image type like the HTTP accepts header

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2236,6 +2215,118 @@ The following is a non-normative example of a Credential Request with Credential
...
+    claim. Valid values defined by this specification are `string`, `number`,
+    and image media types such as `image/jpeg` as defined in IANA media type
+    registry for images
+    (https://www.iana.org/assignments/media-types/media-types.xhtml#image).
+    Other values MAY also be used.
+  * `display`: OPTIONAL. Array of objects, where each object contains display
+    properties of a certain claim in the Credential for a certain language.
+    Below is a non-exhaustive list of valid parameters that MAY be included:
+     * `name`: OPTIONAL. String value of a display name for the claim.
+     * `locale`: OPTIONAL. String value that identifies language of this object
```
sloops77, line 2247: Open badge v3 credentials are used extensively throughout the education sector in the US. OBV3 supports an "identity" object that contains variable data. For example
```json
{
  "identityHash": "jane@example.com",
  "identityType": "email"
}
```
but the same structure can also hold a name
```json
{
  "identityHash": "Jane Foobar",
  "identityType": "name"
}
```

Something like the below would support it, but it requires conditional logic
```json
[{
  "path": ["identifier", "identityHash"],
  "display": [{
    "name": "Name",
    "path": ["identifier", "identityType"],
    "condition": "= 'name'"
  }, {
    "name": "Email",
    "path": ["identifier", "identityType"],
    "condition": "= 'emailAddress'"
  }]
}]
```
where condition must require `[< <= = != > >=] <literal>`.
It benefit being that it can be combined with the locales

paulbastian: Does this mean you have polymorphic claims?

sloops77: @paulbastian yes, and it's not just me. https://1edtech.github.io/openbadges-specification/ob_v3p0.html#identityobject

examples/credential_metadata_sd_jwt_vc.json
sschulz-t: +1 for this! I think supporting this is mandatory for having a clear and defined way to present this kind of data to the user. Let's use the example from your documentation and extend it with an arbitrary example:
```
... 
"address": {
    "street_address": "42 Market Street",
    "city": "Milliways",
    "postal_code": "12345"
},
...
"work_address":  {
   "city": "Berlin",
  ...
},
```

I would expect/imagine this to be displayed in the ui similar to this:

```
place of residence
+- street: 42 Market Street
+- city: Milliways
`- post code: 12345
....
place of work
+- city: Berlin
+...
```

examples/authorization_details_mso_doc.json
```diff
@@ -3,14 +3,14 @@
...
-        "organ_donor": {}
-      }
+      "org.iso.18013.5.1": [
+        {"path": ["given_name"]},
+        {"path": ["family_name"]},
+        {"path": ["birth_date"]}
+      ],
+      "org.iso.18013.5.1.aamva": [
+        {"path": ["organ_donor"]}
+      ]
```
danielfett, line 13: This could be changed to use the same structure as in the new query language (namespace/claim_name)

openid-4-verifiable-credential-issuance-1_0.md
danielfett: I have fixed this in the latest revision.

Sakurann:
agreed to come back to this after agreeing on the new query language design first, since there is an overlap with the new query language (#266) how to clarify how to specify which claim (uses path rn)

openid-4-verifiable-credential-issuance-1_0.md
danielfett: We only have `null` in DCQL in OpenID4VP for now. I suggest that we extend this as need as soon as we have a concrete use case.

danielfett:
Todos for me:
- [ ] Ensure that the mso_mdoc format uses the namespace/claim_name syntax as DCQL does
- [ ] Check if the text for the claims path queries is the same as in DCQL
- [ ] Make the change described by Paul here: https://github.com/openid/OpenID4VCI/pull/276#discussion_r1499627153

openid-4-verifiable-credential-issuance-1_0.md
danielfett: Thanks, I have fixed this, please review the latest version.

openid-4-verifiable-credential-issuance-1_0.md
danielfett: Please file this request for finer-grained value types as a separate issue.

danielfett:
Updated, please re-review.

examples/authorization_details_jwt_vc_json.json
```diff
@@ -3,11 +3,11 @@
     "type": "openid_credential",
     "credential_configuration_id": "UniversityDegreeCredential",
     "credential_definition": {
-      "credentialSubject": {
-        "given_name": {},
-        "family_name": {},
-        "degree": {}
-      }
+      "credentialSubject": [
```
danielfett, line 6: This PR would be the opportunity to rename this to just 'claims'. The open question is whether this should then match all claims or start within "credentialSubject".

sloops77: @danielfett as I mentioned on the wgc it is possible for `claims` to be in the root of the vc in vcdm (eg.  openbadgecredential). Therefore it needs to be 
```
  "claims": {
    "path": ["credentialSubject", "given_name"], 
    "path": ["credentialSubject", "family_name"], 
    ...
   }
```

danielfett: This is a very good point and I'll make this change accordingly.

danielfett: In the latest change, I aligned the "claims" element between all credential formats, i.e., moved it one level higher, called it "claims" instead of "credentialSubject" and in the examples made the change shown in @sloops77's example.

examples/authorization_details_ldp_vc.json
danielfett: @peppelinux @TakahikoKawasaki @vdzhuvinov This has been resolved with my latest change to this PR: https://github.com/openid/OpenID4VCI/pull/276#discussion_r1860262845

openid-4-verifiable-credential-issuance-1_0.md
danielfett: This text is no longer present in the new version, so I'll close this.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2293,6 +2268,178 @@ The following is a non-normative example of a Credential Response containing a C
...
+## Claims Description for Issuer Metadata {#claims-description-issuer-metadata}
+
+A claims description object as used in the Credential Issuer metadata is an
+object used to describe how a certain claim in the Credential should be
+displayed to the End-User. It is used in the `claims`
+parameter in the Credential Issuer metadata defined in (#format-profiles). The
+following keys can be used to describe the claim or claims:
+
+  * `path`: REQUIRED if the Credential Format uses a JSON-based claims
+    structure; MUST NOT be present otherwise. The value MUST be a non-empty
```
jogu, line 2305: I think this is discriminatory to Credential Formats using YAML based claims structures... ;-)

More seriously I do wonder if this goes a little too far with the normative language - it seems unnecessary to completely prevent someone defining a new non-JSON hierarchical credential format from using the `path` key for example.

jogu (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2293,6 +2268,178 @@ The following is a non-normative example of a Credential Response containing a C
...
+    array representing a claims path pointer that specifies the path to a claim
+    within the credential, as defined in (#claims_path_pointer).
+  * `namespace`: REQUIRED if the Credential Format is based on the mdoc format
+    defined in ISO 18013-5; MUST NOT be present otherwise. The value MUST be a
+    string that specifies the namespace of the data element within the mdoc,
+    e.g., `org.iso.18013.5.1`.
```
Sakurann, line 2311:
```suggestion
  * `path`: REQUIRED if the Credential Format uses a JSON-based claims
    structure; MUST NOT be present otherwise. The value MUST be a non-empty
    array representing a claims path pointer that specifies the path to a claim
    within the credential, as defined in (#claims_path_pointer).
  * `namespace`: REQUIRED if the Credential Format is based on the mdoc format
    defined in ISO 18013-5; MUST NOT be present otherwise. The value MUST be a
    string that specifies the namespace of the data element within the mdoc,
    e.g., `org.iso.18013.5.1`.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2293,6 +2268,178 @@ The following is a non-normative example of a Credential Response containing a C
...
+  * `claim_name`: REQUIRED if the Credential Format is based on mdoc format
+    defined in ISO 18013-5; MUST NOT be present otherwise. The value MUST be a
+    string that specifies the data element identifier of the data element within
+    the provided namespace in the mdoc, e.g., `first_name`.
```
Sakurann, line 2315:
```suggestion
  * `path`: REQUIRED if the Credential Format uses a JSON-based claims
    structure; SHOULD NOT be present otherwise. The value MUST be a non-empty
    array representing a claims path pointer that specifies the path to a claim
    within the credential, as defined in (#claims_path_pointer).
  * `namespace`: REQUIRED if the Credential Format is based on the mdoc format
    defined in ISO 18013-5; SHOULD NOT be present otherwise. The value MUST be a
    string that specifies the namespace of the data element within the mdoc,
    e.g., `org.iso.18013.5.1`.
  * `claim_name`: REQUIRED if the Credential Format is based on mdoc format
    defined in ISO 18013-5; SHOULD NOT be present otherwise. The value MUST be a
    string that specifies the data element identifier of the data element within
    the provided namespace in the mdoc, e.g., `first_name`.
```

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: will be resolved with the below suggestion

Sakurann (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2293,6 +2268,178 @@ The following is a non-normative example of a Credential Response containing a C
...
+    array representing a claims path pointer that specifies the path to a claim
+    within the credential, as defined in (#claims_path_pointer).
+  * `namespace`: REQUIRED if the Credential Format is based on the mdoc format
+    defined in ISO 18013-5; SHOULD NOT be present otherwise. The value MUST be a
+    string that specifies the namespace of the data element within the mdoc,
+    e.g., `org.iso.18013.5.1`.
+  * `claim_name`: REQUIRED if the Credential Format is based on mdoc format
+    defined in ISO 18013-5; SHOULD NOT be present otherwise. The value MUST be a
+    string that specifies the data element identifier of the data element within
+    the provided namespace in the mdoc, e.g., `first_name`.
```
jogu, line 2315: I wonder if we should change (remove namespace/claim_name) as per the discussion on https://github.com/openid/OpenID4VP/issues/293

Sakurann: +1 to changing

examples/authorization_details_mso_doc.json
```diff
@@ -2,15 +2,23 @@
...
     "credential_configuration_id": "org.iso.18013.5.1.mDL",
-    "claims": {
-      "org.iso.18013.5.1": {
-        "given_name": {},
-        "family_name": {},
-        "birth_date": {}
+    "claims": [
+      {
+        "namespace": "org.iso.18013.5.1",
+        "claim_name": ["given_name"]
```
tplooker, line 8: Based on the definition [here](https://github.com/openid/OpenID4VCI/pull/276/files#diff-1f424614b35a9899813079f1b1f6218631a2aedd993368ccb89bb81a9eda0289R2312-R2315)

Should all instances of `claim_name` have a value type of string not an array of strings?
```suggestion
        "claim_name": "given_name"
```

examples/credential_issuer_metadata_jwt_vc_json.json
```diff
@@ -34,38 +34,43 @@
       "credential_signing_alg_values_supported": [
         "ES256"
       ],
-      "credential_definition": {
+      "credential_definition":{
```
awoie, line 37: Isn't `@contex` required?

examples/credential_metadata_ldp_vc.json
```diff
@@ -16,35 +16,40 @@
...
-            "display": [
-              {
-                "name": "GPA"
-              }
-            ]
-          }
-        }
+        ]
       },
+      "claims": [
```
awoie, line 21: Isn't `@context` required?

examples/authorization_details_mso_doc.json
```diff
@@ -2,15 +2,23 @@
...
     "type": "openid_credential",
     "credential_configuration_id": "org.iso.18013.5.1.mDL",
-    "claims": {
-      "org.iso.18013.5.1": {
-        "given_name": {},
-        "family_name": {},
-        "birth_date": {}
+    "claims": [
+      {
+        "namespace": "org.iso.18013.5.1",
```
paulbastian, line 7: this is missing path and does not match the claims description further down

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2293,6 +2268,178 @@ The following is a non-normative example of a Credential Response containing a C
...
+For a claims description object, the following keys defined by this
+specification can be used to describe the claim or claims (other keys MAY also
+be used):
+
+  * `path`: REQUIRED. Array. Claim path pointer as defined in (#claims_path_pointer)
+    to identify the claim(s) in the Credential.
+  * `mandatory`: OPTIONAL. Boolean which, when set to `true`, indicates that the
+    Wallet will only accept a Credential that includes this claim. If set to
+    `false`, the claim is not required to be included in the Credential. If the
+    `mandatory` parameter is omitted, the default value is `false`.
```
paulbastian, line 2292: I think authorization details really starts making sense, if we add `value` here, letting the Client say that he wants a Credential with particular values, which would be possible now. Is there any support for this?

examples/credential_metadata_mso_mdoc.json
```diff
@@ -33,36 +31,41 @@
...
-              }
-            ]
-          },
-          "birth_date": {
-            "mandatory": true
-          }
+      "claims": [
+        {
+          "namespace": "org.iso.18013.5.1",
+          "claim_name": "given_name",
```
paulbastian, line 37: I think using `path` for mdoc as well and use namespace+cliam_name as two array elements seems cleaner to me

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2293,6 +2268,178 @@ The following is a non-normative example of a Credential Response containing a C
...
+    defined in ISO 18013-5; SHOULD NOT be present otherwise. The value MUST be a
+    string that specifies the data element identifier of the data element within
+    the provided namespace in the mdoc, e.g., `first_name`.
+  * `mandatory`: OPTIONAL. Boolean which, when set to `true`, indicates that the
+    Credential Issuer will always include this claim in the issued Credential.
+    If set to `false`, the claim is not included in the issued Credential if the
+    wallet did not request the inclusion of the claim, and/or if the Credential
+    Issuer chose to not include the claim. If the `mandatory` parameter is
+    omitted, the default value is `false`.
+  * `value_type`: OPTIONAL. String value determining the type of value of the
```
paulbastian, line 2322: To me the value_type doesn't seem very thoughtthrough/complete, maybe its better to leave it out for this merge and add it later if we really want it

paulbastian (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
c2bo: relevant issue: #262
I still think it would make sense to add as an optional feature for cases where you already know what you want (e.g., from another app) and want to provide a hint to the Issuer.

c2bo (commented):
It feels like we are mixing 2 aspects here: meta information for display and schema information.

Do we really need both? Do we need things like value types or information like this in metadata:
```json
{"path": ["address", "locality"]},
{"path": ["address", "region"]},
{"path": ["address", "country"]},
{"path": ["birthdate"]},
{"path": ["is_over_18"]},
```

I like the way the new display description works with its alignment to DCQL, but I feel we are trying to overload with too much (or I am not fully understanding the requirement which might very well be the case).

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: I wouldn't support polymorphic claims, it seems this is a bad design pattern to me

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: Discussed on the WG call today: tendency to do the change and do it now in VP and VCI.

examples/credential_metadata_mso_mdoc.json
tlodderstedt: Discussed on the WG call today: tendency to do the change and do it now in VP and VCI.

openid-4-verifiable-credential-issuance-1_0.md
sloops77: I agree that it's not a good design. However, I believe that OpenBadgeCredential v3 is one of the top 2 most popular credential types in the world. So, deciding not to support this feature in OpenID4VCI needs to be done consciously and with outreach back to 1edtech

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: Please open another issuer/PR so that we can discuss it there

examples/authorization_details_mso_doc.json
```diff
@@ -2,15 +2,23 @@
...
+        "claim_name": "given_name"
```
Sakurann, line 8:
```suggestion
        "path": [ "org.iso.18013.5.1", "given_name"],
```

examples/authorization_details_mso_doc.json
Sakurann: made a suggestion below

examples/credential_issuer_metadata_jwt_vc_json.json
Sakurann: we used to have `@context` but previously had that discussion before and agreed to remove it

examples/credential_metadata_jwt_vc_json.json
```diff
@@ -9,39 +9,44 @@
       "credential_signing_alg_values_supported": [
         "ES256"
       ],
-      "credential_definition": {
+      "credential_definition":{
```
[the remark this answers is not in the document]
Sakurann: per above, it is not

examples/credential_metadata_ldp_vc.json
Sakurann: for `ldp_vc`, it is required. https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#appendix-A.1.2.2-3.1.2.1

Sakurann:
```suggestion
      "@context": [
          "https://www.w3.org/2018/credentials/v1",
          "https://www.w3.org/2018/credentials/examples/v1"
        ],
        "claims": [
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2085,15 +2085,9 @@ Cryptographic algorithm names used in the `credential_signing_alg_values_support
 
 The following additional Credential Issuer metadata parameters are defined for this Credential Format for use in the `credential_configurations_supported` parameter, in addition to those defined in (#credential-issuer-parameters).
 
-* `credential_definition`: REQUIRED. Object containing the detailed description of the Credential type. It consists of the following parameters defined by this specification:
+* `claims`: OPTIONAL. An array of claims description objects as defined in (#claims-description-issuer-metadata).
+* `credential_definition`: REQUIRED. Object containing the detailed description of the Credential type. It consists of the following parameter:
```
Sakurann, line 2089: why is `claims` not part of `credential_definition`?

openid-4-verifiable-credential-issuance-1_0.md
sloops77: @paulbastian continued in https://github.com/openid/OpenID4VCI/issues/422

openid-4-verifiable-credential-issuance-1_0.md
sloops77: @paulbastian i agree (I made a comment earlier listing a number of `value_type` values that are required to correctly format rendered claims). I think removing `value_type` for now and adding it back in with more consideration would make sense

examples/credential_metadata_sd_jwt_vc.json
```diff
@@ -54,19 +56,17 @@
...
-          "region": {},
-          "country": {}
-        },
-        "birthdate": {},
-        "is_over_18": {},
-        "is_over_21": {},
-        "is_over_65": {}
-      }
+        {"path": ["email"]},
+        {"path": ["phone_number"]},
```
mickrau, line 60:
```suggestion
        {"path": ["phone_number"]},
        {
          "path": ["address"],
          "display": [
            {
              "name": "Place of residence",
              "locale": "en-US"
            },
            {
              "name": "Wohnsitz",
              "locale": "de-DE"
            }
          ]
        }
        {"path": ["address", "street_address"]},
```

Add display information to non-leaf claim, as discussed in https://github.com/openid/OpenID4VCI/pull/276#discussion_r1497169863

openid-4-verifiable-credential-issuance-1_0.md
danielfett: It is a top-level element for all other formats, so should be one here as well.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2293,6 +2268,178 @@ The following is a non-normative example of a Credential Response containing a C
...
+    the provided namespace in the mdoc, e.g., `first_name`.
+  * `mandatory`: OPTIONAL. Boolean which, when set to `true`, indicates that the
+    Credential Issuer will always include this claim in the issued Credential.
+    If set to `false`, the claim is not included in the issued Credential if the
+    wallet did not request the inclusion of the claim, and/or if the Credential
+    Issuer chose to not include the claim. If the `mandatory` parameter is
+    omitted, the default value is `false`.
+  * `value_type`: OPTIONAL. String value determining the type of value of the
+    claim. Valid values defined by this specification are `string`, `number`,
+    and image media types such as `image/jpeg` as defined in IANA media type
```
babisRoutis, line 2324: `boolean` ?

openid-4-verifiable-credential-issuance-1_0.md
babisRoutis: @paulbastian , @sloops77 I also agree with this (to remove `value_type` from this PR)

Current values (`string`, `number` etc) make little sense for `mdoc`, as  @sloops77 commented a while ago.

I tend to believe that this is also the case for JSON based credentials. 
For instance, is it helpful to have for claim `birthdate` a value_type `string`?

babisRoutis (approved):
That's a really valuable PR, since effectively harmonizes the VCI metadata,  VP DCQL.

Have just a comment wrt `value_type`. Yet I believe it is more important to merge this PR.

examples/credential_metadata_ldp_vc.json
```diff
@@ -16,35 +16,44 @@
         "type": [
           "VerifiableCredential",
           "UniversityDegreeCredential"
+        ]
+      },
+      "@context": [
```
sloops77, line 21: @danielfett 
`@context` is not supposed to be here. its already specified under `credential_definition`
Also please fix the indentation

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2293,6 +2268,178 @@ The following is a non-normative example of a Credential Response containing a C
...
+
+Start with an empty array and repeat the following until the full path is formed. 
+
+  - To address a particular claim within an object, append the key (claim name)
+    to the array.
+  - To address an element within an array, append the index to the array (as a
+    non-negative, 0-based integer).
+  - To address all elements within an array, append a null value to the array.
+
+Verifiers MUST NOT point to the same claim more than once in a single query.
```
sloops77, line 2377: i think this should be removed or rephrased as Verifiers are not relevant to creating claims descriptions

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2085,15 +2085,9 @@ Cryptographic algorithm names used in the `credential_signing_alg_values_support
 
 The following additional Credential Issuer metadata parameters are defined for this Credential Format for use in the `credential_configurations_supported` parameter, in addition to those defined in (#credential-issuer-parameters).
 
-* `credential_definition`: REQUIRED. Object containing the detailed description of the Credential type. It consists of the following parameters defined by this specification:
+* `claims`: OPTIONAL. An array of claims description objects as defined in (#claims-description-issuer-metadata).
```
sloops77, line 2088: given that `claims` is now generic across seemingly all credential formats can `claims` be added just once to the the Credential Type Metadata Parameters section?

sloops77 (changes requested):
looks pretty good. The major outstanding issue for me is `value_type` where there are a few of us finding it limiting

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: #425

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: #426 (feel free to rename, not sure i got the name right)

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: #425

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: removing value_type for now

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: verifiers need to be changed to Credential issuers.

pmhsfelix (approved):

danielfett:
> I would prefer to exclude the `value_type`, as it seems under-specified and is not present in any example. Otherwise, I'd be ready to approve.

Done

mickrau (approved):

examples/credential_issuer_metadata_jwt_vc_json.json
```diff
@@ -34,38 +34,43 @@
...
+          "path": ["credentialSubject", "family_name"],
+          "display": [
+            {
+              "name": "Surname",
+              "locale": "en-US"
+            }
+          ]
+        },
+        {
+          "path": ["credentialSubject", "degree"]
```
selfissued, line 63: Is it intentional that there is no "display" element for the degree?

examples/credential_metadata_jwt_vc_json.json
```diff
@@ -13,35 +13,40 @@
...
+          "path": ["credentialSubject", "family_name"],
+          "display": [
+            {
+              "name": "Surname",
+              "locale": "en-US"
+            }
+          ]
+        },
+        {
+          "path": ["credentialSubject", "degree"]
```
selfissued, line 38: Is it intentional that there is no "display" element for the degree?

selfissued (approved):
Looks good.  Please answer my questions on particular lines before merging.

examples/credential_issuer_metadata_jwt_vc_json.json
c2bo: I think that is the same thing I wondered about: in the case of no "display" member, it is basically a schema information that this specific claim exists in the credential issued by this issuer.

selfissued: Should it be explicitly said in the spec what the intended behavior when no `display` member is present is?

c2bo: I will create an issue for that (tomorrow) as discussed in todays call

peppelinux (approved):

awoie (approved):
I still don't like the `path` syntax for ISO mdocs but @leecam convinced me that it would actually safe some code.

c2bo (approved):
Let's move the other discussion into separate issues - definitely a big improvement to align with the dcql syntax

sloops77 (approved):
agreed to move outstanding discussions to other tickets

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2291,6 +2266,193 @@ The following is a non-normative example of a Credential Response containing a C
...
+    to the array.
+  - To address an element within an array, append the index to the array (as a
+    non-negative, 0-based integer).
+  - To address all elements within an array, append a null value to the array.
+
+### Processing
+
+In detail, the array is processed from left to right as follows:
+
+ 1. Select the root element of the Credential, i.e., the top-level JSON object.
```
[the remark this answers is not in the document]
danielfett: I'll change the wording to "Select the root element of the Credential, i.e., the top-level object of the
    JSON structure representing the claims in the Credential"

examples/credential_metadata_ldp_vc.json
```diff
@@ -16,35 +16,44 @@
...
+          "https://www.w3.org/2018/credentials/v1",
+          "https://www.w3.org/2018/credentials/examples/v1"
         ],
```
Sakurann, line 24:
```suggestion
        "@context": [
            "https://www.w3.org/2018/credentials/v1",
            "https://www.w3.org/2018/credentials/examples/v1"
          ]
      },
```
@danielfett

examples/credential_issuer_metadata_jwt_vc_json.json
c2bo: Sorry, took a while longer due to unexpected traveling: #432

examples/credential_metadata_jwt_vc_json.json
Sakurann: #432

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2291,6 +2266,193 @@ The following is a non-normative example of a Credential Response containing a C
...
+For a claims description object, the following keys defined by this
+specification can be used to describe the claim or claims (other keys MAY also
+be used):
+
+  * `path`: REQUIRED. Array. Claim path pointer as defined in (#claims_path_pointer)
+    to identify the claim(s) in the Credential.
+  * `mandatory`: OPTIONAL. Boolean which, when set to `true`, indicates that the
+    Wallet will only accept a Credential that includes this claim. If set to
+    `false`, the claim is not required to be included in the Credential. If the
+    `mandatory` parameter is omitted, the default value is `false`.
```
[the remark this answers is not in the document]
Sakurann: good point. opened #433

PR #277, opened 2024-02-23
Addresses issue #273

### LINKED ISSUE #273: Batch Credential Response, description correction and improvement
The protocol provides the following description:

> The following parameters are used in the Batch Credential Response:
> 
> credential_responses: REQUIRED. Array that contains Credential Response objects, 
> as defined in (#credential-request), and/or Deferred Credential Response objects, as defined in (#deferred-credential-request). 
> Every entry of the array corresponds to the Credential Request object at the same array 
> index in the credential_requests parameter of the Batch Credential Request.
> c_nonce: OPTIONAL. The c_nonce as defined in (#credential-response).
> c_nonce_expires_in: OPTIONAL. The c_nonce_expires_in as defined in (#credential-response).
> 

I think there are two issues

- [ ] In the definition of `credential_responses` two links are being provided pointing to `credential-request` and `deferred-credential-request`. This is not correct though because the text describes credential responses (not requests)
- [ ] The description of `credential_responses` is perhaps not absolutely correct. Reading it literally one could  think that each member of the `credential_responses` array should be exactly like either a credential or deferred response. I think though that's not the case because those responses contain also `c_nonce` & `c_nonce_expires_in`  whereas batch response considers them (correctly) top-level attributes.

babisRoutis:
Closes #273

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1087,7 +1087,7 @@ A successful Batch Credential Response MUST contain all the requested Credential
 
 The following parameters are used in the Batch Credential Response:
 
-* `credential_responses`: REQUIRED. Array that contains Credential Response objects, as defined in (#credential-response), and/or Deferred Credential Response objects, as defined in (#deferred-credential-response). Every entry of the array corresponds to the Credential Request object at the same array index in the `credential_requests` parameter of the Batch Credential Request.
+* `credential_responses`: REQUIRED. Array that contains Credential Response objects, as defined in (#credential-response), and/or Deferred Credential Response objects, as defined in (#deferred-credential-response), except `c_nonce` and `c_nonce_expires_in`. Every entry of the array corresponds to the Credential Request object at the same array index in the `credential_requests` parameter of the Batch Credential Request.
 * `c_nonce`: OPTIONAL. The `c_nonce` as defined in (#credential-response). 
 * `c_nonce_expires_in`: OPTIONAL. The `c_nonce_expires_in` as defined in (#credential-response).
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1087,7 +1087,7 @@ A successful Batch Credential Response MUST contain all the requested Credential
 
 The following parameters are used in the Batch Credential Response:
 
-* `credential_responses`: REQUIRED. Array that contains Credential Response objects, as defined in (#credential-response), and/or Deferred Credential Response objects, as defined in (#deferred-credential-response), except `c_nonce` and `c_nonce_expires_in`. Every entry of the array corresponds to the Credential Request object at the same array index in the `credential_requests` parameter of the Batch Credential Request.
+* `credential_responses`: REQUIRED. Array that contains Credential Response objects, as defined in (#credential-response), and/or Deferred Credential Response objects, as defined in (#deferred-credential-response), with the exception of `c_nonce` and `c_nonce_expires_in`. Each element within the array matches the corresponding Credential Request object by array index in the `credential_requests` parameter of the Batch Credential Request.
 * `c_nonce`: OPTIONAL. The `c_nonce` as defined in (#credential-response). 
 * `c_nonce_expires_in`: OPTIONAL. The `c_nonce_expires_in` as defined in (#credential-response).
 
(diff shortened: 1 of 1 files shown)


peppelinux:
@babisRoutis I agree about the issue https://github.com/openid/OpenID4VCI/issues/273

at the same time this PR exposes to me the evidence that c_once and c_nonce_expires_in are part of the credential response and that using the batch endpoint the parameter credential response is misleading

I propose to change the response schema, WDYT about this?

````
{
  "credentials": [
     "eyJraWQiOiJkaWQ6ZXhhbXBsZTpl...C_aZKPxgihac0aW9EkL1nOzM"
     "YXNkZnNhZGZkamZqZGFza23....29tZTIzMjMyMzIzMjMy"
   ],
  "c_nonce": "fGFF7UkhLa",
  "c_nonce_expires_in": 86400
}
````

In addition to this proposal, I found another issue not related to your PR, when I read:
 `credential_responses: REQUIRED. Array that contains Credential Response objects, as defined in (#credential-request)`
 
 well, as we can read, `credential_responses`:
 
 1. is not defined in Section 7.2
 2. would not be intended to be defined in the section related to the credential request 
 
these are another two reasons to renew the batch response schema, by simplify it using an array of credentials, named `credentials`.

WDYT?

**Comment by babisRoutis** [comment]
> I found another issue not related to your PR, when I read: `credential_responses: REQUIRED. Array that contains Credential Response objects, as defined in (#credential-request)`
> 
> well, as we can read, `credential_responses`:
> 
>     1. is not defined in Section 7.2
> 
>     2. would not be intended to be defined in the section related to the credential request

Hi @peppelinux 
To be frank, I am not sure if your comment is about fixing something in the PR :smile:

**Review by bc-pi — APPROVED** [review]

File: openid-4-verifiable-credential-issuance-1_0.md [hunk 1]
```diff
@@ -1087,7 +1087,7 @@ A successful Batch Credential Response MUST contain all the requested Credential
 
 The following parameters are used in the Batch Credential Response:
 
-* `credential_responses`: REQUIRED. Array that contains Credential Response objects, as defined in (#credential-request), and/or Deferred Credential Response objects, as defined in (#deferred-credential-request). Every entry of the array corresponds to the Credential Request object at the same array index in the `credential_requests` parameter of the Batch Credential Request.
+* `credential_responses`: REQUIRED. Array that contains Credential Response objects, as defined in (#credential-response), and/or Deferred Credential Response objects, as defined in (#deferred-credential-response), except `c_nonce` and `c_nonce_expires_in`. Every entry of the array corresponds to the Credential Request object at the same array index in the `credential_requests` parameter of the Batch Credential Request.
```
peppelinux, line 1090:
```suggestion
* `credential_responses`: REQUIRED. Array that contains Credential Response objects, as defined in (#credential-response), and/or Deferred Credential Response objects, as defined in (#deferred-credential-response), with the exception of `c_nonce` and `c_nonce_expires_in`. Each element within the array matches the corresponding Credential Request object by array index in the `credential_requests` parameter of the Batch Credential Request.
```

peppelinux (approved):
ILGTM 

I shared a quick proposal about the text, minor suggestion.

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: Suggested change  by @peppelinux  LGTM

Sakurann: LGTM too. will merge this PR when this suggestion is incorporated.

babisRoutis: Done Kristina. Please check

jogu (approved):
As a general comment (not requesting any change to this PR) I'm not fond of this "use object defined in 'x' but 'y'" where section 'x' is not intending to define a reusable object - we've seen multiple times that it results in the specification defining more parameters than it should, unintentionally.

I think it would be cleaner and less error prone if an object that is used in multiple requests/responses is defined in a separate section that is defining only the common object.

Thanks for the PR Babis!

bc-pi:
Concur with @jogu's general comment FWIW (and considered saying something similar) but this PR is a pragmatic fix given the time frame and situation.

Sakurann:
@jogu @bc-pi 
> I think it would be cleaner and less error prone if an object that is used in multiple requests/responses is defined in a separate section that is defining only the common object.

> Concur with @jogu's general comment FWIW (and considered saying something similar) but this PR is a pragmatic fix given the time frame and situation.

did one of you wanted to open an issue/PR on that topic?

jogu:
> did one of you wanted to open an issue/PR on that topic?

My inclination was not to address it until we need to make changes to one of the objects or something like that. I really just wanted to make sure we were all aware that it's not a good pattern so we try to avoid it when we create new text. Happy to raise an issue if you think it's worth just trying to fix all the current instances now, though it's probably not a priority.

PR #280, opened 2024-02-28
Add a note about the known limitations of the credential claim metadata

jogu:
As discussed on yesterday's working group call:

https://github.com/openid/OpenID4VCI/issues/266#issuecomment-1967533918

As it seems likely that https://github.com/openid/OpenID4VCI/pull/276 will not make it into implementer's draft 1, we instead add warnings about the known limitations of the current data structure so that implementer's are aware.

Note that I have not applied this to the mdoc profile section, as that sections seems not to allow nested objects in the metadata.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1995,6 +1995,15 @@ The following additional Credential Issuer metadata parameters are defined for t
           * `locale`: OPTIONAL. String value that identifies language of this object represented as language tag values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
 * `order`: OPTIONAL. Array of the claim name values that lists them in the order they should be displayed by the Wallet.
 
+> The above metadata structure has some known limitations:
+>
+> * It cannot be used to describe claims in credentials that have the name `mandatory`, `value_type` or `display`.
+> * It is not possible to provide `mandatory`, `value_type` or `display` values for objects that contain claims 
+> * The `order` parameter cannot be used for claims within objects
+> * Arrays of unknown size cannot be described
+>
+> These limitations are expected to be resolved in the second Implementer's Draft, a proposal can be viewed in [Issue 266](https://github.com/openid/OpenID4VCI/issues/266).
+
 The following is a non-normative example of an object containing the `credential_configurations_supported` parameter for Credential format `jwt_vc_json`:
 
 <{{examples/credential_metadata_jwt_vc_json.json}}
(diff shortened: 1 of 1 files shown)


jogu:
It case it's helpful here's how this renders:

Screenshot 2024-02-28 at 13 56 44

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1995,6 +1995,15 @@ The following additional Credential Issuer metadata parameters are defined for t
           * `locale`: OPTIONAL. String value that identifies language of this object represented as language tag values defined in BCP47 [@!RFC5646]. There MUST be only one object for each language identifier.
 * `order`: OPTIONAL. Array of the claim name values that lists them in the order they should be displayed by the Wallet.
 
+> The above metadata structure has some known limitations:
+>
+> * It cannot be used to describe claims in credentials that have the name `mandatory`, `value_type` or `display`.
+> * It is not possible to provide `mandatory`, `value_type` or `display` values for objects that contain claims
```
Sakurann, line 2001: this is the most important and should be the first one. and it should also be much more clearer.
```suggestion
> * It is not possible to provide `mandatory`, `value_type` or `display` values for the intermediary claims in the nested objects
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1995,6 +1995,15 @@ The following additional Credential Issuer metadata parameters are defined for t
...
 
+> The above metadata structure has some known limitations:
+>
+> * It cannot be used to describe claims in credentials that have the name `mandatory`, `value_type` or `display`.
+> * It is not possible to provide `mandatory`, `value_type` or `display` values for objects that contain claims 
+> * The `order` parameter cannot be used for claims within objects
+> * Arrays of unknown size cannot be described
+>
+> These limitations are expected to be resolved in the second Implementer's Draft, a proposal can be viewed in [Issue 266](https://github.com/openid/OpenID4VCI/issues/266).
+
```
Sakurann, line 2006: we have not used this `> ` format in other places in the spec. I would like to avoid introducing it and drawing unnecessary attention to this

```suggestion
Note that the above metadata structure has some known limitations:

* It cannot be used to describe claims in credentials that have the name `mandatory`, `value_type` or `display`.
* It is not possible to provide `mandatory`, `value_type` or `display` values for objects that contain claims 
* The `order` parameter cannot be used for claims within objects
* Arrays of unknown size cannot be described

These limitations are expected to be resolved in the second Implementer's Draft, a proposal can be viewed in [Issue 266](https://github.com/openid/OpenID4VCI/issues/266).

```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1995,6 +1995,15 @@ The following additional Credential Issuer metadata parameters are defined for t
...
 * `order`: OPTIONAL. Array of the claim name values that lists them in the order they should be displayed by the Wallet.
 
+> The above metadata structure has some known limitations:
+>
+> * It cannot be used to describe claims in credentials that have the name `mandatory`, `value_type` or `display`.
+> * It is not possible to provide `mandatory`, `value_type` or `display` values for objects that contain claims 
+> * The `order` parameter cannot be used for claims within objects
+> * Arrays of unknown size cannot be described
+>
+> These limitations are expected to be resolved in the second Implementer's Draft, a proposal can be viewed in [Issue 266](https://github.com/openid/OpenID4VCI/issues/266).
```
Sakurann, line 2005: I am not sure about the current framing... I think the following is sufficient.
```suggestion
> Mechanisms that could address these limitations are being discussed in [Issue 266](https://github.com/openid/OpenID4VCI/issues/266).
```

Sakurann (changes requested):
reading the PR, honestly, not sure this should be in the ID-1 text...

made some suggestions still..

jogu:
This was discussed on last Thursday's working group call and there wasn't anyone that felt strongly about merging this for ID2. I suggest it can probably be closed.

jogu:
closing as we decided not to include this before ID1.

PR #289, opened 2024-04-03
Fix `uri` field name in `logo` objects in examples

sietseringers:
Per the definition of the data structure that this example shows, the `logo` of the display properties of a Credential should have an `uri` field instead of `url` field.


jogu:
Huh, good catch. But there's also 4 other examples where this same mistake is made?

sietseringers:
Right! Fixed those too.

jogu (approved):

c2bo (approved):

tplooker (approved):

PR #290, opened 2024-04-05
fix ldp_vc issuer metadata example

nempat:
remove '@context' and 'type' attributes from top level issuer metadata document for ldp_vc credential format as these two have been moved to 'credential_definition' structure


Sakurann (approved):

Sakurann:
good catch! thank you!

PR #314, opened 2024-05-06
remove use of the `authorization_pending` and `slow_down` error codes

bc-pi:
per issue #60

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -133,7 +133,6 @@ Existing OAuth 2.0 mechanisms are extended as following:
 
 * A new Grant Type "Pre-Authorized Code" is defined to facilitate flows where the preparation of the Credential issuance is conducted before the actual OAuth flow starts (see (#pre-authz-code-flow)).
 * A new authorization details [@!RFC9396] type `openid_credential` is defined to convey the details about the Credentials (including formats and types) the Wallet wants to obtain (see (#authorization-details)).
-* The Token Response error codes `authorization_pending` and `slow_down` are used to allow for deferred authorization of Credential issuance. These error codes are supported for the Pre-Authorized Code grant type.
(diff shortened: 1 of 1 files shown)


Sakurann (approved):

paulbastian (approved):

jogu (approved):

c2bo (approved):

awoie (approved):

awoie:
I was wondering how we would support the following use case if we remove `authorization_pending`. In some cases, the AS cannot make an immediate decision on the authz request because some human intervention is needed for a final decision even though the initial credentials the user provided were successfully verified. I assume we could do that via deferred issuance, right? In that case, I was wondering what would be the error code in the credential response in case the issuer made a negative decision? Currently, I cannot see a error code that would fit. Should we add a new one?

jogu:
@awoie 

> I was wondering how we would support the following use case if we remove `authorization_pending`. In some cases, the AS cannot make an immediate decision on the authz request because some human intervention is needed for a final decision even though the initial credentials the user provided were successfully verified. I assume we could do that via deferred issuance, right? In that case, I was wondering what would be the error code in the credential response in case the issuer made a negative decision? Currently, I cannot see an error code that would fit. Should we add a new one?

That's all correct, yes. I think the new error code defined here might cover it: https://github.com/openid/OpenID4VCI/pull/321 ?

jogu:
I think we're ready to merge this now - it's been discussed on several working group calls and there were no objections raised based on the notification send to the mailing list a month ago: https://lists.openid.net/pipermail/openid-specs-digital-credentials-protocols/Week-of-Mon-20240415/000278.html

@bc-pi would you mind resolving the trivial conflict please?

awoie:
> @awoie
> 
> > I was wondering how we would support the following use case if we remove `authorization_pending`. In some cases, the AS cannot make an immediate decision on the authz request because some human intervention is needed for a final decision even though the initial credentials the user provided were successfully verified. I assume we could do that via deferred issuance, right? In that case, I was wondering what would be the error code in the credential response in case the issuer made a negative decision? Currently, I cannot see an error code that would fit. Should we add a new one?
> 
> That's all correct, yes. I think the new error code defined here might cover it: #321 ?

Yes, that would cover it. Thanks.

bc-pi:
> @bc-pi would you mind resolving the trivial conflict please?

done!

PR #318, opened 2024-05-07
clarify optionality of scope and authorization_details for Authorizat…

### LINKED ISSUE #312: Clarification on optionality of Scope vs Auhtorization Details
Section 5.1 Auth Request says "There are two possible ways to request issuance of a specific Credential type in an Authorization Request. One way is to use the authorization_details request parameter, as defined in [RFC9396], with one or more authorization details objects of type openid_credential, per Section 5.1.1. The other is through the use of scopes as defined in Section 5.1.2."
Section 5.1.2 on scope  says "In addition to a mechanism defined in Section 5.1, Credential Issuers MAY support requesting authorization to issue a Credential using the OAuth 2.0 scope parameter."

There is no equivalent phrase in Section 5.1.1 authorization_details, instead "The request parameter authorization_details defined in Section 2 of [RFC9396] MUST be used to convey the details about the Credentials the Wallet wants to obtain."

While this last MUST is probably only refering to the parameter named "authorization_details" from RFC9396, it is less clear that authorization_details is also optional

paulbastian:
Closes #312

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -424,7 +424,7 @@ When the grant type `authorization_code` is used, it is RECOMMENDED to use PKCE 
 
 An Authorization Request is an OAuth 2.0 Authorization Request as defined in Section 4.1.1 of [@!RFC6749], which requests that access be granted to the Credential Endpoint, as defined in (#credential-endpoint).
 
-There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope). A Client MUST use one of these methods to request Credentials and MAY use both simultaneously.
+There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope).
 
 ### Request Issuance of a Certain Credential Type using `authorization_details` Parameter {#authorization-details}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -426,7 +426,7 @@ When the grant type `authorization_code` is used, it is RECOMMENDED to use PKCE 
 
 An Authorization Request is an OAuth 2.0 Authorization Request as defined in Section 4.1.1 of [@!RFC6749], which requests that access be granted to the Credential Endpoint, as defined in (#credential-endpoint).
 
-There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope). A Client MUST use one of these methods but MUST NOT use both simultaneously.
+There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope). A Client MUST use one of these methods to request Credentials but MUST NOT use both simultaneously.
 
 ### Request Issuance of a Certain Credential Type using `authorization_details` Parameter {#authorization-details}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -426,7 +426,7 @@ When the grant type `authorization_code` is used, it is RECOMMENDED to use PKCE 
 
 An Authorization Request is an OAuth 2.0 Authorization Request as defined in Section 4.1.1 of [@!RFC6749], which requests that access be granted to the Credential Endpoint, as defined in (#credential-endpoint).
 
-There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope). A Client MUST use one of these methods to request Credentials but MUST NOT use both simultaneously.
+There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope). A Client MUST use one of these methods to request Credentials and MAY use both simultaneously.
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -426,10 +426,12 @@ When the grant type `authorization_code` is used, it is RECOMMENDED to use PKCE
 
 An Authorization Request is an OAuth 2.0 Authorization Request as defined in Section 4.1.1 of [@!RFC6749], which requests that access be granted to the Credential Endpoint, as defined in (#credential-endpoint).
 
-There are two possible ways to request issuance of a specific Credential type in an Authorization Request. One way is to use the `authorization_details` request parameter, as defined in [@!RFC9396], with one or more authorization details objects of type `openid_credential`, per (#authorization-details). The other is through the use of scopes as defined in (#credential-request-using-type-specific-scope).
+There are two possible ways to request issuance of a specific Credential type in an Authorization Request. One way is to use the `authorization_details` request parameter, as defined in [@!RFC9396], with one or more authorization details objects of type `openid_credential`, per (#authorization-details). The other is through the use of scopes as defined in (#credential-request-using-type-specific-scope). A Client MUST use either of them, it MUST NOT use both of them.
```
peppelinux, line 429:
```suggestion
There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope). A Client MUST use one of these methods but MUST NOT use both simultaneously.
```

peppelinux (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -426,10 +426,12 @@ When the grant type `authorization_code` is used, it is RECOMMENDED to use PKCE
 
 An Authorization Request is an OAuth 2.0 Authorization Request as defined in Section 4.1.1 of [@!RFC6749], which requests that access be granted to the Credential Endpoint, as defined in (#credential-endpoint).
 
-There are two possible ways to request issuance of a specific Credential type in an Authorization Request. One way is to use the `authorization_details` request parameter, as defined in [@!RFC9396], with one or more authorization details objects of type `openid_credential`, per (#authorization-details). The other is through the use of scopes as defined in (#credential-request-using-type-specific-scope).
+There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope). A Client MUST use one of these methods but MUST NOT use both simultaneously.
```
jogu, line 429: I suggest just clarifying that to make clear that we're not trying to say, e.g., that you can't request a credential using RAR at the same time as requesting (say) access to an API using scopes (RAR does allow authorisation details & scopes to be used in the same request in general, https://datatracker.ietf.org/doc/html/rfc9396#section-3.1 ):

```suggestion
There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope). A Client MUST use one of these methods to request Credentials but MUST NOT use both simultaneously.
```

jogu (approved):

tlodderstedt:
@paulbastian correct me if I'm wrong. This PR makes clear that scope and authorization_details are equivalent and none of it is mandatory?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -426,10 +426,12 @@ When the grant type `authorization_code` is used, it is RECOMMENDED to use PKCE
 
 An Authorization Request is an OAuth 2.0 Authorization Request as defined in Section 4.1.1 of [@!RFC6749], which requests that access be granted to the Credential Endpoint, as defined in (#credential-endpoint).
 
-There are two possible ways to request issuance of a specific Credential type in an Authorization Request. One way is to use the `authorization_details` request parameter, as defined in [@!RFC9396], with one or more authorization details objects of type `openid_credential`, per (#authorization-details). The other is through the use of scopes as defined in (#credential-request-using-type-specific-scope).
+There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope). A Client MUST use one of these methods to request Credentials but MUST NOT use both simultaneously.
```
paulbastian, line 429:
```suggestion
There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope). A Client MUST use one of these methods to request Credentials and MAY use both simultaneously.
```

paulbastian: @Sakurann like this?

jogu: We already have this text in spec:

> If a scope value related to Credential issuance and the authorization_details request parameter containing objects of type openid_credential are both present in a single request, the Credential Issuer MUST interpret these individually. However, if both request the same Credential type, then the Credential Issuer MUST follow the request as given by the authorization details object.

So the 'MUST NOT' was definitely not right. I'm okay with the 'MAY use both' language.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -424,10 +424,12 @@ When the grant type `authorization_code` is used, it is RECOMMENDED to use PKCE
 
 An Authorization Request is an OAuth 2.0 Authorization Request as defined in Section 4.1.1 of [@!RFC6749], which requests that access be granted to the Credential Endpoint, as defined in (#credential-endpoint).
 
-There are two possible ways to request issuance of a specific Credential type in an Authorization Request. One way is to use the `authorization_details` request parameter, as defined in [@!RFC9396], with one or more authorization details objects of type `openid_credential`, per (#authorization-details). The other is through the use of scopes as defined in (#credential-request-using-type-specific-scope).
+There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope). A Client MUST use one of these methods to request Credentials and MAY use both simultaneously.
```
Sakurann, line 427:
```suggestion
There are two possible methods for requesting the issuance of a specific Credential type in an Authorization Request. The first method involves using the `authorization_details` request parameter, as defined in [@!RFC9396], containing one or more authorization details of type `openid_credential`, as specified in (#authorization-details). The second method utilizes scopes, as outlined in (#credential-request-using-type-specific-scope).
```

Sakurann: yesterday's WG mtg, agreed to remove this sentence and open an issue to come back if we want to prohibit simultaneous usage of scope and RAR. since there is a sentence in the scope section of the spec that allows that rn (https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-5.1.2-12):
> If a scope value related to Credential issuance and the authorization_details request parameter containing objects of type openid_credential are both present in a single request, the Credential Issuer MUST interpret these individually. However, if both request the same Credential type, then the Credential Issuer MUST follow the request as given by the authorization details object.

Sakurann (approved):

Issue #313, opened 2024-05-06
Clarification on the Credential Endpoint


andprian:
Section 7 states that "_The Client can request issuance of a Credential of a certain type multiple times, e.g., to associate the Credential with different public keys/Decentralized Identifiers (DIDs) or to refresh a certain Credential._"

Does this mean that the Issuer is obliged to accept such requests (same type but different public keys for example) with the same Access Token?

Also, what is the expected behavior when the Credential Request is sent multiple times with exactly the same Access Token and the same parameters every time? Should the Issuer:
1. Return the same Credential or
2. Generate a new Credential every time or 
3. Reject the request if he already returned one such Credential?

peppelinux:
> Does this mean that the Issuer is obliged to accept such requests (same type but different public keys for example) with the same Access Token?

Yes, unless the access token is intended for multiple uses with a specific Resource Server, such as the Credential Endpoint.

The Credential Endpoint can issue multiple credentials if the authorization for that specific access token permits it (as specified in the authorization_details), using different JWT proofs for cryptographic key binding. Where the multiple JWT proofs are recommended for privacy reasons, to prevent credential linkability.

> Also, what is the expected behavior when the Credential Request is sent multiple times with exactly the same Access Token and the same parameters every time?

there comes in help the `c_nonce`, that in some implementation profiles is mandatory.

Good points. From my personal perspective if the parameters are the same that would be a reply -> reject.

When the same credential is requested, with a different nonce but the same jwt_proof -> revocation of the previous one and issuance of the new one.

is this something to be further explored for the "implementation considerations" section, or for the security consideration one?

andprian:
> Yes, unless the access token is intended for multiple uses with a specific Resource Server, such as the Credential Endpoint.

OK, section 6.2 brings clarification stating that "_The Authorization Server might decide to authorize issuance of multiple instances for each Credential requested in the Authorization Request._"
But then for me the paragraph in Section 7 saying the Client can request issuance of a Credential of a certain type multiple times is not completely accurate. The Client can request this, but it is not guaranteed that the Issuer supports this.

> there comes in help the c_nonce, that in some implementation profiles is mandatory.

Could you please indicate in which cases the `c_nonce` feature must be implemented by the Issuer? It seems to me that is always optional from the Issuer's point of view.

If the Issuer does not use the  `c_nonce` , should he accept replies?

[...]

Issue #320, opened 2024-05-14
Consider removing `cwt` proof type


awoie:
It would be worth discussing whether we should remove the `cwt` proof type for the following reasons:
- The encoding of `cwt` proofs is not fully specified. I guess it is supposed to be base64url of the COSE_Sign1 structure. But given that nobody complaint about this not being specified, my assumption is that there cannot be many implementers.
- Even if implementers want to avoid using JOSE in favor of COSE, they won't be able to do that in a lot of cases because they will likely need wallet attestations/oauth attestation-based client authentication that will represent the probably same keys and proofs as JWKs and JWTs. To be clear, I'm not recommending we should add `cwt` proofs to those drafts.

babisRoutis:
Totally agree.

TakahikoKawasaki:
Well, I have already implemented the CWT Key Proof for the Interop Event that [POTENTIAL](https://www.digital-identity-wallet.eu/) has been hosting. The Interop Event consists of 6 tracks. In the Track 1, CWT Key Proof is used. Demo steps using CWT Key Proof are written here in detail:

**4.3. POTENTIAL Interop Event / Track 1 / Light Profile**
https://www.authlete.com/developers/oid4vci/#43-potential-interop-event--track-1--light-profile

An open-source CBOR library ([authlete/cbor](https://github.com/authlete/cbor)) I have written contains a tool for generating a CWT Key Proof. It can be used like below.

```
git clone git@github.com:authlete/cbor.git
cd cbor
mvn compile
./bin/generate-cwt-key-proof \
    --issuer ${CREDENTIAL_ISSUER_IDENTIFIER} \
    --key ${PRIVATE_KEY_FILE} \
    --client ${CLIENT_ID} \
    --nonce ${C_NONCE}
```

The name of the property, `cwt`, naturally implies that the format of a CWT Key Proof conforms to "[RFC 8392 CBOR Web Token (CWT)](https://www.rfc-editor.org/rfc/rfc8392.html)". Therefore, **there is no ambiguity**. My OID4VCI implementation can accept the following formats.

- COSE_Sign
- COSE_Sign1
- COSE_Sign with the CWT CBOR Tag (RFC 8392, [6. CWT CBOR Tag](https://www.rfc-editor.org/rfc/rfc8392.html#section-6))
[...]

PR #293, opened 2024-04-15
rework credential and batch credential endpoint

paulbastian:
Closes #93

- [x] discuss whether proofs array should only contain the key values with the same key type instead of each object containing proof_type
- [x] add issuer metadata for support of proofs in credential endpoint
- [x] agree whether to keep or scrap batch credential endpoint or postpone the discussion

PR #323, opened 2024-05-17
chore: c_nonce is recommended and AS takes care

peppelinux:
This PR aims to resolve the issue https://github.com/openid/OpenID4VCI/issues/313 where @andprian expressed her sensibility about some details that amtters about the security and the behaviour that an implementer may expect from the AS.

This PR does not address the revocation of a credential when a request for the same credential type occurs. As noted in the related issue, this behavior may be influenced by various factors outside the scope of this specification, including legal requirements already mentioned.


bc-pi (changes requested):
IMHO c_nonce from the AS's token endpoint shouldn't even be an option and, while there wasn't enough consensus to remove it*, I'm absolutely not okay recommending it. 

* https://github.com/openid/OpenID4VCI/pull/199

andprian:
How about adding this recommendation for the `c_nonce` in section 7.3 for the Credential Response? This would prevent replay attacks for all subsequent requests.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -698,7 +698,7 @@ The Credential Endpoint issues a Credential as approved by the End-User upon pre
 
 Communication with the Credential Endpoint MUST utilize TLS. 
 
-The Client can request issuance of a Credential of a certain type multiple times, e.g., to associate the Credential with different public keys/Decentralized Identifiers (DIDs) or to refresh a certain Credential.
+The Client can request the issuance of a Credential of a certain type multiple times, for example, to associate the Credential with different public keys or Decentralized Identifiers (DIDs), or to refresh a certain Credential. However, it is important to note that the ability to do so depends on whether the Issuer supports the multiple issuance of the same Credential Type or the refresh of the Credential.
```
peppelinux, line 702: given @bc-pi 's feedback I'd suggest for now to split this PR in two different ones, where c_nonce and the clarification about multiple issuance and refresh are handled separately to prevent that our concerns about c_nonce would block an editorial clarification

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -630,7 +630,7 @@ The Authorization Server might decide to authorize issuance of multiple instance
 
 In addition to the response parameters defined in [@!RFC6749], the Authorization Server MAY return the following parameters:
 
-* `c_nonce`: OPTIONAL. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)). When received, the Wallet MUST use this nonce value for its subsequent requests until the Credential Issuer provides a fresh nonce.
+* `c_nonce`: OPTIONAL. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)). While this parameter is optional for traditional Authorization Server implementations according to [RFC6749], it is recommended as a measure to prevent replay attacks. When received, the Wallet MUST use this nonce value for its subsequent requests until the Credential Issuer provides a fresh nonce. It is at the Credential Issuer's discretion how many copies of a Credential it is willing to issue, how often, and whether reissued Credentials automatically update if the underlying claims have changed.
```
pmhsfelix, line 633: Not clear to me how the sentence "It is at the Credential Issuer's discretion how many copies of a Credential it is willing to issue, how often, and whether reissued Credentials automatically update if the underlying claims have changed" is related to the `c_nonce` returned in a token response. What am I missing?

paulbastian, line 633: The suggested text is partially covered in #293 and partially doesn't fit into this part of the specification.
```suggestion
* `c_nonce`: OPTIONAL. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)). While this parameter is optional for traditional Authorization Server implementations according to [RFC6749], it is recommended as a measure to prevent replay attacks. When received, the Wallet MUST use this nonce value for its subsequent requests until the Credential Issuer provides a fresh nonce.
```

paulbastian (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: I made a mistake due to a distraction, it was in a wrong place. now fixed, thanks to you

peppelinux: moved in another place, thank you

Sakurann:
if the point of this PR is to recommend the usage of c_nonce in general (regardless of how the issuer provides c_nocne), the specification right now actually intends to mandate c_nonce (either from the token endpoint or credential endpoint)....

i am inclined to close this issue until there is more clarity on the issues including #320

Sakurann:
with the discussion in #331, the direction is to remove an option to return c_nonce from the token endpoint

PR #325, opened 2024-05-22
clarifications about the uniqueness of the c_nonce value

peppelinux:
This PR aims to resolve part of the points contained in the issue: https://github.com/openid/OpenID4VCI/issues/313#issuecomment-2121942187


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -630,7 +630,7 @@ The Authorization Server might decide to authorize issuance of multiple instance
 
 In addition to the response parameters defined in [@!RFC6749], the Authorization Server MAY return the following parameters:
 
-* `c_nonce`: OPTIONAL. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)). When received, the Wallet MUST use this nonce value for its subsequent requests until the Credential Issuer provides a fresh nonce.
+* `c_nonce`: OPTIONAL. String containing a nonce for key proof possession (see #credential-request). The Wallet MUST use this nonce for subsequent requests until a new nonce is issued by the Credential Issuer. Each issued `c_nonce` value MUST be unique, a Credential Issuer cannot issue duplicate `c_nonces` values.
```
bc-pi, line 633: Seems inappropriate to place requirements on the Credential Issuer in this section about the Authorization Server

bc-pi (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
paulbastian, line 633: Agree, this section should talk about the authorisation server. I'm not really sure if this requirement is a must or recommend only.

paulbastian (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: > not really sure if this requirement is a must or recommend only

or necessary at all ...

peppelinux: yes, I agree, it is now moved in the credential endpoint section.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -728,7 +728,9 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
 
 The `proof_type` parameter is an extension point that enables the use of different types of proofs for different cryptographic schemes.
 
-The `proof` element MUST incorporate the Credential Issuer Identifier (audience), and a `c_nonce` value generated by the Authorization Server or the Credential Issuer to allow the Credential Issuer to detect replay. The way that data is incorporated depends on the key proof type. In a JWT, for example, the `c_nonce` value is conveyed in the `nonce` claim, whereas the audience is conveyed in the `aud` claim. In a Linked Data proof, for example, the `c_nonce` is included as the `challenge` element in the key proof object and the Credential Issuer (the intended audience) is included as the `domain` element.
+The `proof` element MUST incorporate the Credential Issuer Identifier (audience), and a `c_nonce` value generated by the Authorization Server or the Credential Issuer to allow the Credential Issuer to detect replay.  The Wallet MUST use this nonce for subsequent requests until a new nonce is issued by the Credential Issuer. Each issued `c_nonce` value MUST be unique, a Credential Issuer cannot issue duplicate `c_nonces` values.
+
```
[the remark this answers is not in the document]
peppelinux: if we use a nonce more than once we should not use the term nonce. Is mine a wrong assumption or ... ?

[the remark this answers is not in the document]
pmhsfelix: Agree that the `nonce` name is not the best one, however the multiple usage by the wallet is already allowed by the sentence "The Wallet MUST use this nonce for subsequent requests until a new nonce is issued by the Credential Issuer".
If this multiple usage by the wallet is already allowed, I don't see any advantage on requiring the AS/CI to create unique _nonces_.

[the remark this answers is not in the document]
peppelinux: In my interpretation the text doesn't mandate the use of the same nonce value but the same nonce parameter, since there's the mention to the "new nonce" indicating that in each response a brand new nonce is issued.

I propose to apply normative language to clarify this.

TakahikoKawasaki:
FYI: The following is my intepretation when I implemented `c_nonce`.

- **Uniqueness**: When a server issues a new `c_nonce`, its value must differ from any previously issued `c_nonce` values.
- **Multiple Use**: A wallet can use a `c_nonce` value multiple times until the server rejects it.
- **Issuance Timing**: A `c_nonce` in a server response is not always new. Including a `c_nonce` in the response does not necessarily mean a new one has been issued. In other words, the server may include the latest valid `c_nonce` it issued to the wallet in the past instead of issuing a new one.

peppelinux:
@TakahikoKawasaki ILGTM with the following notes in addition:

1. **Uniqueness**: indeed. Perfect to me.
2. **Multiple Use**: A wallet may reuse a c_nonce value until the server rejects it due to invalidation or prior use.
3. **Issuance Timing**: A `c_nonce` in a server response should generally be new. However, there are scenarios where a response may not contain a new nonce, such as when the previously issued nonce has not yet been used by the Client.

TakahikoKawasaki:
@peppelinux
Unless I misread the specification or it has changed since I last read it, the specification does not require the one-time use of `c_nonce`.

The OID4VCI specification may require the one-time use of `c_nonce`, but before making that decision, it would be advisable to ask experts why [Section 8.2](https://www.rfc-editor.org/rfc/rfc9449.html#section-8.2), "Providing a New Nonce Value", of [RFC 9449](https://www.rfc-editor.org/rfc/rfc9449.html), "OAuth 2.0 Demonstrating Proof of Possession (DPoP)", specifically states the following:

>It is up to the authorization server when to supply a new nonce value for the client to use. The client is expected to use the existing supplied nonce in DPoP proofs until the server supplies a new nonce value.

pmhsfelix:
@TakahikoKawasaki 
Not sure that I understood the different between *Uniqueness* and *Issuance Timing*. My current understanding from the spec is that:
- A server (AS or Credential Issuer) MAY return the same `c_nonce` on more than one response.
- A client SHOULD (MUST?) use the same `c_nonce` on multiple requests, until a different `c_nonce` is present on a server response.

This requirements allows the use of "stateless" _nonces_, composed by a signed server timestamp. which ensure the _nonce_ usage by the client happened in a given time window.

TakahikoKawasaki:
@pmhsfelix 

### Pseudo server-side code explaining MULTIPLE USE and UNIQUENESS

```java
// Determine the c_nonce value to be embedded in the server response.
String prepare_c_nonce_embedded_in_server_response(String current_c_nonce)
{
    // If the current c_nonce has not expired yet.
    if (is_c_nonce_expired(current_c_nonce) == false)
    {
        // MULTIPLE USE
        // Allow the wallet to continue to use the current c_nonce.
        return current_c_nonce;
    }

    // UNIQUENESS
    // Generate a new c_nonce whose value has never been used previously.
    return generate_new_c_nonce();
}
```

### Pseudo server-side code explaining ONE-TIME USE and UNIQUENESS

```java
// Determine the c_nonce value to be embedded in the server response.
String prepare_c_nonce_embedded_in_server_response(String current_c_nonce, boolean consumed)
{
    // ONE-TIME USE
    // If the current c_nonce has been consumed.
    if (consumed)
    {
        // UNIQUENESS
        // Generate a new c_nonce whose value has never been used previously.
        return generate_new_c_nonce();
    }

    // If the current c_nonce has expired.
    if (is_c_nonce_expired(current_c_nonce))
    {
        // A new c_nonce needs to be issued because the current one has expired.

        // UNIQUENESS
        // Generate a new c_nonce whose value has never been used previously.
        return generate_new_c_nonce();
    }

    // The current c_nonce is valid until it is consumed.
    return current_c_nonce;
}
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -728,7 +728,9 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
...
+The way that data is incorporated depends on the key proof type. In a JWT, for example, the `c_nonce` value is conveyed in the `nonce` claim, whereas the audience is conveyed in the `aud` claim. In a Linked Data proof, for example, the `c_nonce` is included as the `challenge` element in the key proof object and the Credential Issuer (the intended audience) is included as the `domain` element.
```
paulbastian, line 733: While I was reading this section for #293, I was wondering why we have this text in the first place, it seems like a duplication to me, especially if the Proof type section follows a few dozens lines later.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -728,7 +728,9 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
 
 The `proof_type` parameter is an extension point that enables the use of different types of proofs for different cryptographic schemes.
 
-The `proof` element MUST incorporate the Credential Issuer Identifier (audience), and a `c_nonce` value generated by the Authorization Server or the Credential Issuer to allow the Credential Issuer to detect replay. The way that data is incorporated depends on the key proof type. In a JWT, for example, the `c_nonce` value is conveyed in the `nonce` claim, whereas the audience is conveyed in the `aud` claim. In a Linked Data proof, for example, the `c_nonce` is included as the `challenge` element in the key proof object and the Credential Issuer (the intended audience) is included as the `domain` element.
+The `proof` element MUST incorporate the Credential Issuer Identifier (audience), and a `c_nonce` value generated by the Authorization Server or the Credential Issuer to allow the Credential Issuer to detect replay.  The Wallet MUST use this nonce for subsequent requests until a new nonce is issued by the Credential Issuer. Each issued `c_nonce` value MUST be unique, a Credential Issuer cannot issue duplicate `c_nonces` values.
```
jogu, line 731: I think this is the wrong place in the spec to have this text - this section is about credential requests. A requirement for the wallet to use the provided value seems like it should be in the section where the value is provided.

(And I agree with the other comments, I think the 'MUST be unique' is a separate discussion that we should discuss on a issue first so we can have a proper discussion on whether it needs to be unique or not.)

jogu (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
Sakurann: i think the current intent is that lifetime of c_nonce is controlled by the issuer, and it is not necessarily unique...

jogu:
This one has been sitting stalled for a while - to keep the PR list to things we are actively wanting people to look at I'll closed it - please feel free to continue discussion on https://github.com/openid/OpenID4VCI/issues/313 or please do reopen if you think it can be un-stalled.

PR #326, opened 2024-05-24
clarifications about how many credentials can be issued using the same AT

peppelinux:
This PR aims to resolve part of the comments shared in the issue https://github.com/openid/OpenID4VCI/issues/313

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -698,7 +698,7 @@ The Credential Endpoint issues a Credential as approved by the End-User upon pre
 
 Communication with the Credential Endpoint MUST utilize TLS. 
 
-The Client can request issuance of a Credential of a certain type multiple times, e.g., to associate the Credential with different public keys/Decentralized Identifiers (DIDs) or to refresh a certain Credential.
+The Client can request issuance of a Credential of a certain type multiple times, e.g., to associate the Credential with different public keys/Decentralized Identifiers (DIDs) or to refresh a certain Credential. It is at the Credential Issuer's discretion how many copies of a Credential it is willing to issue, how often, and whether reissued Credentials automatically update if the underlying claims have changed.
 
 If the Access Token is valid for requesting issuance of multiple Credentials, it is at the Client's discretion to decide the order in which to request issuance of multiple Credentials requested in the Authorization Request.
 


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -698,7 +698,7 @@ The Credential Endpoint issues a Credential as approved by the End-User upon pre
 
 Communication with the Credential Endpoint MUST utilize TLS. 
 
-The Client can request issuance of a Credential of a certain type multiple times, e.g., to associate the Credential with different public keys/Decentralized Identifiers (DIDs) or to refresh a certain Credential.
+The Client can request issuance of a Credential of a certain type multiple times, e.g., to associate the Credential with different public keys/Decentralized Identifiers (DIDs) or to refresh a certain Credential. It is at the Credential Issuer's discretion how many copies of a Credential it is willing to issue, how often, and whether reissued Credentials automatically update if the underlying claims have changed.
```
Sakurann, line 701: err, PR #293 here on [line 1399](https://github.com/openid/OpenID4VCI/pull/293/files#diff-1f424614b35a9899813079f1b1f6218631a2aedd993368ccb89bb81a9eda0289R1399) introduces batch_size, so there is a way for the issuer to communicate how many credentials it is willing to issue in a batch, so it is a little more than "at the discretion of the issuer".

also I am not sure what "automatically update" means. there is no way for the issuer to push into the wallet an updated credential, the wallet need to talk to the credential endpoint explicitly to get credentials refreshed.

Sakurann (changes requested):
per comments here https://github.com/openid/OpenID4VCI/pull/326/files#r1613666445

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: I would be ok with this

````
The Client can request the issuance of a Credential of a certain type multiple times, for example, to associate the Credential with different public keys or Decentralized Identifiers (DIDs), or to refresh a certain Credential. However, it is important to note that the ability to do so depends on whether the Issuer supports the multiple issuance of the same Credential Type or the refresh of the Credential.
````

paulbastian: I would rather prefer this as a suggestion in PR#293 to avoid the merge conflict.

peppelinux: yes we can do this, see: https://github.com/openid/OpenID4VCI/pull/293#discussion_r1632079784

pmhsfelix: Is this PR still relevant?

jogu:
I believe this was resolved in PR #293 as per https://github.com/openid/OpenID4VCI/pull/293#discussion_r1632079784 - closing.

PR #334, opened 2024-05-27
clarify that tx_code parameter in the offer is not the code but requirements for it

### LINKED ISSUE #301: tx_code describes both the descriptor and the value
I got Implementer feedback twice, that it is confusing, that `tx_code` is the parameter value for both:
- the descriptor of the Transaction Code in the Credential Offer including the formatting, length and description
- the value of the Transaction Code in the Token Request

It seemed to confuse people that parameter names are the same.

Possible solution could be to rename the object in the Credential Offer to `tx_code_description`.

Sakurann:
resolves #301.

In the token request can also add `itself` of something like below, but honestly felt redundant.

> * `tx_code`: OPTIONAL. String value containing a Transaction Code **itself**.


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -352,7 +352,7 @@ The following values are defined by this specification:
...
+  * `tx_code`: OPTIONAL. Object describing requirements for a transaction Code when the Authorization Server expects presentation of a Transaction Code by the End-User along with the Token Request in a Pre-Authorized Code Flow. If the Authorization Server does not expect a Transaction Code, this object is absent; this is the default. The Transaction Code is intended to bind the Pre-Authorized Code to a certain transaction to prevent replay of this code by an attacker that, for example, scanned the QR code while standing behind the legitimate End-User. It is RECOMMENDED to send the Transaction Code via a separate channel. If the Wallet decides to use the Pre-Authorized Code Flow, the Transaction Code value MUST be sent in the `tx_code` parameter with the respective Token Request as defined in (#token-request). If no `length` or `description` is given, this object may be empty, indicating that a Transaction Code is required.
```
awoie, line 355:
```suggestion
  * `tx_code`: OPTIONAL. Object describing requirements for a Transaction Code when the Authorization Server expects presentation of a Transaction Code by the End-User along with the Token Request in a Pre-Authorized Code Flow. If the Authorization Server does not expect a Transaction Code, this object is absent; this is the default. The Transaction Code is intended to bind the Pre-Authorized Code to a certain transaction to prevent replay of this code by an attacker that, for example, scanned the QR code while standing behind the legitimate End-User. It is RECOMMENDED to send the Transaction Code via a separate channel. If the Wallet decides to use the Pre-Authorized Code Flow, the Transaction Code value MUST be sent in the `tx_code` parameter with the respective Token Request as defined in (#token-request). If no `length` or `description` is given, this object may be empty, indicating that a Transaction Code is required.
```

paulbastian:
The proposal was actually about renaming the parameter. You think the cost/benefit ratio is not worth it?

Sakurann:
discussed in the WG, direction is to clarification and breaking change of a parameter name is not justified.

Sakurann:
applied Paul's suggestion

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -352,7 +352,7 @@ The following values are defined by this specification:
...
+  * `tx_code`: OPTIONAL. Object describing requirements for a Transaction Code when the Authorization Server expects presentation of a Transaction Code by the End-User along with the Token Request in a Pre-Authorized Code Flow. If the Authorization Server does not expect a Transaction Code, this object is absent; this is the default. The Transaction Code is intended to bind the Pre-Authorized Code to a certain transaction to prevent replay of this code by an attacker that, for example, scanned the QR code while standing behind the legitimate End-User. It is RECOMMENDED to send the Transaction Code via a separate channel. If the Wallet decides to use the Pre-Authorized Code Flow, the Transaction Code value MUST be sent in the `tx_code` parameter with the respective Token Request as defined in (#token-request). If no `length` or `description` is given, this object may be empty, indicating that a Transaction Code is required.
```
javereec, line 355: Was missing "the", but also this might be more readable imho

```suggestion
  * `tx_code`: OPTIONAL. Object describing the requirements for a Transaction Code, which the Authorization Server expects the End-User to present along with the Token Request in a Pre-Authorized Code Flow. If the Authorization Server does not expect a Transaction Code, this object is absent; this is the default. The Transaction Code is intended to bind the Pre-Authorized Code to a certain transaction to prevent replay of this code by an attacker that, for example, scanned the QR code while standing behind the legitimate End-User. It is RECOMMENDED to send the Transaction Code via a separate channel. If the Wallet decides to use the Pre-Authorized Code Flow, the Transaction Code value MUST be sent in the `tx_code` parameter with the respective Token Request as defined in (#token-request). If no `length` or `description` is given, this object may be empty, indicating that a Transaction Code is required.
```

paulbastian (approved):

c2bo (approved):

javereec (approved):

PR #367, opened 2024-07-23
Fixes #287; Clarify mso_mdoc credential response and add example

### LINKED ISSUE #287: Rework iso mdoc/mdl section
We should do something with the ISO mdoc section of VCI; it's not clear about various things - e.g. what exact CBOR object should be returned in the credential response.

It currently says:

> string that is the base64url-encoded representation of the issued Credential.

Some implementors are using `Document` but ISO 23220 part 3 current draft seems to have `IssuerSignedDehydrated` as the only option in "C.6.9 Credential response" (although `IssuerSigned` is also mentioned as an option in the intro text in "C.5 Credential Issuance Phase").

One option is to mention ISO 23220 part 3 more explicitly and remove all the other text, similar to what is being done in VP ( https://github.com/openid/OpenID4VP/pull/128 )

awoie:
This PR fixes #287 by clarifying the expected credential response for mso_mdoc as well as adding a shortened example. A full example would be too long.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2266,7 +2266,11 @@ The following is a non-normative example of a Credential Request with Credential
 
 ### Credential Response
 
-The value of the `credential` claim in the Credential Response MUST be a string that is the base64url-encoded representation of the issued Credential.
+The value of the `credential` claim in the Credential Response MUST be a string that is the base64url-encoded representation of the CBOR-encoded `IssuerSigned` structure, as defined in [@!ISO.18013-5].
+
+The following is a non-normative example of a Credential Response containing a Credential of format `mso_mdoc`.
+
+<{{examples/credential_response_mso_mdoc.txt}}
 
 ## IETF SD-JWT VC
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2266,7 +2266,7 @@ The following is a non-normative example of a Credential Request with Credential
 
 ### Credential Response
 
-The value of the `credential` claim in the Credential Response MUST be a string that is the base64url-encoded representation of the CBOR-encoded `IssuerSigned` structure, as defined in [@!ISO.18013-5].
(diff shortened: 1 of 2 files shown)


cobward (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2266,7 +2266,11 @@ The following is a non-normative example of a Credential Request with Credential
 
 ### Credential Response
 
-The value of the `credential` claim in the Credential Response MUST be a string that is the base64url-encoded representation of the issued Credential.
+The value of the `credential` claim in the Credential Response MUST be a string that is the base64url-encoded representation of the CBOR-encoded `IssuerSigned` structure, as defined in [@!ISO.18013-5].
```
cobward, line 2269: Do you think we need to say anything about the content of `IssuerSigned`? In the `IssuerSigned` struct, the `nameSpaces` element is optional. IMO it might be worth saying something like:

> This structure (SHOULD/MUST) contain all `Namespaces` and `IssuerSignedItems` that are included in the `AuthorizedNamespaces` of the `MobileSecurityObject`.

jogu (approved):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: i think this makes sense

Sakurann:
```suggestion
The value of the `credential` claim in the Credential Response MUST be a string that is the base64url-encoded representation of the CBOR-encoded `IssuerSigned` structure, as defined in [@!ISO.18013-5]. This structure SHOULD contain all `Namespaces` and `IssuerSignedItems` that are included in the `AuthorizedNamespaces` of the `MobileSecurityObject`.
```

Sakurann (approved):

c2bo (approved):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: @awoie could you please accept the suggestion, so that we can merge a PR?

jogu:
Thanks for closing off the outstanding comments Oliver! Given the only change is the suggestion from Jacob that both you & Kristina were happy with and we have 4 approvals in total I think this is good to merge.

PR #369, opened 2024-07-25
Removes CWT Proof

### LINKED ISSUE #320: Consider removing `cwt` proof type

babisRoutis:
PR removes CWT Proof

Closes #320

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -754,7 +754,7 @@ The `proof` element MUST incorporate the Credential Issuer Identifier (audience)
 
 The initial `c_nonce` value can be returned in a successful Token Response as defined in (#token-response), or in a Credential Error Response as defined in (#issuer-provided-nonce).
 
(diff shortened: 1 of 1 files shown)


Sakurann (approved):

bc-pi (approved):
@Sakurann says it's good and I'm a trusting person

c2bo (approved):
Looks good to me if the people that wanted cwt proofs in vci are happy with removing it

awoie (approved):

jogu:
5 approvals, no unresolved comments, open more than a week - merging! Thanks Babis!

Issue #331, opened 2024-05-27
Is c_nonce required in proof or not?


awoie:
Nonces in proof types are OPTIONAL but the following sentence confused me:
> The proof element MUST incorporate the Credential Issuer Identifier (audience), and a c_nonce value generated by the Authorization Server or the Credential Issuer to allow the Credential Issuer to detect replay.

It says MUST ..., and a `c_nonce`. Does that MUST refer to `c_nonce` as well? Isn't this conflicting with the optionality of nonces in proofs? If it is not conflicting then it is at least confusing.

awoie:
Assuming it is NOT required, I'd propose changing the text to:
> The proof element MUST incorporate the Credential Issuer Identifier (audience), and **if provided** a `c_nonce` value generated by the Authorization Server or the Credential Issuer to allow the Credential Issuer to detect replay.

awoie:
This text makes it probably more clear:
> The proof element MUST incorporate the Credential Issuer Identifier (audience), and a nonce value if a `c_nonce` value was provided by the Authorization Server or the Credential Issuer to allow the Credential Issuer to detect replay.

Sakurann:
I think `c_nonce` is currently meant to be mandatory (either from a token endpoint or from credential endpoint in the error response)...
but I know an implementer for whom adding c_nonce was a hurdle to implement VCI, so they implemented at_hash in the proof instead to prevent replay and achieve the same result.

awoie:
> I think `c_nonce` is currently meant to be mandatory (either from a token endpoint or from credential endpoint in the error response)... but I know an implementer for whom adding c_nonce was a hurdle to implement VCI, so they implemented at_hash in the proof instead to prevent replay and achieve the same result.

Okay but why is `nonce` OPTIONAL in the proof in that case if it was meant to be mandatory? Does it mean that this particular implementer used `at_hash` for `nonce` in the jwt PoP?

awoie:
For reference, from the spec:
> nonce: OPTIONAL (string). The value type of this claim MUST be a string, where the value is a server-provided c_nonce. It MUST be present when the Wallet received a server-provided c_nonce.

awoie:
> For reference, from the spec:
> 
> > nonce: OPTIONAL (string). The value type of this claim MUST be a string, where the value is a server-provided c_nonce. It MUST be present when the Wallet received a server-provided c_nonce.

This suggests that under some circumstances the intention was to allow JWT proofs without a `nonce`.

[...]

Issue #71, opened 2023-09-07
Wallet Attestations and Nonces


tlodderstedt:
Currently, implementations can protect the PoP for wallet attestations (https://datatracker.ietf.org/doc/draft-looker-oauth-attestation-based-client-auth) by:

- making it short lived
- making it one time use
- binding it to a protocol artifact like the authorization code (as kind of a nonce)

The latter does not work for Pushed Authorization Requests. 
The proposal was made, to let the wallet request a nonce from the Issuer that can then be used for binding the wallet attestation. This came up in various discussions including the eIDAS expert group's touch point meeting on OID4VCI and discussions with member states.  

The basic idea would be to define another endpoint, e.g. nonce endpoint, the wallet could be use to obtain that nonce.

peppelinux:
I agree, in the italian solution we have the nonce endpoint as described in the following like, even if the solution uses RFC7523 and not this draft

https://italia.github.io/eudi-wallet-it-docs/versione-corrente/en/wallet-instance-attestation.html#high-level-design

I just to say that the nonce endpoint makes sense

peppelinux:
why not reducing complexity by signing the request (request) with the private key where the public one is attested within the WIA (client_assertion)?

I believe we can satisfy all the requirements without adding anything new and preserving the current implementations

````
POST /as/par HTTP/1.1
Host: pid-provider.example.org
Content-Type: application/x-www-form-urlencoded

response_type=code
&client_id=$thumprint-of-the-jwk-in-the-cnf-wallet-attestation$
&code_challenge=E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM
&code_challenge_method=S256
&request=eyJhbGciOiJSUzI1NiIsImtpZCI6ImsyYmRjIn0.ew0KIC Jpc3MiOiAiczZCaGRSa3F0MyIsDQogImF1ZCI6ICJodHRwczovL3NlcnZlci5leGFtcGxlLmNvbSIsDQo gInJlc3BvbnNlX3R5cGUiOiAiY29kZSBpZF90b2tlbiIsDQogImNsaWVudF9pZCI6ICJzNkJoZFJrcXQz IiwNCiAicmVkaXJlY3RfdXJpIjogImh0dHBzOi8vY2xpZW50LmV4YW1...
&client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-client-attestation
&client_assertion=$WalletInstanceAttestation$
````

[...]

PR #381, opened 2024-08-27
remove `c_nonce` from the token endpoint response

### LINKED ISSUE #39: Remove c_nonce from the token response
### Original Reporter: b_d_c

The option to return c\_nonce with token response should be dropped to simplify the spec and implementations. It’s a lot of complexity and optionality for the potential avoidance of just one HTTP request/response direct from wallet/client to credential issuer.

bc-pi:
remove `c_nonce` and `c_nonce_expires_in` from the token endpoint response to fix #39

again


tlodderstedt (changes requested):
I suggest to add a dedicated nonce endpoint to the credential issuer as replacement for the c_nonce token response parameter. Just removing the c_nonce will force wallets to try the credential issuance request in order to get a nonce in order to be able to create a proper nonce-bound proof of possession. That's complicated.

jogu:
I think this is the previous issue about a nonce endpoint for wallet attestation that people were asking about on today's call:

https://github.com/openid/OpenID4VCI/issues/71

bc-pi:
There has been a lot of discussion about this (and roughly mostly in favor of it) in recent months in https://github.com/openid/OpenID4VCI/issues/39 and also https://github.com/openid/OpenID4VCI/issues/331 and maybe elsewhere too.

peppelinux:
> I suggest to add a dedicated nonce endpoint to the credential issuer as replacement for the c_nonce token response parameter. 

@tlodderstedt that's exactly what I had in mind when I started writing this draft: https://github.com/peppelinux/draft-demarco-oauth-nonce-endpoint?tab=readme-ov-file

bc-pi:
as requested, https://github.com/openid/OpenID4VCI/pull/381/commits/074065b37c9c247ecc2c838ad0be201ad5f0169e adds a dedicated nonce endpoint to the credential issuer as replacement for the c_nonce token response parameter

paulbastian:
Could we please get rid of this [ugly piece](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-7.3.2), then I'm on board!

paulbastian (changes requested):
I would like to mandate nonce endpoint if the issuer requires the use of c_nonce, i.e. the presence of the endpoint enforces its usage, therefore we could get rid of sending fake ugly Credential requests to fetch a c_nonce completely.

Having this endpoint optional and no other indication of the issuer whether he requires c_nonce for proofs seems inefficient.

bc-pi:
> I would like to mandate nonce endpoint if the issuer requires the use of c_nonce

48a367810ef59a559869be3bd74d191d6591b673 updates the proposed text to do that.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -137,7 +138,6 @@ Existing OAuth 2.0 mechanisms are extended as following:
 * A new authorization details [@!RFC9396] type `openid_credential` is defined to convey the details about the Credentials (including Credential Dataset, Credential Formats, and Credential types) the Wallet wants to obtain (see (#authorization-details)).
 * Client metadata is used to convey the Wallet's metadata. The new Client metadata parameter `credential_offer_endpoint` is added to allow a Wallet (acting as OAuth 2.0 client) to publish its Credential Offer Endpoint (see (#client-metadata)).
 * Authorization Endpoint: The additional parameter `issuer_state` is added to convey state in the context of processing an issuer-initiated Credential Offer (see (#credential-authz-request)). Additional parameters `wallet_issuer` and `user_hint` are added to enable the Credential Issuer to request Verifiable Presentations from the calling Wallet during Authorization Request processing.
-* Token Endpoint: Optional response parameters `c_nonce` and `c_nonce_expires_in` are added to the Token Endpoint, and Credential Endpoint to provide the Client with a nonce to be used for proof of possession of key material in a subsequent request to the Credential Endpoint (see (#token-response)).
```
Sakurann, line 140: I think we need to keep the part on credential endpoint extended to support c_nonce
```
* Credential Endpoint: Optional response parameters `c_nonce` and `c_nonce_expires_in` are added to the Credential Endpoint to provide the Client with a nonce to be used for proof of possession of key material in a subsequent request to the Credential Endpoint.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,45 @@ Cache-Control: no-store
 }
 ```
 
+# Nonce Endpoint {#nonce-endpoint}
+
+A Credential Issuer that requires `c_nonce` values to be incorporated into proofs in the Credential Request (see (#credential-request)) MUST offer a Nonce Endpoint. This endpoint allows a Client to acquire a fresh `c_nonce` value without the overhead of a full Credential Request.
```
Sakurann, line 711: improve readability
```suggestion
This endpoint allows a Client to acquire a fresh `c_nonce` value without the overhead of a full Credential Request. A Credential Issuer that requires `c_nonce` values to be incorporated into proofs in the Credential Request (see (#credential-request)) MUST offer a Nonce Endpoint.
```
but also do we really want to mandate nonce endpoint for all issuers that require c_nonce? than why keep an option to return c_nonce in credential response?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,45 @@ Cache-Control: no-store
...
+
+The `nonce_endpoint` Credential Issuer Metadata parameter, as defined in (#credential-issuer-parameters), contains the URL of the Credential Issuer's Nonce Endpoint.
```
```suggestion
The Credential Issuer indicates support for this endpoint using `nonce_endpoint` Credential Issuer Metadata parameter, as defined in (#credential-issuer-parameters), which contains the URL of the Credential Issuer's Nonce Endpoint.
```
I was not sure what is the purpose of the original sentence...

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,45 @@ Cache-Control: no-store
...
+Below is a non-normative example of a Nonce Request:
+
+```
+GET /nonce HTTP/1.1
+Host: credential-issuer.example.com
+```
+
+## Nonce Response {#nonce-response}
+
+The Credential Issuer provides a nonce value in the HTTP response with a 2xx status code and the following parameters included as top-level members in the message body of the HTTP response using the application/json media type:
```
Sakurann, line 729: why not 200?

Sakurann (changes requested):
overall looks good. I almost approved, but I think we should discuss if we really want to mandate nonce endpoint for all issuers that require c_nonce? than why keep an option to return c_nonce in credential response?

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: The credential endpoint is not being extend. It is introduced by this specification.

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: to say where the actual endpoint value can be found

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: was trying to keep the scope of the changes to something that had been agreed on while also accommodating Paul's request for changes

bc-pi: while also keeping my own frustration in check about the fact that these are all things that were only hidden by the nonce from token endpoint but have existed all along so why do they always and only come up in the context of trying to make an improvement

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: Because I was pretty sure someone would put forth the belief that a 201 is more true to Roy Fielding's vision and I can see arguments on both sides but am wholly uninterested in having them while I don't think it actually matters at all and 2xx will let people do whatever.

bc-pi:
> overall looks good. I almost approved, but I think we should discuss if we really want to mandate nonce endpoint for all issuers that require c_nonce? than why keep an option to return c_nonce in credential response?

fair questions that i guess should be discussed in the WG

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,45 @@ Cache-Control: no-store
...
+# Nonce Endpoint {#nonce-endpoint}
+
+This endpoint allows a Client to acquire a fresh `c_nonce` value without the overhead of a full Credential Request. A Credential Issuer that requires `c_nonce` values to be incorporated into proofs in the Credential Request (see (#credential-request)) MUST offer a Nonce Endpoint.
+
+The `nonce_endpoint` Credential Issuer Metadata parameter, as defined in (#credential-issuer-parameters), contains the URL of the Credential Issuer's Nonce Endpoint.
+
+
+## Nonce Request {#nonce-request}
+
+A request for a nonce is made by sending an HTTP GET request to the URL provided in the `nonce_endpoint` Credential Issuer Metadata parameter.
```
paulbastian, line 718: is there any benefit of using GET over POST?

paulbastian (commented):
I like the current direction.

If we keep things as-is, we can completely delete the [chapter 7.3.2](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-7.3.2) as the nonce endpoint is doing the job and all other details are already covered in the main section of Credential Response

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: is there any benefit of using POST over GET?

paulbastian: In OpenID4VP having the request_uri with POST in the first place would have made things easier instead of adding the POST method later

bc-pi: A lot of IMHO very questionable things led to request_uri with POST, which itself is IMHO a very awkward construct. And other than being in the umbrella of OpenID4VP, I don't see what they have in common. 

Using a GET to get something seemed like the most straightforward thing to do.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,45 @@ Cache-Control: no-store
...
+
```
pmhsfelix, line 719: As it is, this endpoint does not uses any authentication or access control. Any HTTP client (that doesn't even need to be an OAuth client) can use it, right?
My concern is the potential for DoS, namely if a GET produces meaningful state changes (e.g. adding entries to a nonces table somewhere).

We could have the endpoint protected by the same access token used to request the credential (I didn't identify any use case where the wallet doesn't have such AT and still needs to access this endpoint).

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: I would argue most calls in OpenID4VCI are POST calls?

bc-pi: > I would argue most calls in OpenID4VCI are POST calls?

Because they send data in the request.

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: "Brian also clarified that the Nonce Endpoint being unprotected follows a previous WG meeting decision." - quoting you paraphrasing me :) from https://lists.openid.net/pipermail/openid-specs-digital-credentials-protocols/Week-of-Mon-20240909/000448.html

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,45 @@ Cache-Control: no-store
...
+* `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)).
+* `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`.
+
+Below is a non-normative example of a Nonce Response:
+
+```
+HTTP/1.1 200 OK
+Content-Type: application/json
+Cache-Control: no-store
```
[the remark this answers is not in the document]
bc-pi: probably, yeah. I'll add something to that effect.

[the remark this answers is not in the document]
bc-pi: see what is now line 734 above that was added with 0e0d18813bb49635747e856751e974e23dfa9750

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: Also removing the ability to have send a c_nonce in the credential response would probably be a breaking change by most measures and it's unclear if the WG has the appetite for that.

bc-pi:
> > overall looks good. I almost approved, but I think we should discuss if we really want to mandate nonce endpoint for all issuers that require c_nonce? than why keep an option to return c_nonce in credential response?
> 
> fair questions that i guess should be discussed in the WG

Noting that removing the ability to have send a c_nonce in the credential response would probably be a breaking change by most measures. Which, I think, would be okay by me. But should have some buy-in from the WG.

openid-4-verifiable-credential-issuance-1_0.md
c2bo: I would agree with GET here as well unless we already have a clear idea that we will need to send data (POST)

paulbastian: Not a hill I want to die on, was just a question coming to my mind. If others think GET is better, I'm fine to close this conversation

bc-pi: Resolving conversation to avoid any deaths on hills.

Sakurann:
Discussed on the WG call, direction on removing section 7.3.2 and consolidating its content in Nonce Endpoint definition.
https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-7.3.2

also in this PR, add a little more explanation that issuer can invalidate c_nonce any time. please open an issue on removing c_nonce_expires_in parameter (@tplooker).

also agreed to open a separate issue on removing an option to return c_nonce from the credential error response, so that we can merge this PR as-is (@bc-pi).

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,47 @@ Cache-Control: no-store
...
+GET /nonce HTTP/1.1
+Host: credential-issuer.example.com
+```
+
+## Nonce Response {#nonce-response}
+
+The Credential Issuer provides a nonce value in the HTTP response with a 2xx status code and the following parameters included as top-level members in the message body of the HTTP response using the application/json media type:
+
+* `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)).
+* `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`.
```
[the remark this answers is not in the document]
bc-pi: how about this? 
```suggestion
* `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`. This value serves only as a hint to the client, indicating how long the Credential Issuer is likely to accept the `c_nonce` as valid.
```

bc-pi:
> Discussed on the WG call, direction on removing section 7.3.2 and consolidating its content in Nonce Endpoint definition. https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-7.3.2

While I was part of that discussion, after looking again at that section and content of this PR, I don't see how that consolation could actually manifest in a meaningful way. I suggest this PR proceed without and consider that as part of the soon to be separate issue on removing an option to return c_nonce from the credential [error] response.

> also in this PR, add a little more explanation that issuer can invalidate c_nonce any time.

https://github.com/openid/OpenID4VCI/pull/381#discussion_r1764090443 has a 

> 
> also agreed to open a separate issue on removing an option to return c_nonce from the credential error response, so that we can merge this PR as-is (@bc-pi).

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
bc-pi: going with the addition of "This value serves only as a hint to the client, indicating how long the Credential Issuer is likely to accept the `c_nonce` as valid."

Sakurann (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1258,6 +1295,7 @@ This specification defines the following Credential Issuer Metadata parameters:
...
 * `credential_endpoint`: REQUIRED. URL of the Credential Issuer's Credential Endpoint, as defined in (#credential-request). This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components.
+* `nonce_endpoint`: OPTIONAL. URL of the Credential Issuer's Nonce Endpoint, as defined in (#nonce-endpoint). This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Credential Issuer does not support the Nonce Endpoint.
```
awoie, line 1298:
```suggestion
* `nonce_endpoint`: OPTIONAL. HTTPS URL of the Credential Issuer's Nonce Endpoint, as defined in (#nonce-endpoint). This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Credential Issuer does not support the Nonce Endpoint.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,47 @@ Cache-Control: no-store
 }
 ```
 
+# Nonce Endpoint {#nonce-endpoint}
+
+This endpoint allows a Client to acquire a fresh `c_nonce` value without the overhead of a full Credential Request. A Credential Issuer that requires `c_nonce` values to be incorporated into proofs in the Credential Request (see (#credential-request)) MUST offer a Nonce Endpoint.
```
pmhsfelix, line 711: What is the meaning of `fresh` in this context? A nonce that is valid (i.e. usable) at the time it was issued? Or does this require two distinct requests to provide two distinct nonces?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,47 @@ Cache-Control: no-store
...
+```
+
+## Nonce Response {#nonce-response}
+
+The Credential Issuer provides a nonce value in the HTTP response with a 2xx status code and the following parameters included as top-level members in the message body of the HTTP response using the application/json media type:
+
+* `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)).
+* `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`. This value serves only as a hint to the client, indicating how long the Credential Issuer is likely to accept the `c_nonce` as valid.
+
+Due to the temporal and contextually sensitive nature of the `c_nonce` value, the Credential Issuer MUST make the response uncacheable by adding a `Cache-Control` header field including the value `no-store`.
```
pmhsfelix, line 734: Perhaps this text could be a bit more lenient, namely to allow cases where the same `c_nonce` can be cached for a small amount of time (via the `max-age` value on the `Cache-Control`).
```suggestion
Due to the temporal and contextually sensitive nature of the `c_nonce` value, the Credential Issuer MUST explicitly define the allowed caching for the response, for instance by adding the `Cache-Control` header with the `no-store` value, if the response must be uncacheable.
```

pmhsfelix (approved):

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: yes, since a nonce is no more than once we assume that each request return an unique and fresh nonce value. I suppose that the term "fresh" made this implicit. I see space to replace it with something more explicit and with few words.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,47 @@ Cache-Control: no-store
...
+```
+GET /nonce HTTP/1.1
+Host: credential-issuer.example.com
+```
+
+## Nonce Response {#nonce-response}
+
+The Credential Issuer provides a nonce value in the HTTP response with a 2xx status code and the following parameters included as top-level members in the message body of the HTTP response using the application/json media type:
+
+* `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)).
```
peppelinux, line 731: I suggest to rename c_nonce to nonce, making the nonce endpoint of general purpose. A nonce is issued by an entity for a specific scope known to its issuer, we don't need to use a particular member name to say that it is a credential nonce, it's a only nonce

nonce and nonce_expires_in would be enough from my perspective, allowing the nonce endpoint to be a general and reusable endpoint within the OAuth framework

openid-4-verifiable-credential-issuance-1_0.md
OR13: I filed https://github.com/openid/OpenID4VCI/issues/395

No reason to delay this PR, but I would like to be able to use this endpoint without needing to pull in all of OIDC4VCI.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: discussed on the WG call
> fresh -> usable or valid at the time of issuance

Sakurann:
on protecting the endpoint, discussed on the WG call, no strong reason to do it, agreed to proceed with unprotected endpoint, but open an issue to gather an implementation experience if anyone wants/needs a stateful c_nonce

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: I don't think that should be accounted for at the HTTP layer of cache.

peppelinux:
> on protecting the endpoint, discussed on the WG call, no strong reason to do it, agreed to proceed with unprotected endpoint, but open an issue to gather an implementation experience if anyone wants/needs a stateful c_nonce

interesting is the idea to use the AT released by the token endpoint, and for the purpose of the credential issuance, to the oauth nonce endpoint as a protected endpoint.

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: A general and reusable endpoint within the OAuth framework was in no way a goal of this work. Not at all. 

And I used c_nonce (a name that I honestly think is terrible for many reasons) to allow for this PR to fit into the broader document more easily.

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: I mean, sure the client will hold onto the thing so it can use it on a subsequent credential request but that's not something that should be conveyed at the HTTP cache level.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,47 @@ Cache-Control: no-store
...
+* `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`. This value serves only as a hint to the client, indicating how long the Credential Issuer is likely to accept the `c_nonce` as valid.
```
peppelinux, line 732:
```suggestion
* `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`. This value serves only as a hint to the Client, indicating how long the Credential Issuer is likely to accept the `c_nonce` as valid.
```

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: Okay, since this Nonce Endpoint is specifically tailored for credential issuance within the sole scope of the OpenID4VCI specs, it would make sense to rename the endpoint metadata parameter from `nonce_endpoint` to `credential_nonce_endpoint`.

In this way, other nonce endpoints created for other OpenID/IETF specs will not collide with this one.
The related section should therefore be renamed to "Credential Nonce Endpoint," adding "Credential" as a prefix to contextualize its scope.

peppelinux (changes requested):
I have requested some small changes about naming, you can consider this PR approved by me

c2bo:
> overall looks good. I almost approved, but I think we should discuss if we really want to mandate nonce endpoint for all issuers that require c_nonce? than why keep an option to return c_nonce in credential response?

I was wondering about this as well. One reason why I think it would make sense to keep the option to return `c_nonce` in the Credential Response would be to allow the Issuer to invalidate an existing nonce because of some kind security concern (some kind of anomaly detection etc. that got triggered). That way the issuer could create a stateful nonce for a specific client via the credential response if necessary.

c2bo (approved):

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: As @jogu points out (thanks @jogu!) there is no problem with collision because this endpoint is at the credential issuer. 

This seems to be in the same vein of what I said about this in slack to you @peppelinux yesterday but bears repenting, "the context is totally different."

bc-pi:
> > overall looks good. I almost approved, but I think we should discuss if we really want to mandate nonce endpoint for all issuers that require c_nonce? than why keep an option to return c_nonce in credential response?
> 
> I was wondering about this as well. One reason why I think it would make sense to keep the option to return `c_nonce` in the Credential Response would be to allow the Issuer to invalidate an existing nonce because of some kind security concern (some kind of anomaly detection etc. that got triggered). That way the issuer could create a stateful nonce for a specific client via the credential response if necessary.

I've created https://github.com/openid/OpenID4VCI/issues/393 to discuss/track "why keep an option to return c_nonce in credential response?"

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -118,6 +118,7 @@ Deferred Credential Issuance:
 This specification defines an API for Credential issuance provided by a Credential Issuer. The API is comprised of the following endpoints:
 
 * A mandatory Credential Endpoint from which Credentials can be issued (see (#credential-endpoint)). From this endpoint, one Credential, or multiple Credentials with the same Credential Dataset can be issued in one request.
+* An optional Nonce Endpoint from which a fresh a `c_nonce` value can be obtained to be used in proof of possession of key material in a subsequent request to the Credential Endpoint (see (#nonce-endpoint)).
```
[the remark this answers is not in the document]
bc-pi: thank you

bc-pi:
> Just to be sure, this PR creates two possible ways for the Issuer to provide the c_nonce, as both are described with MAY, this means the Wallet must implement both. Is this analysis correct?

I don't think so, no.

paulbastian:
> > Just to be sure, this PR creates two possible ways for the Issuer to provide the c_nonce, as both are described with MAY, this means the Wallet must implement both. Is this analysis correct?
> 
> I don't think so, no. 

What do you think then?

bc-pi:
> > > Just to be sure, this PR creates two possible ways for the Issuer to provide the c_nonce, as both are described with MAY, this means the Wallet must implement both. Is this analysis correct?
> > 
> > 
> > I don't think so, no.
> 
> What do you think then?

@jogu's analysis from over in one of the 87 other communication channels seemed, as I said there, constructive: 

> It depends a little on the wallet's viewpoint. In some ways, it replaces a mechanism that may or may not happen (cnonce at token endpoint, cnonce error from credential endpoint) with a "if I list this endpoint in my metadata, you need to call it, and once you call it your credential endpoint call should succeed first time".

But also I think that this PR is a huge step in the right direction that is long overdue. 

And further improvements should be considered seriously per https://github.com/openid/OpenID4VCI/issues/393 soon after this is merged. Which is hopefully soon.

openid-4-verifiable-credential-issuance-1_0.md
nemqe: if we allow max-age I think we just define the `nonce_expires_in` on an HTTP level

openid-4-verifiable-credential-issuance-1_0.md
nemqe, line 711: > without the overhead of a full Credential Request

I guess this refers to the "nonce by error" delivery mechanism, but this comes later in the text so it might be confusing to phrase it like this. Maybe we could drop this part?

```suggestion
This endpoint allows a Client to acquire a fresh `c_nonce` value. A Credential Issuer that requires `c_nonce` values to be incorporated into proofs in the Credential Request (see (#credential-request)) MUST offer a Nonce Endpoint.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -750,9 +787,9 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
...
+The proof(s) in the `proof` or  `proofs` parameter MUST incorporate the Credential Issuer Identifier (audience), and optionally a `c_nonce` value generated by the Credential Issuer to allow the Credential Issuer to detect replay. The way that data is incorporated depends on the key proof type. In a JWT, for example, the `c_nonce` value is conveyed in the `nonce` claim, whereas the audience is conveyed in the `aud` claim. In a Linked Data proof, for example, the `c_nonce` is included as the `challenge` element in the key proof object and the Credential Issuer (the intended audience) is included as the `domain` element.
 
-The initial `c_nonce` value can be returned in a successful Token Response as defined in (#token-response), or in a Credential Error Response as defined in (#issuer-provided-nonce).
+The initial `c_nonce` value can be returned in a Nonce Response as defined in (#nonce-response), or in a Credential Error Response as defined in (#issuer-provided-nonce).
```
nemqe, line 792: a nitpick, I think that the nonce from the nonce endpoint is fetched while the nonce in the error response is returned

nemqe (approved):

paulbastian (approved):

paulbastian:
Re-reading section 7.3.2 I think we should delete the following paragraph:

> 
If the Client has not received a c_nonce and the Credential Issuer Metadata contains proof_types_supported indicating a key proof is required for the requested Credential, the Client MUST send a Credential Request that contains a proof or proofs parameter that is fully valid but does not include a c_nonce value. It is the Credential Issuer policy whether or not a c_nonce value is required in the key proofs.If the Client received a c_nonce, the c_nonce value MUST be incorporated in the respective parameter in the proof or proofs object.

I think leaving this paragraph in would lead to confusion.

babisRoutis:
FWIW, 

I think that the removal of `c_nonce` from Token Endpoint is a big improvement. Now, it is only the credential issuer's responsibility to provide such a  `c_nonce`.

I was hoping, though, to have a single way (for the caller/wallet) to obtain such a `c_nonce`.  Introducing the nonce endpoint and keeping `c_nonce` in both success and/or error response, I have the feeling makes solution more complex that is needed.

For instance, in draft 13, it was clear - although not explicitly stated - that the main way to obtain `c_nonce` was via credential response. Token response was an optimization and in worst case, wallet could just "ignore" the token response in the expense of an additional credential request 

With this PR, unless I am missing something, a wallet should be prepared to handle 
* Issuers that do provide nonce endpoint but don't provide c_nonce in the credential response
* Issuers that do NOT provide nonce endpoint but provide c_nonce in the credential response
* Issuers that support both nonce endpoint & `c_nonce` in the responses.

bc-pi:
> Re-reading section 7.3.2 I think we should delete the following paragraph:
> 
> > If the Client has not received a c_nonce and the Credential Issuer Metadata contains proof_types_supported indicating a key proof is required for the requested Credential, the Client MUST send a Credential Request that contains a proof or proofs parameter that is fully valid but does not include a c_nonce value. It is the Credential Issuer policy whether or not a c_nonce value is required in the key proofs.If the Client received a c_nonce, the c_nonce value MUST be incorporated in the respective parameter in the proof or proofs object.
> 
> I think leaving this paragraph in would lead to confusion, especially given that weird MUST that now contradicts the Nonce endpoint

That that weird MUST needs to change but I'm reluctant to delete the whole paragraph.

paulbastian:
As I would read this PR, it recommends the wallet to use the nonce endpoint if it's present and the parameter in credential response is kind of an alternative/fallback. Therefore this paragraph does not fit any longer.

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: I am unable to see the distinction.

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: it was a nod to the folks who find the "nonce by error" delivery mechanism so deeply objectionable and I'd prefer to keep it until such time as #393 is acted on.

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: or not at any level https://github.com/openid/OpenID4VCI/issues/394

bc-pi: by which i mean, it's not appropriate at the HTTP layer and it's not needed at the application layer so why have it at all?

bc-pi:
aadb2bb8b0a245970dc7d3e9beafe552864eecf7 is what I was able to summon the appetite for

tplooker (approved):

openid-4-verifiable-credential-issuance-1_0.md
peppelinux: bo-ring :-)

peppelinux: I menat IANA metadata registry

https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml

OAuth Authorization Server Metadata

nonce_endpoint might collide over there, yes different contexts (openid != IETF) but if this endpoint is specialized for credential issuance it would be better to give to it a name pointing exactly and esplicitly in this way

bc-pi: that context is still totally different

peppelinux (approved):

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: after a maybe productive but definitely not appropriate side discussion about this - I think we can agree to disagree

jogu: > I menat IANA metadata registry
> 
> https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml
> 
> OAuth Authorization Server Metadata
> 
> nonce_endpoint might collide over there, yes different contexts (openid != IETF) but if this endpoint is specialized for credential issuance it would be better to give to it a name pointing exactly and esplicitly in this way

Just to try to clarify, the nonce endpoint here is part of the issuer metadata. We will not be making an IANA registration for nonce_endpoint in the OAuth Authorization Server Metadata because we're not defining a nonce endpoint for the authorization server. The suggested nonce_endpoint is (technically) a resource server endpoint as that's how we define the credential issuer.

Adding nonce_endpoint here would not prevent the OAuth WG defining a nonce endpoint for authorization servers, nor would it conflict with any such endpoint. It also wouldn't stop the OAuth WG defining a general nonce endpoint for resource servers should they choose to do so, as resource server metadata would also not have a name space that overlaps with credential issuer metadata (i.e. you could have a nonce endpoint defined in `.well-known/oauth-resource server` and a potentially different nonce endpoint defined in `/.well-known/openid-credential-issuer`.

paulbastian (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,47 @@ Cache-Control: no-store
...
+
+
+## Nonce Request {#nonce-request}
+
+A request for a nonce is made by sending an HTTP GET request to the URL provided in the `nonce_endpoint` Credential Issuer Metadata parameter.
+
+Below is a non-normative example of a Nonce Request:
+
+```
+GET /nonce HTTP/1.1
```
tlodderstedt, line 723: Wouldn't a POST request be more appropriate?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1044,7 +1081,7 @@ Cache-Control: no-store
...
 
-If the Client has not received a `c_nonce` and the Credential Issuer Metadata contains `proof_types_supported` indicating a key proof is required for the requested Credential, the Client MUST send a Credential Request that contains a `proof` or `proofs` parameter that is fully valid but does not include a `c_nonce` value. It is the Credential Issuer policy whether or not a `c_nonce` value is required in the key proofs.
+If the Client has not received a `c_nonce` and the Credential Issuer Metadata contains `proof_types_supported` indicating a key proof is required for the requested Credential, the Client can send a Credential Request that contains a `proof` or `proofs` parameter that is fully valid but does not include a `c_nonce` value. It is the Credential Issuer policy whether or not a `c_nonce` value is required in the key proofs.
```
tlodderstedt, line 1084: When is this paragraph apply? I'm asking since the client can always obtain a c_nonce from the nonce endpoint.

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: +1, I tried to push in this direction as well: https://github.com/openid/OpenID4VCI/pull/381#discussion_r1752363400

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: +1, I was trying to push in this direction as well: https://github.com/openid/OpenID4VCI/pull/381#issuecomment-2366932177

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: I didn't realize this was GET. my suggestion has been POST. In german wallet project we have been prototyping with POST and it works well https://bmi.usercontent.opencode.de/eudi-wallet/eidas-2.0-architekturkonzept/flows/PID-IssuerSigned-cloud/#issuer-session-endpoint-at-the-pid-provider

Sakurann (changes requested):
i did not realize this was GET, approved assuming this was POST, sorry about that...

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: As I've tried to convey in comments on this PR and in calls over the course of the last month, I attempted to keep the scope of this change narrow so as to hopefully make it acceptable to the wide range of opinions and oftentimes irregular participation of the WG members. I have clearly failed in my efforts. I believe the text in question here was written recently by @awoie and because it is "Credential Issuer policy whether or not a `c_nonce` value is required in the key proofs" and a "Client can send a Credential Request that contains a `proof` or `proofs` parameter that is fully valid but does not include a `c_nonce` value" it seemed like it applied enough to not delete it.

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: > Wouldn't a POST request be more appropriate? 

No.

> I mean the endpoint is supposed to provide a new/different value with every response and the result should not be cached (in my opinion).

That's why it was made not cacheable in the example https://github.com/openid/OpenID4VCI/pull/381/files#diff-1f424614b35a9899813079f1b1f6218631a2aedd993368ccb89bb81a9eda0289R741 and text https://github.com/openid/OpenID4VCI/pull/381/files#diff-1f424614b35a9899813079f1b1f6218631a2aedd993368ccb89bb81a9eda0289R734

bc-pi: > +1, I tried to push in this direction as well: [#381 (comment)](https://github.com/openid/OpenID4VCI/pull/381#discussion_r1752363400)

My naive feeling that using a GET to get some data is the most straightforward way to get the data remained unchanged by that discussion over the course of the last two weeks.

nemqe: My 2 cents for what they are worth,

If we asked the question "is this call safe (i.e. is it read-only)" I think the answer would be "maybe". Safe calls shouldn't alter the state of the server in any way but without knowing how the server implements the nonce endpoint we would not know the answer to the question.

If we asked the question "is this call idempotent" I think we would also get a "maybe".  Client could do this call just to "fetch a value" which might not assume any intention of a side-effect, and from the perspective of the server it would depend on the implementation.

I do see the point in making it a GET because it is technically just fetching a random value, but there might be an embedded assumption here that this random value is a prerequisite for some other steps, and that getting a nonce be viewed as setting up a state that needs to be validated by the server.

One could argue that it might be safer to make this a `POST` because this would communicate that the call might not be safe or idempotent and that subsequent calls could cause load, burden, and side-effects for the server.

bc-pi: I can see arguments for both as well but was really trying to avoid them entirely with a simple (maybe simplistic) approach.

bc-pi: but avoidance didn't work and so f7b4dc8e816a99f8ab3568be14ea62b4446c822f Removed the HTTP method specificity from the Nonce Request text and change the example from GET to POST

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: One reason why I requested the addition of the nonce endpoint was to avoid this kind of incomplete request just to trigger the provision of a fresh nonce. The wallet know that it needs a nonce, so it can directly request it from the nonce endpoint. 
I would suggest to delete this paragraph. If needed, similar text should be added  

```
If the Client is not in possession of a valid `c_nonce` and the Credential Issuer Metadata contains `proof_types_supported` indicating a key proof is required for the requested Credential, the Client MUST obtain a fresh nonce from the nonce endpoint. 
```

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: After reading the comments and listening to the discussions, I'm ok with GET or POST, but we need to pick one.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -710,6 +706,48 @@ Cache-Control: no-store
...
+# Nonce Endpoint {#nonce-endpoint}
+
+This endpoint allows a Client to acquire a fresh `c_nonce` value without the overhead of a full Credential Request. A Credential Issuer that requires `c_nonce` values to be incorporated into proofs in the Credential Request (see (#credential-request)) MUST offer a Nonce Endpoint.
+
+The `nonce_endpoint` Credential Issuer Metadata parameter, as defined in (#credential-issuer-parameters), contains the URL of the Credential Issuer's Nonce Endpoint.
+
+
+## Nonce Request {#nonce-request}
+
+A request for a nonce is made by sending an HTTP request to the URL provided in the `nonce_endpoint` Credential Issuer Metadata parameter.
```
bc-pi, line 718:
```suggestion
A request for a nonce is made by sending an HTTP POST request to the URL provided in the `nonce_endpoint` Credential Issuer Metadata parameter.
```

OR13 (approved):

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: as discussed on the call - it's POST now https://github.com/openid/OpenID4VCI/pull/381/commits/55c8388480cfa5daee7396954283ef34cb606c66

Sakurann (approved):

danielfett (approved):

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: So how does the wallet know it needs to include a proof?

nemqe: I think that the current text says that the presence of `proof_types_supported` in the metadata indicates that the prof is required in the request

https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-credential-issuer-provided-

jogu: I think I now understand @tlodderstedt's point.

Some of this overlaps with https://github.com/openid/OpenID4VCI/issues/393 but not completely.

How do people feel about this making the use of the nonce_endpoint by the client mandatory if it's present and a nonce is required:

```suggestion
If the Credential Issuer Metadata contains a `nonce_endpoint` and a `proof_types_supported` indicating a key proof is required for the requested Credential, the Client MUST obtain a `c_nonce` value from the `nonce_endpoint` and send a Credential Request that contains a `proof` or `proofs` parameter that includes a `c_nonce` value. It is the Credential Issuer policy whether or not a `c_nonce` value is required in the key proofs.
```

tlodderstedt: If I get it right, the issuer would signal the requirement to add the nonce to the proof by the existence of its nonce endpoint. 
That's ok with me.

jogu: Correct, that is my suggestion. If a issuer requires a nonce for the proofs for some credentials and not others (not sure if this is a likely case) then this may end up with the client provided a nonce unnecessarily, but that doesn't really hurt anything.

David-Chadwick: I support this text. It clarifies how the c_nonce should work

tlodderstedt (changes requested):
Happy to accept if Joseph's proposal is incorporated. 
https://github.com/openid/OpenID4VCI/pull/381/files#r1787382536

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: the situation was supposed to be accounted for by the text "A Credential Issuer that requires `c_nonce` values to be incorporated into proofs in the Credential Request (see (#credential-request)) MUST offer a Nonce Endpoint." above in the Nonce Endpoint {#nonce-endpoint} section while not disturbing the existing optionality and whatever on nonce in the existing text with the nonce-by-error (or success) but I don't know anymore and apparently that was a dumb thing to do

bc-pi: if it'll help move this forward, i'll commit the suggestion

bc-pi:
> Happy to accept if Joseph's proposal is incorporated. https://github.com/openid/OpenID4VCI/pull/381/files#r1787382536

it's done with 9e0d0599fd81bfa8abe6a13be4e5b948f5f4aca5

jogu (approved):

tlodderstedt:
> > Happy to accept if Joseph's proposal is incorporated. https://github.com/openid/OpenID4VCI/pull/381/files#r1787382536
> 
> 
> 
> it's done with 9e0d0599fd81bfa8abe6a13be4e5b948f5f4aca5 

Great, thanks!

tlodderstedt (approved):

jogu:
As discussed on today's WG call - we now have an astounding 12 approvals and no remaining objections/comments - merging! Thank you everyone!

PR #382, opened 2024-08-31
Enable non-breaking extensibility

### LINKED ISSUE #375: Enable non-breaking extensibility
The OAuth and OpenID Connect specifications use a standard pattern to enable extensibility in a way that the use of extensions does not break existing deployments. That pattern is:

- Where a set of values is defined, specify that additional values MAY be defined and used.
- Where a set of values is used, specify that values that are not understood MUST be ignored.

Language incorporating this pattern is currently missing from the specification. We should add it where metadata parameters are defined and also for other parameter sets. Otherwise, we'll be building a brittle specification whose deployments will break by rejecting unexpected values when they are added as their ecosystems evolve.

For examples of the use of this pattern in OpenID and OAuth specifications, see the companion issue https://github.com/openid/OpenID4VP/issues/227.

selfissued:
Fixes #375

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -449,8 +449,11 @@ The request parameter `authorization_details` defined in Section 2 of [@!RFC9396
 * `credential_configuration_id`: REQUIRED when `format` parameter is not present. String specifying a unique identifier of the Credential being described in the `credential_configurations_supported` map in the Credential Issuer Metadata as defined in (#credential-issuer-parameters). The referenced object in the `credential_configurations_supported` map conveys the details, such as the format, for issuance of the requested Credential. This specification defines Credential Format specific Issuer Metadata in (#format-profiles). It MUST NOT be present if `format` parameter is present.
 * `format`: REQUIRED when `credential_configuration_id` parameter is not present. String identifying the format of the Credential the Wallet needs. This Credential Format Identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format-profiles). It MUST NOT be present if `credential_configuration_id` parameter is present.
 
-Additional `authorization_details` data fields MAY be defined and used,
-as described in [@!RFC9396].
+Additional `authorization_details` data fields MAY be defined and used
+when the `type` value is `openid_credential`.
+Note that when using this type, this overrides the language
+"The AS MUST abort processing and respond with an error ``invalid_authorization_details` to the client if [an object in the authorization_details structure] ...
+is an object of known type but containing unknown fields" in [@!RFC9396].
 The Credential Issuer MUST ignore any unrecognized parameters.
 
 The following is a non-normative example of an `authorization_details` object with a `credential_configuration_id`:
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -828,7 +828,7 @@ This specification defines the following proof types:
 * `ldp_vp`: A W3C Verifiable Presentation object signed using the Data Integrity Proof [@VC_Data_Integrity] as defined in [@VC_DATA_2.0] or [@VC_DATA] is used for proof of possession. When a `proof_type` parameter in a `proof` object is set to `ldp_vp`, it MUST also contain an `ldp_vp` parameter that includes a [W3C Verifiable Presentation](https://www.w3.org/TR/vc-data-model-2.0/#presentations-0) defined in (#ldp-vp-proof-type). When a `proofs` object is using a `ldp_vp` proof type, it MUST include an `ldp_vp` parameter with its value being an array of [W3C Verifiable Presentations](https://www.w3.org/TR/vc-data-model-2.0/#presentations-0), where each of these W3C Verifiable Presentation is formed as defined in (#ldp-vp-proof-type).
 
 Additional proof types MAY be defined and used.
-Proof types used MUST be understood by both the Wallet and the Credential Issuer.
+Proof types MUST be recognized and supported by both the Wallet and the Credential Issuer in order to bind an issued Credential to specific cryptographic key material.
 
 #### `jwt` Proof Type {#jwt-proof-type}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -452,7 +452,7 @@ The request parameter `authorization_details` defined in Section 2 of [@!RFC9396
 Additional `authorization_details` data fields MAY be defined and used
 when the `type` value is `openid_credential`.
 Note that when using this type, this overrides the language
-"The AS MUST abort processing and respond with an error ``invalid_authorization_details` to the client if [an object in the authorization_details structure] ...
+"The AS MUST abort processing and respond with an error `invalid_authorization_details` to the client if [an object in the authorization_details structure] ...
 is an object of known type but containing unknown fields" in [@!RFC9396].
 The Credential Issuer MUST ignore any unrecognized parameters.
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -830,7 +830,6 @@ This specification defines the following proof types:
 * `ldp_vp`: A W3C Verifiable Presentation object signed using the Data Integrity Proof [@VC_Data_Integrity] as defined in [@VC_DATA_2.0] or [@VC_DATA] is used for proof of possession. When a `proof_type` parameter in a `proof` object is set to `ldp_vp`, it MUST also contain an `ldp_vp` parameter that includes a [W3C Verifiable Presentation](https://www.w3.org/TR/vc-data-model-2.0/#presentations-0) defined in (#ldp-vp-proof-type). When a `proofs` object is using a `ldp_vp` proof type, it MUST include an `ldp_vp` parameter with its value being an array of [W3C Verifiable Presentations](https://www.w3.org/TR/vc-data-model-2.0/#presentations-0), where each of these W3C Verifiable Presentation is formed as defined in (#ldp-vp-proof-type).
 
 Additional proof types MAY be defined and used.
-Proof types MUST be recognized and supported by both the Wallet and the Credential Issuer in order to bind an issued Credential to specific cryptographic key material.
 
 #### `jwt` Proof Type {#jwt-proof-type}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -451,7 +451,7 @@ The request parameter `authorization_details` defined in Section 2 of [@!RFC9396
 
 Additional `authorization_details` data fields MAY be defined and used,
 as described in [@!RFC9396].
-The Wallet MUST ignore any unrecognized parameters.
+The Credential Issuer MUST ignore any unrecognized parameters.
 
 The following is a non-normative example of an `authorization_details` object with a `credential_configuration_id`:
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -451,7 +451,7 @@ The request parameter `authorization_details` defined in Section 2 of [@!RFC9396
 
 Additional `authorization_details` data fields MAY be defined and used
 when the `type` value is `openid_credential`.
-Note that this effectively defines the type such that it is never considered invalid due to unknown fields.
+Note that this effectively defines an authorization details type that is never considered invalid due to unknown fields.
 
 The following is a non-normative example of an `authorization_details` object with a `credential_configuration_id`:
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1131,6 +1131,9 @@ Authorization: BEARER czZCaGRSa3F0MzpnWDFmQmF0M2JW
 
 The Deferred Credential Response uses the `credential` parameter as defined in (#credential-response).
 
+Additional Deferred Credential Response parameters MAY be defined and used.
+The Wallet MUST ignore any unrecognized parameters.
+
 The Deferred Credential Response MUST be sent using the `application/json` media type.
 
 ## Deferred Credential Error Response {#deferred-credential-error-response}
@@ -1175,7 +1178,7 @@ The Wallet sends an HTTP POST request to the Notification Endpoint with the foll
 * `event_description`: OPTIONAL. Human-readable ASCII [@!USASCII] text providing additional information, used to assist the Credential Issuer developer in understanding the event that occurred. Values for the `event_description` parameter MUST NOT include characters outside the set `%x20-21 / %x23-5B / %x5D-7E`.
 
 Additional Notification Request parameters MAY be defined and used.
-The Wallet MUST ignore any unrecognized parameters.
+The Credential Issuer MUST ignore any unrecognized parameters.
 
 Below is a non-normative example of a Notification Request when a credential was successfully accepted by the End-User:
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1360,7 +1360,7 @@ This specification also defines a new OAuth 2.0 Authorization Server metadata [@
 
 Additional Authorization Server metadata parameters MAY be defined and used,
 as described in [@!RFC8414].
-The wallet MUST ignore any unrecognized parameters.
+The Wallet MUST ignore any unrecognized parameters.
 
 # Security Considerations {#security-considerations}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -449,9 +449,8 @@ The request parameter `authorization_details` defined in Section 2 of [@!RFC9396
 * `credential_configuration_id`: REQUIRED when `format` parameter is not present. String specifying a unique identifier of the Credential being described in the `credential_configurations_supported` map in the Credential Issuer Metadata as defined in (#credential-issuer-parameters). The referenced object in the `credential_configurations_supported` map conveys the details, such as the format, for issuance of the requested Credential. This specification defines Credential Format specific Issuer Metadata in (#format-profiles). It MUST NOT be present if `format` parameter is present.
 * `format`: REQUIRED when `credential_configuration_id` parameter is not present. String identifying the format of the Credential the Wallet needs. This Credential Format Identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format-profiles). It MUST NOT be present if `credential_configuration_id` parameter is present.
 
-Additional `authorization_details` parameters MAY be defined and used,
-as described in [@!RFC9396],
-including when the `type` value is `openid_credential`.
+Additional `authorization_details` data fields MAY be defined and used,
+as described in [@!RFC9396].
 The Wallet MUST ignore any unrecognized parameters.
 
 The following is a non-normative example of an `authorization_details` object with a `credential_configuration_id`:
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -451,10 +451,7 @@ The request parameter `authorization_details` defined in Section 2 of [@!RFC9396
 
 Additional `authorization_details` data fields MAY be defined and used
 when the `type` value is `openid_credential`.
-Note that when using this type, this overrides the language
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -804,6 +827,9 @@ This specification defines the following proof types:
...
 * `ldp_vp`: A W3C Verifiable Presentation object signed using the Data Integrity Proof [@VC_Data_Integrity] as defined in [@VC_DATA_2.0] or [@VC_DATA] is used for proof of possession. When a `proof_type` parameter in a `proof` object is set to `ldp_vp`, it MUST also contain an `ldp_vp` parameter that includes a [W3C Verifiable Presentation](https://www.w3.org/TR/vc-data-model-2.0/#presentations-0) defined in (#ldp-vp-proof-type). When a `proofs` object is using a `ldp_vp` proof type, it MUST include an `ldp_vp` parameter with its value being an array of [W3C Verifiable Presentations](https://www.w3.org/TR/vc-data-model-2.0/#presentations-0), where each of these W3C Verifiable Presentation is formed as defined in (#ldp-vp-proof-type).
 
+Additional proof types MAY be defined and used.
+Proof types used MUST be understood by both the Wallet and the Credential Issuer.
```
nemqe, line 831: Not that I don't agree with this, I am just curious to see why this text is formulated in this way as this one breaks the pattern of "receiver decides".

I ask because a Credential Issuer can contain a wide set of optional `proof_type`s, and I expected a sentence like "Wallet MUST ignore any unrecognized proof types from Credential Issuer Metadata."
"Credential Issuer MUST reject any Credential Request containing an unrecognized proof type."

selfissued: The normal "MUST ignore any unrecognized parameters" language would be wrong here, because it's a security issue if you can't verify the proof.  But this is nonetheless an extension point, so I wanted to be explicit about the rules for handling this kind of extension.

nemqe: Understood. I just wondered because unilateral extension where a Credential Issuer defines a new proof type the spec and offers more to a wallet that is oblivious to the extension in my mind did not fit "MUST be understood by both"

Sakurann: we should discuss this. I agree the MUST here feels different

bc-pi: I also agree with the intent of this MUST text. Maybe expand it a bit more to better describe what using a proof type entails?:

```suggestion
Proof types MUST be recognized and supported by both the Wallet and the Credential Issuer in order to bind an issued Credential to specific cryptographic key material.
```

bc-pi (approved):
one suggestion but otherwise LGTM

paulbastian (commented):
@selfissued Any reason why Deferred Credential Response and Notification Response were omitted?

selfissued:
I enabled additional Deferred Credential Response parameters in my latest commit.

I didn't add Notification Response parameters because the defined response is 204 No Content.  But we could allow for Notification Response parameters if we choose.  What do people think?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -446,6 +449,11 @@ The request parameter `authorization_details` defined in Section 2 of [@!RFC9396
...
 * `format`: REQUIRED when `credential_configuration_id` parameter is not present. String identifying the format of the Credential the Wallet needs. This Credential Format Identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format-profiles). It MUST NOT be present if `credential_configuration_id` parameter is present.
 
+Additional `authorization_details` parameters MAY be defined and used,
+as described in [@!RFC9396],
+including when the `type` value is `openid_credential`.
```
Sakurann, line 454:
```suggestion
Additional `authorization_details` parameters MAY be defined and used,
when the `type` value is `openid_credential`, as described in [@!RFC9396].
```
I don't think we want to allow new `type` values? even if we do, this text should be clearer if we allow new type values or more extensions to type openid_credential

Sakurann (changes requested):
I would really like us to discuss the "both wallet and the issuer must understand the proof type" requirement. it we keep this requirement, i really don't understand why we don't have similar requirements for other issuer metadata that the wallet has to pre-discover from the issuer metadata
we are not enabling extensibility within credential format profiles?

jogu (approved):

openid-4-verifiable-credential-issuance-1_0.md
selfissued: The spec says this, so it's already clear what the type value must be when using it:
* `type`: REQUIRED. String that determines the authorization details type. It MUST be set to `openid_credential` for the purpose of this specification.

Therefore, I believe it was unnecessary for me to say anything about the type value.  I propose to simply delete the qualifier, so the sentence is more parallel to other such sentences in the spec, and simply becomes:
> Additional `authorization_details` parameters MAY be defined and used, as described in [@!RFC9396].

paulbastian (approved):

jogu:
> I would really like us to discuss the "both wallet and the issuer must understand the proof type" requirement. it we keep this requirement, i really don't understand why we don't have similar requirements for other issuer metadata that the wallet has to pre-discover from the issuer metadata we are not enabling extensibility within credential format profiles?

Discussed on today's WG call, there was consensus on keeping the requirement.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -446,6 +449,10 @@ The request parameter `authorization_details` defined in Section 2 of [@!RFC9396
...
 * `format`: REQUIRED when `credential_configuration_id` parameter is not present. String identifying the format of the Credential the Wallet needs. This Credential Format Identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format-profiles). It MUST NOT be present if `credential_configuration_id` parameter is present.
 
+Additional `authorization_details` data fields MAY be defined and used,
+as described in [@!RFC9396].
+The Wallet MUST ignore any unrecognized parameters.
```
pmhsfelix, line 454: Since the `authorization_details` in the authorization request is created by the wallet itself, does it make sense to state that "The Wallet MUST ignore any unrecognized parameters"? Perhaps this sentence should be in the `token-response` section, where the `authorization_details` is generated by the AS.

selfissued:
> Could you please elaborate why? could not find in wg call minutes

As discussed in the call, the reason that both parties must understand the proof is that understanding it is integral to the security of the interaction.  It's not like most extension points where not-understood values can be ignored.  That's why people on the call were good with documenting that this particular extension point doesn't fit the normal must-ignore-if-not-understood pattern.

openid-4-verifiable-credential-issuance-1_0.md
selfissued: @pmhsfelix Thanks.  I corrected the party that must ignore `authorization_details` extensions.

pmhsfelix: Should we also add the equivalent sentence on the `authorization_details` returned in the token response, where now is the wallet that MUST ignore unrecognized parameters put there by the AS?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -446,6 +449,10 @@ The request parameter `authorization_details` defined in Section 2 of [@!RFC9396
...
 * `format`: REQUIRED when `credential_configuration_id` parameter is not present. String identifying the format of the Credential the Wallet needs. This Credential Format Identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format-profiles). It MUST NOT be present if `credential_configuration_id` parameter is present.
 
+Additional `authorization_details` data fields MAY be defined and used,
+as described in [@!RFC9396].
+The Credential Issuer MUST ignore any unrecognized parameters.
```
ju-cu, line 454:
```suggestion
The Credential Issuer MUST ignore any unrecognized data fields.
```

Technically, the key/value pairs in the authorization details object are not parameters...

ju-cu, line 454: I wonder if this extensibility is actually compliant with RAR. Section 5 of RFC 9396 states that:

> The AS MUST abort processing and respond with an error `invalid_authorization_details` to the client if [an object in the authorization_details structure] [...] is an object of known type but containing unknown fields,

Does the proposed extensibility mean that, when it comes to the `openid_credential` type, there are no "unknown fields" because by definition, any unrecognized field gets ignored and consequently all fields are expected and thus "known"?

jogu: Ouch, well spotted, I'd missed that. That's definitely awkward...

selfissued: Thanks for pointing out this conflict, @ju-cu.  I added language to the PR explicitly addressing this conflict.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -446,6 +449,13 @@ The request parameter `authorization_details` defined in Section 2 of [@!RFC9396
...
 * `format`: REQUIRED when `credential_configuration_id` parameter is not present. String identifying the format of the Credential the Wallet needs. This Credential Format Identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format-profiles). It MUST NOT be present if `credential_configuration_id` parameter is present.
 
+Additional `authorization_details` data fields MAY be defined and used
+when the `type` value is `openid_credential`.
+Note that when using this type, this overrides the language
+"The AS MUST abort processing and respond with an error ``invalid_authorization_details` to the client if [an object in the authorization_details structure] ...
```
ju-cu, line 455:
```suggestion
"The AS MUST abort processing and respond with an error `invalid_authorization_details` to the client if [an object in the authorization_details structure] ...
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -655,6 +673,10 @@ In addition to the response parameters defined in [@!RFC6749], the Authorization
   
 Note: The `credential_identifiers` parameter cannot be used when the `scope` parameter is used in the Authorization Request to request issuance of a Credential.
 
+Additional Token Response parameters MAY be defined and used,
+as described in [@!RFC6749].
+The Wallet MUST ignore any unrecognized parameters.
```
ju-cu, line 678: Does that include unknown data fields in an `openid_credential` within the `authorization_details` structure?

RFC 9396, Section 7.1 states that the AS may enrich `authorization_details` in the Token Response. Since we added extensibility for the `openid_credential` type, I think we should be clear on how the Wallet must treat a response with unknown data fields. We can add a normative statement here or under #(authorization-details. @pmhsfelix had the same concern [in another comment](https://github.com/openid/OpenID4VCI/pull/382#discussion_r1757151581) and suggested that the Wallet MUST ignore any unrecognized data fields.

selfissued: I'm not sure what you mean by "unknown data fields" or where you'd like them to be ignored when they occur.  Can you maybe give an example?

FYI, I believe I addressed [@pmhsfelix's comment](https://github.com/openid/OpenID4VCI/pull/382#discussion_r1757151581) in [this commit](https://github.com/openid/OpenID4VCI/pull/382/commits/6659b29032ef359be0f87029366e76f45907420a) in which I corrected the party that must ignore the not-understood parameter.

Sakurann:
> Could you please elaborate why? could not find in wg call minutes
> 
> > As discussed in the call, the reason that both parties must understand the proof is that understanding it is integral to the security of the interaction. It's not like most extension points where not-understood values can be ignored. That's why people on the call were good with documenting that this particular extension point doesn't fit the normal must-ignore-if-not-understood pattern.

I understand the thinking, but I still believe this is not enforceable. You can't stop the wallet from sending the proof type that the issuer does not support (just like for any other issuer metadata). it would be much better to paraphrase this to be an error condition - "issuer MUST return an error if the wallet uses a proof type not supported by the issuer. 

I am sorry, but I am not comfortable merging this PR with this text. please separate this proof related language into an another PR, so that we could merge everything else meanwhile, thank you.

openid-4-verifiable-credential-issuance-1_0.md
pmhsfelix: `authorization_details` can be used in *two* directions:
- On an authorization request, produced by the Wallet and consumed by the AS.
- On a token response, produced by the AS and consumed by the Wallet.
The later usage of `authorization_details` doesn't seem to be covered by the PR, i.e., there isn't any mention to the Wallet ignoring extra fields in the `authorization_details` produced by the AS and present in the *token response*.

Detail: regarding the use of `authorization_details` in the authorization request, should it be the *Credential Issuer* or the *Authorization Server* that must ignore them.

selfissued:
> I am sorry, but I am not comfortable merging this PR with this text. please separate this proof related language into an another PR, so that we could merge everything else meanwhile, thank you.

I deleted the text you're not comfortable with so that we can merge the rest.

openid-4-verifiable-credential-issuance-1_0.md
selfissued: If you have better text in mind @pmhsfelix please make a change suggestion that I can apply to the PR.  Thanks.

ju-cu: @selfissued 
I can try to elaborate more on the issue. Using RAR, an Authorization Server MUST return `authorization_details` in the Token Response (see [RFC 9396 Section 7](https://datatracker.ietf.org/doc/html/rfc9396#name-token-response)). The `authorization_details` in the Token Response are not necessarily the same as in the Authorization or Token Request (see [RFC 9396 Section 7.1](https://datatracker.ietf.org/doc/html/rfc9396#name-enriched-authorization-deta)). Technically, with the extension mechanism for the `openid_credential` type, there could be a discrepancy between the Wallet and the Authorization Server, where the Authorization Server enriches an `openid_credential` type in the `authorization_details` array of a Token Response with extra fields that the Wallet does not understand (RFC 9396 assumes that all possible fields of a type are known upfront, which is not the case here). To avoid any undefined behavior, I suggest, that the Wallet must ignore any unknown fields of an `openid_credential` type in an `authorization_details` array in a Token Response.

ju-cu: To be clear, the "unrecognized parameters" that the Wallet must ignore are not only parameters of the Token Response (as defined in RFC 6749) but also unrecognized data fields of the `openid_credential` type defined in this spec. The former is already defined, and the letter should also be defined imo.

c2bo (approved):

Sakurann (approved):

Sakurann:
@selfissued thank you! let us know when all other comments are addressed so that we can merge this PR.

openid-4-verifiable-credential-issuance-1_0.md
pmhsfelix, line 678:
```suggestion
The Wallet MUST ignore any unrecognized parameters in the Token Response.
An included `authorization_details` parameter MAY also have additional data fields defined and used
when the `type` value is `openid_credential`.
The Wallet MUST ignore any unrecognized data fields in the `authorization_details` present in the Token Response.

File: openid-4-verifiable-credential-issuance-1_0.md (hunk 4 above)
Comment (Sakurann) [reply openid-4-verifiable-credential-issuance-1_0.md:457]: why do we want to overwrite RAR language here though? I think we should rather stick to RAR and remove 
the following sentence:

> Additional `authorization_details` data fields MAY be defined and used
> when the `type` value is `openid_credential`.

File: openid-4-verifiable-credential-issuance-1_0.md [hunk 7]
```diff
@@ -446,6 +449,13 @@ The request parameter `authorization_details` defined in Section 2 of [@!RFC9396
...
 * `format`: REQUIRED when `credential_configuration_id` parameter is not present. String identifying the format of the Credential the Wallet needs. This Credential Format Identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format-profiles). It MUST NOT be present if `credential_configuration_id` parameter is present.
 
+Additional `authorization_details` data fields MAY be defined and used
+when the `type` value is `openid_credential`.
+Note that when using this type, this overrides the language
+"The AS MUST abort processing and respond with an error `invalid_authorization_details` to the client if [an object in the authorization_details structure] ...
+is an object of known type but containing unknown fields" in [@!RFC9396].
+The Credential Issuer MUST ignore any unrecognized parameters.
```
Sakurann, line 457:
```suggestion
```

Sakurann (changes requested):
RAR language is awkward and needs to be fixed

openid-4-verifiable-credential-issuance-1_0.md
selfissued: We're deviating from RAR for this type because that enables non-breaking extensibility when using the type `openid_credential`.  Without this language, if RAR if followed strictly, if a new `authorization_details` parameter is used, it would cause an error in existing deployments.  Meaning that this type would not be extensible.

peppelinux (approved):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: @jogu's proposal is to define a claim for this authorization_details type as an alternative extensibility point

Sakurann: @tlodderstedt WDYT?

bc-pi: I don't think anything is being overwritten. If the openid_credential type is defined such that it can have additional data fields, then by definition it can have additional data fields. I think this can probably be noted with language that's less likely to be objectionable.

openid-4-verifiable-credential-issuance-1_0.md
bc-pi, line 457:
```suggestion
Note that this effectively defines the type such that it is never considered invalid due to unknown fields.
```

selfissued: Brian's suggested change superseded the one above.

openid-4-verifiable-credential-issuance-1_0.md
selfissued: This text has been superseded by other suggested changes.

bc-pi (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -526,6 +533,10 @@ This specification defines the following request parameters that can be supplied
 
 Note: When processing the Authorization Request, the Credential Issuer MUST take into account that the `issuer_state` is not guaranteed to originate from this Credential Issuer in all circumstances. It could have been injected by an attacker.
 
+Additional Authorization Request parameters MAY be defined and used,
```
tlodderstedt, line 536: do we really need to state this again?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -605,6 +616,10 @@ If the Token Request contains a scope value related to Credential issuance and t
 
 When the Pre-Authorized Grant Type is used, it is RECOMMENDED that the Credential Issuer issues an Access Token valid only for the Credentials indicated in the Credential Offer (see (#credential-offer)). The Wallet SHOULD obtain a separate Access Token if it wants to request issuance of any Credentials that were not included in the Credential Offer, but were discoverable from the Credential Issuer's `credential_configurations_supported` metadata parameter.
 
+Additional Token Request parameters MAY be defined and used,
```
tlodderstedt, line 619: do we really need to state this again?

tlodderstedt (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -446,6 +449,10 @@ The request parameter `authorization_details` defined in Section 2 of [@!RFC9396
...
 * `format`: REQUIRED when `credential_configuration_id` parameter is not present. String identifying the format of the Credential the Wallet needs. This Credential Format Identifier determines further claims in the authorization details object needed to identify the Credential type in the requested format. This specification defines Credential Format Profiles in (#format-profiles). It MUST NOT be present if `credential_configuration_id` parameter is present.
 
+Additional `authorization_details` data fields MAY be defined and used
+when the `type` value is `openid_credential`.
+Note that this effectively defines the type such that it is never considered invalid due to unknown fields.
```
Sakurann, line 454:
```suggestion
Note that this effectively defines an authorization details type that is never considered invalid due to unknown fields.
```

Sakurann (approved):
approve with an editorial suggestion

nemqe (approved):

openid-4-verifiable-credential-issuance-1_0.md
jogu: Definitely not technically necessary, but I think doing so will help implementers that aren't that familiar with OAuth2.

jogu:
I believe all comments have been responded to and/or resolved now (if I missed any feel free to open issues) and we have 8 approvals so merging!

PR #391, opened 2024-09-13
Add extensibility to Credential Response

### LINKED ISSUE #58: Issuance of transaction_ids
### Original Reporter: danielfett

The following points are not clearly defined in the spec right now:

\(A\) In a batch \(deferred\) issuance, will there be one transaction\_id for all deferred credentials or will it be the same for all?

\(B\) How is the transaction\_id or are the transaction\_ids returned? There is the following example in the spec:

```
HTTP/1.1 200 OK
Content-Type: application/json
Cache-Control: no-store

{
   "credential_responses":[
      {
         "transaction_id":"8xLOxBtZp8"
      },
      {
         "format":"jwt_vc_json",
         "credential":"YXNkZnNhZGZkamZqZGFza23....29tZTIzMjMyMzIzMjMy"
      }
   ],
   "c_nonce":"fGFF7UkhLa",
   "c_nonce_expires_in":86400
} 
```

This shows one transaction\_id per credential; in case there are many deferred credentials, will it be the same for all credentials \(repeated, say, 100 times for 100 credentials\) or will there be different ones?

\(C\) How is the wallet supposed to hit the deferred credential endpoint and what will be returned from the endpoint? See also [https://bitbucket.org/openid/connect/issues/1759/no-need-for-credential-endpoint-when-we](https://bitbucket.org/openid/connect/issues/1759/no-need-for-credential-endpoint-when-we)

### LINKED ISSUE #386: Add extensibility to Credential Response
It seems there could be benefit to have the ability to add auxilliary/associated/additional data/metadata to an issued credential. This could apply to a single credential but specifically also to each instance of Credential if multiple are issued using batch issuance with proof*s* and credential*s*.

This may be used for example key handles of ARKG or other advanced key derivation schemes, like described in #359 where each credential is using a public key that is derived by the issuer, so there is a need to pass a keyhandle for each credential instance. This could also be used for other features in the future, where there is data associated to a credential instance useful to the Wallet that should not be included into the credential itself.

I see some options:

## Option 1:
Adding a `credentials_metadata` array in non-breaking way that includes all potential associated data, assuming same ordering as `credentials` array.
```
{
  "credentials": [
    "LUpixVCWJk0eOt4CXQe1NXK....WZwmhmn9OQp6YxX0a2L",
    "YXNkZnNhZGZkamZqZGFza23....29tZTIzMjMyMzIzMjMy"
  ],
  "credentials_metadata" :  [
    { 
      "keyhandle" : "..." //keyhandle for credential1,
      "some-other-metadata" : ".."
    },
    { 
      "keyhandle" : "..." //keyhandle for credential2,
      "some-other-metadata" : ".."
    }
  ]
}
```

## Option 2:
Enhance `credentials` array in breaking way that includes all potential associated data.
```
{
"credentials": [
    { 
      "credential" : "LUpixVCWJk0eOt4CXQe1NXK....WZwmhmn9OQp6YxX0a2L",
      "keyhandle" : "..." //keyhandle,
      "some-other-metadata" : ".."
    },
    { 
      "credential" : "LUpixVCWJk0eOt4CXQe1NXK....WZwmhmn9OQp6YxX0a2L",
      "keyhandle" : "..." //keyhandle,
      "some-other-metadata" : ".."
    }
  ]
}
```

## Option 3:
Each feature that uses/requires associated data introduces its own parameter array. This means though there are proof type specific parameters.
```
{
  "credentials": [
    "LUpixVCWJk0eOt4CXQe1NXK....WZwmhmn9OQp6YxX0a2L",
    "YXNkZnNhZGZkamZqZGFza23....29tZTIzMjMyMzIzMjMy"
  ],
  "keyhandles" :  [
    //keyhandle for credential1,
    //keyhandle for credential2
  ]
}
```

c2bo:
Option 2 as introduced by @paulbastian and discussed in #386: 
- Credential Response always returns an array
- This array holds an object with the credential and additional (optional) metadata
- notification_id is also moved into this new object
- deferred Endpoint re-uses this array for its response

Closes #386
Closes #58


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2312,6 +2322,10 @@ The technology described in this specification was made available from contribut
 
    [[ To be removed from the final specification ]]
 
+   -15
+
+   * credential response always returns an array when not returning a transaction_i with the option for additional meta-data
```
babisRoutis, line 2327: - There is a typo `transaction_i` instead of `transaction_id`
- I would add the following: 
  - `notification_id` moved to the `credentials` element 
  - Deferred Response returns `credentials` array, as well

c2bo: Agreed, should be fixed with my last commit

babisRoutis: Thank you Christian

(That's the first section to check whenever I read the spec :smile: )

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -946,12 +946,12 @@ If the Credential Response is not encrypted, the media type of the response MUST
...
-* `transaction_id`: OPTIONAL. String identifying a Deferred Issuance transaction. This claim is contained in the response if the Credential Issuer cannot immediately issue the Credential. The value is subsequently used to obtain the respective Credential with the Deferred Credential Endpoint (see (#deferred-credential-issuance)). It MUST not be used  if `credential` or `credentials` is present. It MUST be invalidated after the Credential for which it was meant has been obtained by the Wallet.
+* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. The following parameters are used inside this object (other parameters MAY also be included):
+   * `credential`: REQUIRED. Contains one issued Credential. It MAY be a string or an object, depending on the Credential Format. See Appendix A for the Credential Format-specific encoding requirements.
+   * `notification_id`: OPTIONAL. String identifying an issued Credential that the Wallet includes in the Notification Request as defined in (#notification).
```
paulbastian, line 951: Shouldn't we have a single `notification_id` for the whole response?

paulbastian (approved):
I'm not entirely sure about `notification_id`, but apart from that, looks good to me! We also figured out that there may be cases around revocation that may benefit from this in the future.

openid-4-verifiable-credential-issuance-1_0.md
jogu: I've been struggling with the same question for a few days:

A `notification_id` for each credential in the batch is cleaner/simpler, as we don't need to worry about partial states (i.e. the question of what notification to make if you receive 10 credentials and only manage to store 5 of them).

However it's kindof sucky to make 10 notification http calls for a batch of 10 credentials.

I'm probably veering towards cleaner/simpler and ignoring the 'kind of sucky' part.

jogu (approved):

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: We may change the notification request to an array as well?

jogu: I think that could make a lot of sense.

babisRoutis: FWIW, I think that `notification_id` should point to a single credential not in a batch.

For instance, an Issuer can react to a notification event carrying a `credential_deleted` or `credential_failure` to invalidate the credential from a status list.

We can avoid multiple posts for notifications by using an array of `notification_id` per possible event.

```json
{
  "credential_accepted" : [ "notId#1"],
  "credential_failure" : ["notid#4"],
  "credential_deteled": [ ]
}
```

Something like a summary notification

babisRoutis (approved):

c2bo:
How should we proceed with the notification endpoint? If we want to change that one to expect an array, should this happen in this PR or should we create another issue/PR?

paulbastian:
I would say it makes sense in this PR, as this one also changes that the behavior of notification_id in the credential response

openid-4-verifiable-credential-issuance-1_0.md
awoie: I think it might be useful to move it in because of `credential_deleted` which is specific to one issued credential although when I think about it, `notification_id` overlaps a bit with the concept of credential version IDs or data set IDs but this is out of scope of this PR.

awoie (approved):
I'm ok with the direction of this PR. I believe it is useful that `notification_id` is moved in too. Although I believe that `notification_id` should be potentially renamed in the future to credential version ID or credential data set ID.

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: Why would a wallet delete a single instance rather than the whole batch or the same Credential Dataset?

awoie (changes requested):
When I think about this another time, I also that a single notification id would be more logical When I approved, I thought it would also allow for different data sets. If the use case is really just different keys, not data sets such as math-exam vs science-exam, I think it should be a top-level notification id instead of individual notification ids.

c2bo:
discussed at  DCP WG call on 17/09:
People seem to prefer `notification_id` as a top level parameter and convey only the overall issuance process (batch or not)
- notification endpoint can remain unchanged
- we need a bit more language on how to deal with partial failures
I'll adjust the PR accordingly and mark it as ready for review

babisRoutis:
> discussed on the DCP WG call (17/09): People seem to prefer `notification_id` as a top level parameter and convey only the overall issuance process (batch or not)
> 
>     * notification endpoint can remain unchanged
> 
>     * we need a bit more language on how to deal with partial failures
>       I'll adjust the PR accordingly and mark it as ready for review

The `notification_id` represents a notification (event) that can be send from the wallet.

Currently (d13,d14), such a notification event is related to a single - issued - credential, hence the event types names: `credential_accepted`, `credential_failed`, `credential_deleted`.

Should we change `notification_id` to be applicable for the `credentials` array, I think we should also change the names of the events and their description, to be consistent.

For instance ,  `credentials_accepted`, `credentials_failed`, `credentials_deleted`. 

BTW, I don't understand very well what's the difference between failed & deleted credentials.

babisRoutis (changes requested):
Please consider keeping consistent the granularity of notification (event) names with the `notification_id` scope.

Keeping `notification_id` as a top-level item of credential response, means that a notification is related to the whole array of `credentials`. 

Perhaps the names `credential_accepted`, `credential_failed` and `credential_deleted` should be in plural.

babisRoutis:
> discussed on the DCP WG call (17/09): People seem to prefer `notification_id` as a top level parameter and convey only the overall issuance process (batch or not)

It is not very clear (to me) what the rules for this top-level attributes would be.
I guess there are two options:
a) `notification_id` can be provided only if there is the `credentials` array
b) `notification_id` can be provided in any case (either when `credentials` or `transaction_id` is present.)

I think (a) is more clear because you get the `notification_id` together with your credentials both in the simple and deferred cases.

(Sorry for the multiple comments)

jogu:
@babisRoutis 

>  For instance ,  credentials_accepted, credentials_failed, credentials_deleted.
>
> BTW, I don't understand very well what's the difference between failed & deleted credentials.

From memory, "failed" is "a technical failure somewhere, e.g. wallet thinks the credential is invalid" and "deleted" is more along the lines of an operation explicitly performed by the user (so e.g. the wallet shows the user the issued credential, and the user decides they don't want to actually store it). If this needs to be clarified in the spec I'd suggest a new issue is opened for that.

c2bo:
I've tried to add all of the discussed changes. I agree that the name of the events is not a 100% fit now, but am not sure if we should change them at this state.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1136,10 +1164,10 @@ Communication with the Notification Endpoint MUST utilize TLS.
...
+The Wallet sends an HTTP POST request to the Notification Endpoint with the following parameters in the entity-body and using the `application/json` media type. If the Wallet supports the Notification Endpoint, the Wallet MAY send one or more Notification Requests per Credential issuance flow.
+* `notification_id`: REQUIRED. String received in the Credential Response or Deferred Credential Response identifying an issuance flow that contained one or more Credentials with the same Credential Configuration and Credential Dataset.
+* `event`: REQUIRED. Type of the notification event. It MUST be a case sensitive string whose value is either `credential_accepted`, `credential_failure`, or `credential_deleted`. `credential_accepted` is to be used when the Credentials were successfully stored in the Wallet, with or without user action. `credential_deleted` is to be used when the unsuccessful Credential issuance was caused by a user action. In all other unsuccessful cases, `credential_failure` is to be used. Partial errors in an Credential issuance flow (e.g., one of the Credentials could not be stored) SHOULD be treated as the overall issuance flow failing.
```
babisRoutis, line 1169: I believe that either we should drop the `credential_` prefix or change it to `credentials_`. 

Perhaps, though, this could be addressed in a separate issue/PR.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2312,6 +2340,11 @@ The technology described in this specification was made available from contribut
 
    [[ To be removed from the final specification ]]
 
+   -15
+
+   * credential response always returns an array when not returning a transaction_id with the option for additional meta-data
+   * deferred credential response always returns an array (same as credential response)
+
```
babisRoutis, line 2347: Until d14, `notification_id` was associated to a single credential.

I think that readers/implementers would appreciate to highlight the change on the granularity of `notification_id` (and notification events)

babisRoutis (approved):

openid-4-verifiable-credential-issuance-1_0.md
c2bo: That is a good point, I will add an explicit point to the document history

openid-4-verifiable-credential-issuance-1_0.md
Sakurann:
```suggestion
* `event`: REQUIRED. Type of the notification event. It MUST be a case sensitive string whose value is either `credential_accepted`, `credential_failure`, or `credential_deleted`. `credential_accepted` is to be used when the Credentials were successfully stored in the Wallet, with or without user action. `credential_deleted` is to be used when the unsuccessful Credential issuance was caused by a user action. In all other unsuccessful cases, `credential_failure` is to be used. Partial errors during batch credential issuance (e.g., one of the Credentials in the batch could not be stored) MUST be treated as the overall issuance flow failing.
```

i am really not sure what "an Credential issuance flow" is. I think what we mean here is `Partial errors during batch credential issuance`?
also, I think it should be a MUST. SHOULD here is not really helpful.
i am pretty neutral on `credential_` vs `credentials_`. it's a batch of credentials, but of the same credential type

examples/credential_response_ldp_vc.txt
```diff
@@ -3,34 +3,38 @@ Content-Type: application/json
...
+        },
+        "proof": {
+          "type": "Ed25519Signature2020",
+          "created": "2022-02-25T14:58:43Z",
+          "verificationMethod": "https://example.edu/issuers/565049#key-1",
+          "proofPurpose": "assertionMethod",
+          "proofValue": "zeEdUoM7m9cY8ZyTpey83yBKeBcmcvbyrEQzJ19rD2UXArU2U1
+                         jPGoEtrRvGYppdiK37GU4NBeoPakxpWhAvsVSt"
+        }
       }
```
Sakurann, line 35:
```suggestion
    {
        "@context": [
          "https://www.w3.org/2018/credentials/v1",
          "https://www.w3.org/2018/credentials/examples/v1"
        ],
        "id": "http://example.edu/credentials/3732",
        "type": [
          "VerifiableCredential",
          "UniversityDegreeCredential"
        ],
        "issuer": "https://example.edu/issuers/565049",
        "issuanceDate": "2010-01-01T00:00:00Z",
        "credentialSubject": {
          "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
          "degree": {
            "type": "BachelorDegree",
            "name": "Bachelor of Science and Arts"
          }
        },
        "proof": {
          "type": "Ed25519Signature2020",
          "created": "2022-02-25T14:58:43Z",
          "verificationMethod": "https://example.edu/issuers/565049#key-1",
          "proofPurpose": "assertionMethod",
          "proofValue": "zeEdUoM7m9cY8ZyTpey83yBKeBcmcvbyrEQzJ19rD2UXArU2U1
                         jPGoEtrRvGYppdiK37GU4NBeoPakxpWhAvsVSt"
      }
```
why do we need another `credential` parameter?

Sakurann:
I must say credentials.credential looks weird.... 

```
{
  "credentials": [
    {  "credential": "omppc3N1ZXJBdXRohEOhASahG...ArQwggKwMIICVqADAgEC"   }
  ]
}
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -946,12 +946,12 @@ If the Credential Response is not encrypted, the media type of the response MUST
...
 * `c_nonce`: OPTIONAL. String containing a nonce to be used to create a proof of possession of key material when requesting a Credential (see (#credential-request)). When received, the Wallet MUST use this nonce value for its subsequent Credential Requests until the Credential Issuer provides a fresh nonce.
 * `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`.
-* `notification_id`: OPTIONAL. String identifying an issued Credential that the Wallet includes in the Notification Request as defined in (#notification). This parameter MUST NOT be present if `credential` parameter is not present.
+* `notification_id`: OPTIONAL. String identifying an issuance flow in which one or more Credentials were issued. It MUST be included in the Notification Request as defined in (#notification). It MUST not be used if the `credentials` parameter is not present.
```
Sakurann, line 954:
```suggestion
* `notification_id`: OPTIONAL. String identifying a Credential Request in which one or more Credentials were issued. It MUST be included in the Notification Request as defined in (#notification). It MUST not be used if the `credentials` parameter is not present.
```

can we please not use a term "issuance flow"? it could mean so many things.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -967,25 +967,34 @@ Content-Type: application/json
...
+    }
+  ],
   "c_nonce": "fGFF7UkhLa",
-  "c_nonce_expires_in": 86400  
+  "c_nonce_expires_in": 86400
 }
 ```
 
-Below is a non-normative example of a Credential Response in an immediate issuance flow for multiple Credential instances in JWT VC format (JSON encoded):
+Below is a non-normative example of a Credential Response in an immediate issuance flow for multiple Credential instances in JWT VC format (JSON encoded) with the additional `notification_id` parameter:
```
Sakurann, line 980:
```suggestion
Below is a non-normative example of a Credential Response in an immediate issuance flow for multiple Credential instances in JWT VC format (JSON encoded) with an additional `notification_id` parameter:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -946,12 +946,12 @@ If the Credential Response is not encrypted, the media type of the response MUST
...
-* `credentials`: OPTIONAL. Contains an array of issued Credentials. It MUST NOT be used if `credential` or `transaction_id` parameter is present. The values in the array MAY be a string or an object, depending on the Credential Format. See (#format-profiles) for the Credential Format-specific encoding requirements.
-* `transaction_id`: OPTIONAL. String identifying a Deferred Issuance transaction. This claim is contained in the response if the Credential Issuer cannot immediately issue the Credential. The value is subsequently used to obtain the respective Credential with the Deferred Credential Endpoint (see (#deferred-credential-issuance)). It MUST not be used  if `credential` or `credentials` is present. It MUST be invalidated after the Credential for which it was meant has been obtained by the Wallet.
+* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. The following parameters are used inside this object (other parameters MAY also be included):
```
Sakurann, line 949:
```suggestion
* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. This specification defines the following parameters to be used inside this object:
```

I think PR #382 should cover the MAY part `Additional Credential Response parameters MAY be defined and used.`

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1124,7 +1152,7 @@ Cache-Control: no-store
 
 # Notification Endpoint {#notification-endpoint}
 
-This endpoint is used by the Wallet to notify the Credential Issuer of certain events for issued Credentials. These events enable the Credential Issuer to take subsequent actions after issuance. The Credential Issuer needs to return one or more `notification_id` parameters in the Credential Response for the Wallet to be able to use this endpoint. Support for this endpoint is OPTIONAL. The Issuer cannot assume that a notification will be sent for every issued Credential since the use of this Endpoint is not mandatory for the Wallet.
+This endpoint is used by the Wallet to notify the Credential Issuer of certain events for issued Credentials. These events enable the Credential Issuer to take subsequent actions after issuance. The Credential Issuer needs to return a `notification_id` parameter in the Credential Response or Deferred Credential Response for the Wallet to be able to use this endpoint. Support for this endpoint is OPTIONAL. The Issuer cannot assume that a notification will be sent for every issued Credential since the use of this Endpoint is not mandatory for the Wallet.
```
Sakurann, line 1155:
```suggestion
This endpoint is used by the Wallet to notify the Credential Issuer of certain events for issued Credentials. These events enable the Credential Issuer to take subsequent actions after issuance. The Credential Issuer needs to return one `notification_id` parameter per Credential Response or Deferred Credential Response for the Wallet to be able to use this endpoint. Support for this endpoint is OPTIONAL. The Issuer cannot assume that a notification will be sent for every issued Credential since the use of this Endpoint is not mandatory for the Wallet.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1097,10 +1106,29 @@ Authorization: BEARER czZCaGRSa3F0MzpnWDFmQmF0M2JW
 
 ## Deferred Credential Response {#deferred-credential-response}
 
-The Deferred Credential Response uses the `credential` parameter as defined in (#credential-response).
+The Deferred Credential Response uses the `credentials` and `notification_id` parameters as defined in (#credential-response).
```
Sakurann, line 1109:
```suggestion
The Deferred Credential Response uses the `credentials` parameter as defined in (#credential-response).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1097,10 +1106,29 @@ Authorization: BEARER czZCaGRSa3F0MzpnWDFmQmF0M2JW
...
+{
+  "credentials": [
+    {
+      "credential": "LUpixVCWJk0eOt4CXQe1NXK....WZwmhmn9OQp6YxX0a2L"
+    },
+    {
+      "credential": "YXNkZnNhZGZkamZqZGFza23....29tZTIzMjMyMzIzMjMy"
+    }
+  ],
+  "notification_id": "3fwe98js"
```
Sakurann, line 1128:
```suggestion
  ]
```

Sakurann (changes requested):
not sure why this PR modifies deferred endpoint to accept notification_id? was it meant to say transaction_id?

also would like to see some descriptions to be more precise

pmhsfelix (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1136,10 +1164,10 @@ Communication with the Notification Endpoint MUST utilize TLS.
 
 ## Notification Request {#notification}
 
-The Wallet sends an HTTP POST request to the Notification Endpoint with the following parameters in the entity-body and using the `application/json` media type. If the Wallet supports the Notification Endpoint, the Wallet MAY send one or more Notification Requests per Credential issued.
+The Wallet sends an HTTP POST request to the Notification Endpoint with the following parameters in the entity-body and using the `application/json` media type. If the Wallet supports the Notification Endpoint, the Wallet MAY send one or more Notification Requests per Credential issuance flow.
```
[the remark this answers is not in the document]
babisRoutis: I think the "notification_id` value received" is indeed better. 

I wallet may call multiple times the credential endpoint, using the same `access_token`. Thus, it can receive more than one `notification_id`.

c2bo:
> not sure why this PR modifies deferred endpoint to accept notification_id? was it meant to say transaction_id?

It is modifying the Deferred Credential Response not the Request. The idea would be that `notification_id` is always returned when the wallet receives credentials.

In the case of deferred issuance that means that the credential endpoint returns a `transaction_id` and the deferred credential endpoint then returns `credentials` and `notification_id`

openid-4-verifiable-credential-issuance-1_0.md
c2bo: My thought was that it makes sense to state it here explicitly because this is a nested object, not a direct Credential Response parameter and we are already expecting more parameters to exist here for things like key derivation. Would you be fine with keeping it like this?
```suggestion
* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. This specification defines the following parameters to be used inside this object: (other parameters MAY also be included):
```

peppelinux (approved):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: maybe because the user has used that one instance? but not sure that information needs to be communicated to the issuer real-time, since that would allow the issuer to get more information about when credential is used

paulbastian (approved):

Sakurann (approved):

Sakurann:
good to merge once @awoie approves

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1136,10 +1166,10 @@ Communication with the Notification Endpoint MUST utilize TLS.
...
+The Wallet sends an HTTP POST request to the Notification Endpoint with the following parameters in the entity-body and using the `application/json` media type. If the Wallet supports the Notification Endpoint, the Wallet MAY send one or more Notification Requests per `notification_id` value received.
+* `notification_id`: REQUIRED. String received in the Credential Response or Deferred Credential Response identifying an issuance flow that contained one or more Credentials with the same Credential Configuration and Credential Dataset.
+* `event`: REQUIRED. Type of the notification event. It MUST be a case sensitive string whose value is either `credential_accepted`, `credential_failure`, or `credential_deleted`. `credential_accepted` is to be used when the Credentials were successfully stored in the Wallet, with or without user action. `credential_deleted` is to be used when the unsuccessful Credential issuance was caused by a user action. In all other unsuccessful cases, `credential_failure` is to be used. Partial errors during batch credential issuance (e.g., one of the Credentials in the batch could not be stored) MUST be treated as the overall issuance flow failing.
```
nemqe, line 1171: if the response is always a list containing a `credentials` key, does it make sense to rename these event to use the plural form like `credentials_accepted`...

nemqe (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -936,22 +936,24 @@ These checks may be performed in any order.
...
+* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. This specification defines the following parameters to be used inside this object:
+   * `credential`: REQUIRED. Contains one issued Credential. It MAY be a string or an object, depending on the Credential Format. See Appendix A for the Credential Format-specific encoding requirements.
+* `transaction_id`: OPTIONAL. String identifying a Deferred Issuance transaction. This claim is contained in the response if the Credential Issuer cannot immediately issue the Credential. The value is subsequently used to obtain the respective Credential with the Deferred Credential Endpoint (see (#deferred-credential-issuance)). It MUST not be used if the `credentials` parameter is present. It MUST be invalidated after the Credential for which it was meant has been obtained by the Wallet.
```
bc-pi, line 953: Not everything in JSON is a claim. Most things, in fact, are not. 

```suggestion
* `transaction_id`: OPTIONAL. String identifying a Deferred Issuance transaction. This parameter is contained in the response if the Credential Issuer cannot immediately issue the Credential. The value is subsequently used to obtain the respective Credential with the Deferred Credential Endpoint (see (#deferred-credential-issuance)). It MUST not be used if the `credentials` parameter is present. It MUST be invalidated after the Credential for which it was meant has been obtained by the Wallet.
```

awoie (approved):

bc-pi (approved):

Issue #294, opened 2024-04-15
Potential improvements for the big picture of issuance flows


paulbastian:
I've tried to summarize all issuance options regarding the use of:
- authorization_details / scope
- credential_identifiers

![OpenID4VCI_v3](https://github.com/openid/OpenID4VCI/assets/16320000/1be235d7-4aa9-4c1c-bac2-faa7969f3e0e)

It seems that there are many possible combinations to do issuance with OpenID4VCI, in my opinion it may be too many and we should consider consolidation.

My observations:
- `scope` is an easy/uncomplicated way to do most use cases :+1:
  - no support for different Credential Datasets but thats ok
  - it is weird that Credential Request uses `format` +format specific parameters instead of `scope`  :thinking: 
- authorization_details enables the Wallet to choose between different Credential Datasets :+1:
  - authorization_details has two options: `credential_configuration_id`  vs `format` + format specific parameter :thinking:
  - when authorization_details is used `credential_identifiers` is still optional :thinking:
  - this means that Credential Request depends on the Token Response :-1:
- in preAuth Code Flow, we have a third option, providing `credential_configuration_ids`, but probably fine
  - its unclear which Credential the Wallet is requesting if multiple IDs are provided
  - `credential_identifiers` cannot be used, because it is not authorization_details flow :-1:
  - Credential Datasets cannot be requested :-1:

Proposals:
1: make `credential_identifiers` mandatory
2: enable `credential_identifiers` for Pre-Auth Code Flow
3: enable `authorization_details` with `credential_configuration_id` in Token Request for Pre-Auth Code Flow
4: consider using `scope` in Credential Request

babisRoutis:
> Proposals: 
1: make `credential_identifiers` mandatory 
2: enable `credential_identifiers` for Pre-Auth Code Flow 
3: enable `authorization_details` with `credential_configuration_id` in Token Request for Pre-Auth Code Flow 
4: consider using `scope` in Credential Request

2,3. This would bring in par (functional-wise) the two flows, I guess. Totally, agree.
4. I think that there isn't one-to-one relationship between a scope and a credential (That is,there could multiple `credential_configuration_ids` that use the same scope)

In general, I feel that VCI has too many alternatives that make it complex without a good reason.
As you pointed out, 
- why have two ways of assembling an `authorization_details`?  Some claim backward compatibility reasons, but I find that this not so strong argument given the extended changes introduced to d13 compared to d12.
[...]

Issue #234, opened 2024-01-25
Make Issuer metadata normative


paulbastian:
As I stated in the WG Call, I believe that `credential_configuration_id` is the better choice. In general, I have trouble understanding how OpenID4VCI works well without metadata, as the Wallet needs to known:

- the credential endpoint
- supported proof types
- supported credentials and formats
- display data for the issuer and the credential

Therefore I believe in productive environment it will be very common that the Issuer has the ability to host metadata.

_Originally posted by @paulbastian in https://github.com/openid/OpenID4VCI/pull/219#pullrequestreview-1833193401_
            
As stated in #219 I believe that Credential Issuer metadata should be mandatory, it just doesn't make sense to me with the current specification otherwise.

jogu:
As mentioned in https://github.com/openid/OpenID4VCI/pull/219#issuecomment-1902864880 I think there are two separate but related questions:

1. Are credential issuers required to have metadata?
2. Does credential issuer metadata need to list all the credentials the issuer supports?

People were reluctant to agree to '2', and there's precdent for this in OAuth, e.g.  in https://datatracker.ietf.org/doc/html/rfc8414#section-2 the AS isn't required to list every supported scope in scopes_supported.)

Sakurann:
OAuth being framework, I am hesitant to require all issuers to have metadata. I think think discussion in issue #82 made it clear that some implementers want to have out of band discovery of the issuer metadata, and I don't see any reason why we should prohibit that.

Sakurann:
Can this be closed in favor of #392, too? do we want a small PR clarifying that issuers do not have to list all credentials in the issuer metadata (which kind of implies the wallet has ways to obtain those credential configurations using means other than issuer metadata)?

Sakurann:
closing in a week unless objections. another option can be to move this issue to HAIP. @paulbastian let us know the preference

paulbastian:
It seems reasonable to me to have this requirement in HAIP

PR #392, opened 2024-09-16
add an option to use credential_configuration_id in credential request

### LINKED ISSUE #132: using credential_configuration_id in the credential request when scopes are used in the authorization request
Currently, Credential Request has two options to specify which credential is being requested: `credential_identifier` or combination of `format` and credential format specific `credential_definition`. 

We (Microsoft) are currently implementing VCI and we use `scopes` and cannot return `credential_identifier` from the token endpoint (because of the limitations of a big AS we are using), but we found that the performance (especially for the wallet) is better when identifier in the `credential_configurations_supported` issuer metadata is used in the credential request instead of format/type combination: it saves from comparing the combination of format and type (that is mapped to an identifier anyway) and helps prevent errors. (I think this is also what @pmhsfelix suggested at IIW)

The flow would look like:
- the wallet receives credential offer with an identifier. 
- the wallet uses that identifier to look up an entry in `credential_configurations_supported` issuer metadata 
- the wallet uses a scope value in the authorization request 
- the wallet uses the same identifier in the credential request

The parameter name should probably be `credential_configuration_id` which is what is used in RAR. I don't think `credential_identifier` should be reused because that one has a distinct meaning and purpose.

Credential Request will look pretty simple:

>* credential_condifuration_id: REQUIRED. String that identifies a Credential that is being requested to be issued. When credential_identifier was returned from the Token Response, it MUST match the value of this string. Otherwise, it MUST match the value of an identifier in the `credentials_supported` issuer metadata that the wallet used to look credential being requested within the issuer metadata.
> * proof: OPTIONAL. Object containing the proof of possession of the cryptographic key material the issued Credential would be bound to. The proof object MUST contain a following claim:
>
>   * proof_type: REQUIRED. String denoting the key proof type. The value of this claim determines other claims in the key proof object and its respective processing rules. Key proof types defined in this specification can be found in [Section 7.2.1](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#proof_types).
>  + encryption related parameters

(mandating to return `credential_identifier`  from the token response is not an option because some ASs cannot make breaking changes and cannot return `credential_identifiers` in the Token Endpoint.)

### LINKED ISSUE #175: Wallet can't always identity whether key proof is required
https://github.com/openid/OpenID4VCI/pull/87 added a mechanism within `credentials_configurations_supported` for the wallet to know if the issuer requires proof of possession for a particular credential configuration, and if so what types are supported.

There are some cases where the wallet can't lookup this value though / it isn't clear to the credential issuer which credential the wallet is asking for.

An oversimplified example that shows the problem, e.g. in the pre-authorised code flow where the wallet has been offered more than one credential:

**credential_configurations_supported**
```
{
  "credential_1": {
    "format": "my_format",
    "claims": { "family_name": {} }
  },
  "credential_2": {
    "format": "my_format",
    "claims": { "family_name": {} },
    "proof_types": [ "jwt" ]
  }
}
```

**credential offer**
```
{
  "credential_issuer": "...",
  "credential_configurations": [ "credential_1", "credential_2" ],
  "grants": { ... }
}
```
**credential request**
```
{
  "format": "my_format",
  "claims": { "family_name": {} }
}
```

This could be solved by having the wallet supply `credential_configuration_id ` in the credential request. (Hence making `format` unnecessary.) i.e.:

**credential request**
```
{
  "credential_configuration_id": "credential_2",
  "claims": { "family_name": {} }
}
```

(Spotted by Taka@Authlete)

### LINKED ISSUE #342: Problems in Authorization Code with scope and mutliple Credential Datasets
Imagine a basic scenario where a Wallet wants to request multiple Credentials with different Credential Datasets, e.g. multiple ePrescriptions. The challenge is how a Wallet knows how many Credential Requests to do.

Within Authorization Code Flow with `authorization_details`, the Wallet can deduct the number of Credential Datasets from the Token Response with `credential_identifiers` array size.

Within Authorization Code Flow with `scope`, this method does not work (as it is specific to authorization_details) and the Wallet may need to send requests until it gets an error, which seems to be insufficient.

Proposal could be to allow `credential_identifiers` for all flows. Changes discussed in #294 are also relevant.

Sakurann:
Closes one remaining item from PR #294 as agreed on[ May 29th 2024 call](https://github.com/openid/OpenID4VCI/issues/132#issuecomment-2137991476).
Closes #132
Closes #175 

enables to use credential_configuration_id in the credential request.

I am increasingly in favor of dropping an option to use format + type in authorization request and credential request, and instead clarify that implementations that do not use issuer metadata should define the mapping between scope (optionally), format, type and credential_configuration_id out of band? #234 is the related issue though it proposes a different solution. (probably needs discussion with torsten on the call)

also resolves #342


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -748,6 +749,8 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
     * `alg`: REQUIRED. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses.
     * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
 
+`credential_configuration_id` or `format` parameters are used when the Credential issuance was requested using a `scope` value in the Authorization Request or a pre-authorisation code was used that did not return an Authorization Details.
```
jogu, line 752: I very much appreciate the use of the British spelling of authorisation 😁 - but I suspect "authorized" was intended:

```suggestion
The `credential_configuration_id` or `format` parameters are used when the Credential issuance was requested using a `scope` value in the Authorization Request or a Pre-Authorized Code was used that did not return an Authorization Details.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -738,8 +738,9 @@ For cryptographic binding, the Client has the following options defined in (#cre
...
-* `format`: REQUIRED if an Authorization Details of type `openid_credential` was not returned from the Token Response (e.g. when the credential was requested using a `scope` value in the authorization request or a pre-authorisation code was used that did not return an Authorization Details). It MUST NOT be used otherwise. A string that determines the format of the Credential to be issued, which may determine the type and any other information related to the Credential to be issued. Credential Format Profiles consist of the Credential format specific parameters that are defined in (#format-profiles). When this parameter is used, the `credential_identifier` Credential Request parameter MUST NOT be present.
+* `credential_identifier`: REQUIRED when an Authorization Details of type `openid_credential` was returned from the Token Response. It MUST NOT be used otherwise. A string that identifies a Credential Dataset that is requested for issuance. When this parameter is used, the `credential_configuration_id` parameter nor `format` parameter with any other Credential format specific parameters such as those defined in (#format-profiles) MUST NOT be present.
```
jogu, line 741: This might read slightly better, not sure:

```suggestion
* `credential_identifier`: REQUIRED when an Authorization Details of type `openid_credential` was returned from the Token Response. It MUST NOT be used otherwise. A string that identifies a Credential Dataset that is requested for issuance. When this parameter is used, the `credential_configuration_id` and `format` parameters, as well as any other Credential format specific parameters such as those defined in (#format-profiles), MUST NOT be present.
```

I think we could potentially drop the 'as well as any other Credential format specific parameters such as those' part to simplify the language here, as they already say "can only be used when format parameter is present".

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -738,8 +738,9 @@ For cryptographic binding, the Client has the following options defined in (#cre
...
+* `credential_configuration_id`: REQUIRED if an Authorization Details of type `openid_credential` was not returned from the Token Response or if `format` parameter is not present. It MUST NOT be used otherwise. A string that identifies one of the keys in the name/value pairs stored in the `credential_configurations_supported` Credential Issuer metadata. The value of this parameter MUST match with one of those in the Credential Issuer metadata. When this parameter is used, the `credential_identifier` parameter nor `format` parameter with any other Credential format specific parameters such as those defined in (#format-profiles) MUST NOT be present.
```
jogu, line 742: Same tweak as above:

```suggestion
* `credential_configuration_id`: REQUIRED if an Authorization Details of type `openid_credential` was not returned from the Token Response or if `format` parameter is not present. It MUST NOT be used otherwise. A string that identifies one of the keys in the name/value pairs stored in the `credential_configurations_supported` Credential Issuer metadata. The value of this parameter MUST match with one of those in the Credential Issuer metadata. When this parameter is used, the `credential_identifier` and `format` parameters, as well as any other Credential format specific parameters such as those defined in (#format-profiles), MUST NOT be present.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -738,8 +738,9 @@ For cryptographic binding, the Client has the following options defined in (#cre
...
+* `format`: REQUIRED if an Authorization Details of type `openid_credential` was not returned from the Token Response or if `credential_configuration_id`. It MUST NOT be used otherwise. A string that determines the format of the Credential to be issued, which may determine the type and any other information related to the Credential to be issued. Credential Format Profiles consist of the Credential format specific parameters that are defined in (#format-profiles). When this parameter is used, the `credential_identifier` nor `credential_configuration_id` parameters MUST NOT be present.
```
jogu, line 743: Similar tweak:

```suggestion
* `format`: REQUIRED if an Authorization Details of type `openid_credential` was not returned from the Token Response or if `credential_configuration_id`. It MUST NOT be used otherwise. A string that determines the format of the Credential to be issued, which may determine the type and any other information related to the Credential to be issued. Credential Format Profiles consist of the Credential format specific parameters that are defined in (#format-profiles). When this parameter is used, the `credential_identifier` and `credential_configuration_id` parameters MUST NOT be present.
```

jogu (approved):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: i considered that, but given that I have received a lot of questions from the implementers not knowing there are credential format profiles, i am keen to keep it..

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: i also use british spelling for `diving licence` :D

paulbastian (changes requested):
My understanding from the discussion around #294 was that `credential_configuration_id`option replaces the `format` option. I would rather replace an option than adding a third one. Apart from them, everything seems fine to me.

paulbastian:
This is missing an entry for the document history

Sakurann:
> My understanding from the discussion around https://github.com/openid/OpenID4VCI/issues/294 was that credential_configuration_idoption replaces the format option. I would rather replace an option than adding a third one.

I don't think we can do it without removing format option in authorization request

Sakurann:
discussed in the WG call, ask the ML if implementers using format + type would be against removing that option in favor of using credential_configuration_id only. and that does not necessarily mean that implementations that do not use issuer metadata are not allowed because we could say that implementations that do not use issuer metadata should define the mapping between scope (optionally), format, type and credential_configuration_id out of band.

@pmhsfelix pointed out that after ID-1 we have already made a change that prohibited using format + type with RAR.

jogu:
> discussed in the WG call, ask the ML if implementers using format + type would be against removing that option in favor of using credential_configuration_id only.

Email sent to mailing list requesting feedback within the next week: https://lists.openid.net/pipermail/openid-specs-digital-credentials-protocols/Week-of-Mon-20240923/000455.html

babisRoutis:
I think that

- `credential_configuration_id` should replace `format` + format specific attributes, in the credential request.
- Keep in the `authorization_details` only the `credential_configuration_id` and remove the `format` (in https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-using-authorization-details)

paulbastian:
> I think that
> 
> - `credential_configuration_id` should replace `format` + format specific attributes, in the credential request.
> - Keep in the `authorization_details` only the `credential_configuration_id` and remove the `format` (in https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-using-authorization-details)
> 

I couldn't agree more

awoie:
Wasn't the `format` option demonstrated in the POTENTIAL interop event and therefore there are existing implementations that have support for this (also including our own implementation)?

awoie:
> > I think that
> > 
> > * `credential_configuration_id` should replace `format` + format specific attributes, in the credential request.
> > * Keep in the `authorization_details` only the `credential_configuration_id` and remove the `format` (in https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-using-authorization-details)
> 
> I couldn't agree more

I would suggest creating a particular issue for removing the `format` option and keep it separate from this PR.

babisRoutis:
> Wasn't the `format` option demonstrated in the POTENTIAL interop event and therefore there are existing implementations that have support for this (also including our own implementation)?

Good morning, @awoie 

For every feature/option removed there was an implementation. 
- There was support for CWT proofs. 
- There were implementations of Batch Endpoint 
- There is support for obtaining `c_nonce` from token endpoint etc. 

The `credential_configuration_id` is a simpler & format-agnostic way to uniquely identify a credential configuration compared to `format`. 
Perhaps, the only feature that could be added to the former would be the option to use format-specific `claims` attribute to control the attributes issued (this is the case for `authorization_details`, if I am not mistaken).

Sakurann:
@babisRoutis @paulbastian i am repeating myself, but what you are suggesting needs to be done in conjunction with the changes to the format+type option in the authorization request and text explaining how the protocol works without issuer metadata, for example by saying that `implementations that do not use issuer metadata should define the mapping between scope (optionally), format, type and credential_configuration_id out of band`.

paulbastian:
> @babisRoutis @paulbastian i am repeating myself, but what you are suggesting needs to be done in conjunction with the changes to the format+type option in the authorization request and text explaining how the protocol works without issuer metadata, for example by saying that `implementations that do not use issuer metadata should define the mapping between scope (optionally), format, type and credential_configuration_id out of band`.

Looks like a good suggestion to me. Anything else that would be missing apart from clarifications that issuer metadata could be communicated out of band?

Sakurann:
Talked to @tlodderstedt, who said that it is better to close this PR and keep things as-is, because...
1. an option to use one `credential_configuration_id` parameter might be easier to the developers than using combination of `format` and `type` parameters, but it is probably making it harder for the ecosystem(s)/use-cases because they will have to define `credential_configuration_id` for `format` and `type` combinations, which might be easy if there is only one issuer, but suddenly becomes harder when there is more than one ecosystem/use-case, which is pretty common.
2. ... and `credential_configuration_id` is a construct in VCI while format and type are credential format construct, so always mandating a mapping to exist between `credential_configuration_id` and `format` and `type` combinations mixes these things up.

babisRoutis:
> Talked to @tlodderstedt, who said that it is better to close this PR and keep things as-is, because...
> 
>     1. an option to use one `credential_configuration_id` parameter might be easier to the developers than using combination of `format` and `type` parameters, but it is probably making it harder for the ecosystem(s)/use-cases because they will have to define `credential_configuration_id` for `format` and `type` combinations, which might be easy if there is only one issuer, but suddenly becomes harder when there is more than one ecosystem/use-case, which is pretty common.
> 
>     2. ... and `credential_configuration_id` is a construct in VCI while format and type are credential format construct, so always mandating a mapping to exist between `credential_configuration_id` and `format` and `type` combinations mixes these things up.

I don't agree on closing the PR.

VCI is all about allowing wallet to place a credential request. It doesn't seem consistent that throughout the issuance process (offer, `authorization_details`, auth. req) an unambiguous identifier can be used, except the most important step.

If there isn't a consensus to remove format + type from credential request, at least , we should have the alternative option to use `credential_configuration_id`.

tlodderstedt:
@babisRoutis 
> 
> VCI is all about allowing wallet to place a credential request. It doesn't seem consistent that throughout the issuance process (offer, `authorization_details`, auth. req) an unambiguous identifier can be used, except the most important step.

What do you expect the wallet to know, when it starts the process? I would expect the wallet to know format and type of the credential the holder is looking for. That's what is being standardized not a credential configuration id. Just as an example, ISO 18013-05 defines the mDL doc type. It does not define how the credential configuration id of an OID4VCI issuer supporting this doc types is. As this is a most likely locally defined metadata identifier. And it is protocol specific. 

So how to cope with this situation? 
1) the wallet could determine the issuer specific credential configuration id(s) for a certain format/type and use the respective credential configuration id in the authorization request. 
2) the wallet sends the format/type in the authorization request, the issuer provides the wallet with its local reference (either through a credential configuration and/or a credential identifier) in the token response to be used in subsequent steps. 
3) The credential configuration id was defined somewhere (in addition to credential types and schemas), and is used by all issuers supporting the same credential type
4) A wallet only works with one issuer for a certain type and is pre-configured or hard coded to use its credential configuration id.

What would you prefer and why?

babisRoutis:
> What do you expect the wallet to know, when it starts the process? I would expect the wallet to know format and type of the credential the holder is looking for

Hi @tlodderstedt 

I think option 1 is the way to go.

For an issuance process, that is initiated at issuer's site, wallet's participation begins with a credential offer  (No format/type)

On the other hand, for a wallet-initiated issuance, wallet should know (or could search, to a local or trusted registry of issuers, for):
-  credential issuer id 
-  format
-  type 

Of course, holder is interested to get a specific credential type, from a trusted issuer. Few would care or even understand about formats. 

With these three  values, wallet can "assemble" on its own a stateless credential offer as follows:
- Obtain issuer metadata, then
- Determine the credential configuration id that for  the type and format the holder is interested in.
- Assemble a credential offer that uses authorization code grant 

So, actually, I believe that wallet has to have always a credential offer (issuer provided or locally created) to interact with the issuer. 

Format/type can be used as criteria to just query issuer's metadata. With this problem solved, I find no reason to use format/type in any of the subsequent steps (authorization, authorization_details & credential request), should this PR be accepted.

Sakurann:
Is credential configuration identifier specific to each issuer or specific to a use-case and is determined by multiple issuers? 
If it is the former, the wallet has to look up every time it is talking to a new issuer, which makes wallet-initiated flows harder. If it is the latter what if a credential of the same format/type combination is used in multiple use-cases and use-case 1 wants to use credential_configuration_id A and use-case 2 wants to use credential_configuration_id B for that same credential format&type? 
I honestly think we don't know yet, and I do see the risk of mandating credential_configuration_id making wallet-initiated flows using scopes harder for the ecosystems.

tlodderstedt:
In the most general case, every issuer has its own credential configuration ids. Implementation experience will tell whether there are situations with shared credential configuration definitions.

babisRoutis:
Dear @Sakurann 

> Is credential configuration identifier specific to each issuer or specific to a use-case and is determined by multiple issuers? 

I think it is clearly the former. Every issuer is free to define its own values for scope & credential configuration id for a given credential. 

> If it is the former, the wallet has to look up every time it is talking to a new issuer, which makes wallet-initiated flows harder. 
If it is the latter what if a credential of the same format/type combination is used in multiple use-cases and use-case 1 wants to use credential_configuration_id A and use-case 2 wants to use credential_configuration_id B for that same credential format&type? 

We can agree, I guess, that the wallet has to lookup credential issuer's metadata, in any case:

- For a wallet initiated flow and/or while handling a credential offer
- For a "known" or a new Issuer
- Using format/type (as we currently do) or credential configuration id (if PR makes it to the spec)

So, if wallet cannot avoid grabbing the metadata, then it can easily determine the configuration for the format/type the holder is looking for, by querying the metadata.

> I honestly think we don't know yet, and I do see the risk of mandating credential_configuration_id making wallet-initiated flows using scopes harder for the ecosystems.

Format/Type have a business value for (or recognized by) the holder and the verifier.
In VP this explains why they both appear to queries. Verifier expresses his query in terms of format/type/attributes.

With regards to VCI,  we already have format/type to our metadata. 
I feel that we don't have to "leak" those business values to every step/call. Their place is in the metadata :smile: 

Moving towards 1.0, perhaps, it would be more reasonable to use credential configuration id instead of format/type, and if there is a need/demand add the format/type afterwards.

jogu:
It looks like we have 3 options on the table and I'm unclear if we have consensus around any of them. I'll add them as separate comments, can people please thumbs up options they like, ignore options they can live with, and thumbs down options they hate?

jogu:
Option 1. We close this PR and keep the spec as is, for scopes you would need to use format+type in the credential request

jogu:
Option 3. We remove the format+type option now

awoie:
Can we change the title of the PR if we discuss removing certain functionality?

Sakurann:
discussed in a WG:
- when scopes were used in the authorization response, 
- if AS is capable of returning credential_identifiers parameter in the token response, wallet MUST use that in the credential request.
- if AS is NOT capable of returning credential_identifiers parameter in the token response, wallet MUST use credential_configuration_id in the credential request.

jruizaranguren:
> Option 3. We remove the format+type option in this PR before merging it

If I understand correctly, this would also remove the format-specific attributes such as `vct`  in the credential request, right?

For example, in the [sd-jwt-vc profile](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-credential-request-6), the `vct` attribute is required when the `format` option is used. This `vct` value must also be specified in the `credential_configuration_supported` metadata and must match the `credential_configuration_id`.

This creates considerable complexity just to use `credential_configuration_id`.

Hence, we may have two kind of flows:

1. those using out-of-band correlation to the specific credentials to be issued, up to the AS (`scopes`,   `access_token`, etc.).
2. those using `authorization_details` and `credential_configuration_id`. `credential_configuration_id`  values might be retrieved from `credential offers` , `issuer meta-data` or `out-of-band` means.

Aside from potentially breaking some implementations (ours included), in what scenarios does only maintaining `credential_configuration_id` and `credential_identifiers`  become problematic? 

@jogu, this continues the chat at last WG.

paulbastian:
I applied the changes discussed in hybrid call to this PR. Please check @Sakurann @jogu @babisRoutis @tlodderstedt @c2bo @jruizaranguren @nemqe

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -798,8 +798,8 @@ For cryptographic binding, the Client has the following options defined in (#cre
...
+* `credential_identifier`: REQUIRED when an Authorization Details of type `openid_credential` was returned from the Token Response. It MUST NOT be used otherwise. A string that identifies a Credential Dataset that is requested for issuance. When this parameter is used, the `credential_configuration_id` MUST NOT be present.
+* `credential_configuration_id`: REQUIRED if an Authorization Details of type `openid_credential` was not returned from the Token Response or if `format` parameter is not present. It MUST NOT be used otherwise. A string that identifies one of the keys in the name/value pairs stored in the `credential_configurations_supported` Credential Issuer metadata. The value of this parameter MUST match with one of those in the Credential Issuer metadata. When this parameter is used, the `credential_identifier` MUST NOT be present.
```
nemqe, line 802: Isn't `format` going to be permanently no present after this change?

```suggestion
* `credential_configuration_id`: REQUIRED if an Authorization Details of type `openid_credential` was not returned from the Token Response. It MUST NOT be used otherwise. A string that identifies one of the keys in the name/value pairs stored in the `credential_configurations_supported` Credential Issuer metadata. The value of this parameter MUST match with one of those in the Credential Issuer metadata. When this parameter is used, the `credential_identifier` MUST NOT be present.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -849,13 +848,51 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
...
-      "eyJ0eXAiOiJvcGVuaWQ0dmNpL...Lb9zioZoipdP-jvh1WlA",
-      "eyJraWQiOiJkaWQ6ZXhhbXBsZ...KPxgihac0aW9EkL1nOzM"
+      "eyJ0eXAiOiJvcGVuaWQ0dmNpLXByb29mK2p3dCIsImFsZyI6IkVTMjU2IiwiandrIjp7Imt0eSI6IkVDIiwiY3J2IjoiUC0yNTYiLCJ4IjoiblVXQW9BdjNYWml0aDhFN2kxOU9kYXhPTFlGT3dNLVoyRXVNMDJUaXJUNCIsInkiOiJIc2tIVThCalVpMVU5WHFpN1N3bWo4Z3dBS18weGtjRGpFV183MVNvc0VZIn19",
+      "eyJraWQiOiJkaWQ6ZXhhbXBsZTplYmZlYjFmNzEyZWJjNmYxYzI3NmUxMmVjMjEva2V5cy8xIiwiYWxnIjoiRVMyNTYiLCJ0eXAiOiJKV1QifQ"
     ]
   }
 }
 ```
 
+Below is a non-normative example of a Credential Request for one Credential in W3C VCDM format using a Credential identifier from the Token Response and key proof type `ldp_vp`:
```
nemqe, line 858: Is this sentence correct, how do we know that it is a request for a W3C VCDM credential if we only have the `credential_identifier`?

I guess `ldp_vp` proof type can be used for any kind of credential request.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -826,18 +828,15 @@ Content-Type: application/json
...
-    "jwt": "..."
+    "jwt": "eyJraWQiOiJkaWQ6ZXhhbXBsZTplYmZlYjFmNzEyZWJjNmYxYzI3NmUxMmVjMjEva2V5cy8xIiwiYWxnIjoiRVMyNTYiLCJ0eXAiOiJKV1QifQ"
   }
 }
 ```
 
-Either the `proof` or `proofs` parameter MUST be present if the `proof_types_supported` parameter is present in the `credential_configurations_supported` parameter of the Issuer metadata for the requested Credential.
-
-Below is a non-normative example of a Credential Request for two Credential instances in an IETF SD-JWT VC [@!I-D.ietf-oauth-sd-jwt-vc] format using a Credential instance identifier and key proof type `jwt`:
+Below is a non-normative example of a Credential Request for two Credential instances in an IETF SD-JWT VC [@!I-D.ietf-oauth-sd-jwt-vc] format using a Credential identifier from the Token Response and key proof type `jwt`:
```
nemqe, line 839: Is this sentence correct, how do we know that it is a request for a IETF SD-JWT VC credential if we only have the credential_identifier?

nemqe, line 839: Is this sentence correct, how do we know what exact format we are asking for if we only have the `credential_identifier`?

Similar question for: https://github.com/openid/OpenID4VCI/pull/392/files?diff=split&w=0#diff-1f424614b35a9899813079f1b1f6218631a2aedd993368ccb89bb81a9eda0289R858

nemqe (approved):

openid-4-verifiable-credential-issuance-1_0.md
nemqe, line 802: This one confuses me a bit.

> or if `format` parameter is not present

Where is it not present, in this request (we removed it) or in the Authorization Details?

```suggestion
* `credential_configuration_id`: REQUIRED if an Authorization Details of type `openid_credential` was not returned from the Token Response. It MUST NOT be used otherwise. A string that identifies one of the keys in the name/value pairs stored in the `credential_configurations_supported` Credential Issuer metadata. The value of this parameter MUST match with one of those in the Credential Issuer metadata. When this parameter is used, the `credential_identifier` MUST NOT be present.
```

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: Credential identifiers are returned for a specific credential configuration or scope, so I think it should be clear

babisRoutis (approved):

babisRoutis (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -665,7 +665,7 @@ The Authorization Server might decide to authorize issuance of multiple instance
 
 In addition to the response parameters defined in [@!RFC6749], the Authorization Server MAY return the following parameters:
 
-* `authorization_details`: REQUIRED when the `authorization_details` parameter is used to request issuance of a certain Credential Configuration as defined in (#authorization-details). It MUST NOT be used otherwise. It is an array of objects, as defined in Section 7 of [@!RFC9396]. In addition to the parameters defined in (#authorization-details), this specification defines the following parameter to be used with the authorization details type `openid_credential` in the Token Response:
```
paulbastian, line 668: Please verify that this is your interpretation of the Hybrid DCP Call

jruizaranguren:
> I applied the changes discussed in hybrid call to this PR. Please check @Sakurann @jogu @babisRoutis @tlodderstedt @c2bo @jruizaranguren @nemqe

I find it ready for implementation :smile:

paulbastian (approved):
I approve as I submitted most of the changes

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -637,7 +637,7 @@ grant_type=authorization_code
 &client_assertion=eyJhbGciOiJSU...
 ```
 
-### Request Issuance of a Certain Credential using authorization_details Parameter
+### Request Credential using authorization_details Parameter
```
Sakurann, line 640:
```suggestion
### Request Credential Issuance using authorization_details Parameter
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -665,7 +665,7 @@ The Authorization Server might decide to authorize issuance of multiple instance
 
 In addition to the response parameters defined in [@!RFC6749], the Authorization Server MAY return the following parameters:
 
-* `authorization_details`: REQUIRED when the `authorization_details` parameter is used to request issuance of a certain Credential Configuration as defined in (#authorization-details). It MUST NOT be used otherwise. It is an array of objects, as defined in Section 7 of [@!RFC9396]. In addition to the parameters defined in (#authorization-details), this specification defines the following parameter to be used with the authorization details type `openid_credential` in the Token Response:
+* `authorization_details`: REQUIRED when the `authorization_details` parameter is used to request issuance of a certain Credential Configuration as defined in (#authorization-details). The Authorization Server may also support this when issuance was requested using `scope`. It is an array of objects, as defined in Section 7 of [@!RFC9396]. In addition to the parameters defined in (#authorization-details), this specification defines the following parameter to be used with the authorization details type `openid_credential` in the Token Response:
```
Sakurann, line 668:
```suggestion
* `authorization_details`: REQUIRED when the `authorization_details` parameter is used to request issuance of a Credential of a certain Credential Configuration as defined in (#authorization-details). OPTIONAL when `scope` parameter was used to request issuance of a Credential of a certain Credential Configuration. It is an array of objects, as defined in Section 7 of [@!RFC9396]. In addition to the parameters defined in (#authorization-details), this specification defines the following parameter to be used with the authorization details type `openid_credential` in the Token Response:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -798,8 +798,8 @@ For cryptographic binding, the Client has the following options defined in (#cre
...
+* `credential_identifier`: REQUIRED when an Authorization Details of type `openid_credential` was returned from the Token Response. It MUST NOT be used otherwise. A string that identifies a Credential Dataset that is requested for issuance. When this parameter is used, the `credential_configuration_id` MUST NOT be present.
+* `credential_configuration_id`: REQUIRED if an Authorization Details of type `openid_credential` was not returned from the Token Response. It MUST NOT be used otherwise. A string that identifies one of the keys in the name/value pairs stored in the `credential_configurations_supported` Credential Issuer metadata. The value of this parameter MUST match with one of those in the Credential Issuer metadata. When this parameter is used, the `credential_identifier` MUST NOT be present.
```
Sakurann, line 802:
```suggestion
* `credential_configuration_id`: REQUIRED if an `authorization_details` parameter of type `openid_credential` was not returned from the Token Response. It MUST NOT be used otherwise. String that uniquely identifies one of the keys in the name/value pairs stored in the `credential_configurations_supported` Credential Issuer metadata. The corresponding object in the `credential_configurations_supported` map MUST contain one of the value(s) used in the `scope` parameter in the Authorization Request. When this parameter is used, the `credential_identifier` MUST NOT be present.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2455,6 +2437,8 @@ The technology described in this specification was made available from contribut
 
    -15
 
+   * add `credential_configuration_id` to Credential Request
```
Sakurann, line 2440:
```suggestion
   * add an option to return credential_identifiers in the Token Response and use them in the Credential Request, when scopes are used in the Authorization Request.
      * add an option to use `credential_configuration_id` in the Credential Request when scopes were used in the authorization request and no credential_identifiers returned in the token response
```

Sakurann:
honestly, it feels weird AS returning credential_identifier in the aurhorization_details object in the token response, when AS uses scopes (and not authorization details) in the authorization request, but guess that is the least optionality.

Also added a section summarizing all of the options for clarity: https://github.com/openid/OpenID4VCI/pull/392/files#diff-1f424614b35a9899813079f1b1f6218631a2aedd993368ccb89bb81a9eda0289R189

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -186,6 +186,27 @@ The issuance can have multiple characteristics that can be combined depending on
 
 The following subsections illustrate some of the authorization flows supported by this specification.
 
+### Identifying Credential(s) Being Issued throughout the Issuance Flow {#identifying_credential}
```
danielfett, line 189:
```suggestion
### Identifying Credentials Being Issued Throughout the Issuance Flow {#identifying_credential}
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -186,6 +186,27 @@ The issuance can have multiple characteristics that can be combined depending on
...
+
+Below is the summary of how Credential(s) that are being issued are identified throughout the issuance flow:
+
+- In the Credential Offer, the Credential Issuer identifies offered Credential(s)
+  using `credential_configuration_ids` parameter.
```
danielfett, line 194:
```suggestion
  using the `credential_configuration_ids` parameter.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -186,6 +186,27 @@ The issuance can have multiple characteristics that can be combined depending on
 
 The following subsections illustrate some of the authorization flows supported by this specification.
 
+### Identifying Credential(s) Being Issued throughout the Issuance Flow {#identifying_credential}
+
+Below is the summary of how Credential(s) that are being issued are identified throughout the issuance flow:
+
+- In the Credential Offer, the Credential Issuer identifies offered Credential(s)
```
danielfett, line 193:
```suggestion
- In the Credential Offer, the Credential Issuer identifies offered Credential Configurations
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -186,6 +186,27 @@ The issuance can have multiple characteristics that can be combined depending on
...
+  using `credential_configuration_ids` parameter.
+- When the Wallet uses Authorization Details in the Authorization Request, the Wallet uses
+  `credential_configuration_id` parameter(s) or `format` and other Credential Format
+  specific parameter to identify requested Credential(s). In which case,
```
danielfett, line 197:
```suggestion
  either `credential_configuration_id` parameters or `format` and other Credential Format
  specific parameters to identify the requested Credentials. In this case,
```

danielfett (commented):
Mostly editorial feedback so far.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -798,8 +824,8 @@ For cryptographic binding, the Client has the following options defined in (#cre
...
+* `credential_identifier`: REQUIRED when an Authorization Details of type `openid_credential` was returned from the Token Response. It MUST NOT be used otherwise. A string that identifies a Credential Dataset that is requested for issuance. When this parameter is used, the `credential_configuration_id` MUST NOT be present.
+* `credential_configuration_id`: REQUIRED if an `authorization_details` parameter of type `openid_credential` was not returned from the Token Response. It MUST NOT be used otherwise. String that uniquely identifies one of the keys in the name/value pairs stored in the `credential_configurations_supported` Credential Issuer metadata. The corresponding object in the `credential_configurations_supported` map MUST contain one of the value(s) used in the `scope` parameter in the Authorization Request. When this parameter is used, the `credential_identifier` MUST NOT be present.
```
Sakurann, line 828:
```suggestion
* `credential_configuration_id`: REQUIRED if a `credential_identifiers` parameter was not returned from the Token Response. It MUST NOT be used otherwise. String that uniquely identifies one of the keys in the name/value pairs stored in the `credential_configurations_supported` Credential Issuer metadata. The corresponding object in the `credential_configurations_supported` map MUST contain one of the value(s) used in the `scope` parameter in the Authorization Request. When this parameter is used, the `credential_identifier` MUST NOT be present.
```

Sakurann:
discussed on the WG call, agreed to use credential_identifiers in the authorization_details in the token response even when AS used scopes in the authotization request.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -186,6 +186,26 @@ The issuance can have multiple characteristics that can be combined depending on
...
 
+### Identifying Credentials Being Issued Throughout the Issuance Flow {#identifying_credential}
+
+Below is the summary of how Credential(s) that are being issued are identified throughout the issuance flow:
+
+- In the Credential Offer, the Credential Issuer identifies offered Credential Configurations
+  using the `credential_configuration_ids` parameter.
+- When the Wallet uses Authorization Details in the Authorization Request, the Wallet uses
+  either `credential_configuration_id` parameters or `format` and other Credential Format
+  specific parameters to identify the requested Credentials. In this case,
```
Sakurann, line 197:
```suggestion
  specific parameters to identify the requested Credential Configurations. In this case,
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -186,6 +186,26 @@ The issuance can have multiple characteristics that can be combined depending on
...
+  the Authorization Server MUST return `credential_identifiers` parameter in the Token Response,
+  and the Wallet uses those `credential_identifier` values in the Credential Request.
+- When the Wallet uses `scope` parameter in the Authorization Request, the `scope` value(s)
+  are used to identify requested Credential(s). In this case, the Authorization Server has two options.
```
Sakurann, line 201:
```suggestion
  are used to identify requested Credential Configurations. In this case, the Authorization Server has two options.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -186,6 +186,27 @@ The issuance can have multiple characteristics that can be combined depending on
 
 The following subsections illustrate some of the authorization flows supported by this specification.
 
+### Identifying Credential(s) Being Issued throughout the Issuance Flow {#identifying_credential}
+
+Below is the summary of how Credential(s) that are being issued are identified throughout the issuance flow:
```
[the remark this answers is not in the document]
Sakurann: made changes accordingly

PR #402, opened 2024-10-04
Clean Up Outdated Files in Repo

javereec:
While working on a PR, I noticed that there are some files in the repo that don't seem to be maintained and contain potentially conflicting information to the spec. While they might be useful, maybe they are better hosted at a different location.

I was also wondering if we can outright remove https://github.com/openid/OpenID4VCI/tree/main/diagrams as those diagrams for example contain `c_nonce` references that will be removed after https://github.com/openid/OpenID4VCI/pull/381 is merged


jogu (approved):

javereec:
@jogu what is your opinion on removing the diagrams directory?

Sakurann (approved):
I am ok removing these. though I kind of miss the table with all of the parameters in one view...

jogu:
@Sakurann 

> @jogu what is your opinion on removing the diagrams directory?

It does look like that directory hasn't had much love lately. I definitely wouldn't be against removing it.

jogu:
Editorial, 2 approvals -> merging!

Issue #394, opened 2024-09-17
Consider removing the `c_nonce_expires_in` parameter


tplooker:
The `c_nonce_expires_in` parameter is intended to signal to the client/wallet when the `c_nonce` expires and therefore when a client/wallet should obtain a new one. However, given a credential issuer can invalidate a `c_nonce` at any stage this parameter is a bit of an un-reliable signal to clients/wallets. Practically this means clients/wallets will have to account for an error at the credential response endpoint due to an invalid nonce even if it isn't expired. For that reason I think we should remove the `c_nonce_expires_in` parameter and therefore simplify the nonce handling.

bc-pi:
Concur with the removal of `c_nonce_expires_in` for the reasons @tplooker mentions and others like just general aesthetics. I was wanting to drop it with the work on #381 but decided to *try* and keep the scope of changes there narrow in hopes of landing[1] the PR in a timely manner.

[1] I think that's what the cool kids say nowadays

babisRoutis:
Hi @tplooker  & @bc-pi 

Totally agree.

IMHO, it is easier for a wallet to handle an `invalid_proof` by obtaining a new `c_nonce`, re-calculating the proofs and repeating the call, than having to track the `c_nonce_expires_in`

tplooker:
> Would others agree?

I do and would be happy to do the PR if we have consensus.

jogu:
I think the practical outcome of removing `c_nonce_expires` might well be that many wallets treat nonce as single use and fetch a new one immediately prior to generating each set of proofs. (Maybe that's fine but I thought I'd mention it.)

Basically the "well I don't know if this one might still be valid so I might as well get a new one before I do a bunch of potentially 'expensive' generation of proofs" approach.

tplooker:
> I think the practical outcome of removing c_nonce_expires might well be that many wallets treat nonce as single use and fetch a new one immediately prior to generating each set of proofs. (Maybe that's fine but I thought I'd mention it.)

If we are concerned that this sort of implementation behaviour might set in, we could always add some implementation guidance that makes it clear that a wallet should continue to use an obtained nonce until its given a new one, for instance DPoP has the following to say on this topic.

"It is up to the authorization server when to supply a new nonce value for the client to use. The client is expected to use the existing supplied nonce in DPoP proofs until the server supplies a new nonce value."

PR #404, opened 2024-10-10
remove c_nonce and c_nonce_expires_in from credential endpoint response

### LINKED ISSUE #393: remove the option to return c_nonce from the credential response
This only makes sense when/if/after #381 is merged. 

_Originally posted by @bc-pi in relation to https://github.com/openid/OpenID4VCI/issues/381#issuecomment-2357047069_

> > Discussed on the WG call, direction on removing section 7.3.2 and consolidating its content in Nonce Endpoint definition. https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-7.3.2

> While I was part of that discussion, after looking again at that section and content of this PR, I don't see how that consolation could actually manifest in a meaningful way. I suggest this PR proceed without and consider that as part of the soon to be separate issue on removing an option to return c_nonce from the credential [error] response.

 > > also agreed to open a separate issue on removing an option to return c_nonce from the credential error response, so that we can merge this PR as-is (@bc-pi).

that is this

mickrau:
resolves #393

Simplify spec by only provide one way to get a c_nonce


paulbastian (approved):

c2bo (commented):
This PR provides a nice simplification but from my understanding has 2 potential downsides:
- adds a round trip in the corner cases where a server rejects a proof because the nonce was deemed invalid
- removes the possibility to provide stateful nonces - the nonce endpoint is not protected and there is no possibility (I think?) for this right now

The only thing that worries me a bit is losing the possibility to provide stateful nonces, but I am not entirely sure if we need them - are people using that right now? If we really want/need to preserve the ability to have stateful nonces, then we could also add the option to have the nonce endpoint protected and signal that via metadata I guess?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1100,7 +1092,7 @@ If the Wallet is requesting the issuance of a Credential that is not supported b
   * `invalid_credential_request`: The Credential Request is missing a required parameter, includes an unsupported parameter or parameter value, repeats the same parameter, or is otherwise malformed.
   * `unsupported_credential_type`: Requested Credential type is not supported.
   * `unsupported_credential_format`: Requested Credential Format is not supported.
-  * `invalid_proof`: The `proof` or `proofs` parameter in the Credential Request is invalid: (1) if both fields are missing, or (2) both are present simultaneously, or (3) one of the provided key proofs is invalid, or (4) if a `c_nonce` was previously provided and at least one of the key proofs is not linked to a valid `c_nonce` value (refer to (#issuer-provided-nonce)).
+  * `invalid_proof`: The `proof` or `proofs` parameter in the Credential Request is invalid: (1) if both fields are missing, or (2) both are present simultaneously, or (3) one of the provided key proofs is invalid, or (4) if a `c_nonce` was previously provided and at least one of the key proofs is not linked to a valid `c_nonce` value (refer to (#nonce-response)).
```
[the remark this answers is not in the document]
paulbastian: agree very much, distinguishing this error, makes it clear for the wallet what to do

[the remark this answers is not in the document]
mickrau: Thanks @c2bo, i also thought about a specific error code and like your suggestion

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1100,7 +1092,8 @@ If the Wallet is requesting the issuance of a Credential that is not supported b
...
   * `unsupported_credential_type`: Requested Credential type is not supported.
   * `unsupported_credential_format`: Requested Credential Format is not supported.
-  * `invalid_proof`: The `proof` or `proofs` parameter in the Credential Request is invalid: (1) if both fields are missing, or (2) both are present simultaneously, or (3) one of the provided key proofs is invalid, or (4) if a `c_nonce` was previously provided and at least one of the key proofs is not linked to a valid `c_nonce` value (refer to (#issuer-provided-nonce)).
+  * `invalid_proof`: The `proof` or `proofs` parameter in the Credential Request is invalid: (1) if both fields are missing, or (2) both are present simultaneously, or (3) one of the provided key proofs is invalid, or (4) if at least one of the key proofs is not linked to a required `c_nonce` value (refer to (#nonce-response)).
+  * `invalid_nonce`: The `proof` or `proofs` parameter in the Credential Request uses an invalid nonce: at least one of the key proofs is not linked to a valid `c_nonce` value. The wallet should retrieve a new `c_nonce` value (refer to (#nonce-endpoint)).
```
nemqe, line 1096: If we have the `invalid nonce` error would it make sense to use it also for the `invalid_proof` case (4)?

nemqe (approved):

nemqe:
If I see it correctly this PR also addresses https://github.com/openid/OpenID4VCI/issues/394 right?

Sakurann:
with my chair hat on, noting that there is not enough discussion in the issue to be called an agreement to remove this feature.

paulbastian:
Neither was there enough discussion on this in the none endpoint PR, but Brian just said he doesn't see this. I think the PR helps to make people understand this and reflects our real developer experience from implementing nonce endpoint.

Sakurann:
Brian first did a PR without discussion and that one got closed, we did discuss that in the following calls and there are multiple other issuer where the topic was discussed so when the PR came up the second time, the WG was more prepared.

bc-pi:
> If I see it correctly this PR also addresses #394, right?

No, the `## Nonce Response {#nonce-response}` section  still has a `c_nonce_expires_in` so this one doesn't fully address #394.

https://github.com/openid/OpenID4VCI/pull/404/files#diff-1f424614b35a9899813079f1b1f6218631a2aedd993368ccb89bb81a9eda0289R755

bc-pi:
> Neither was there enough discussion on this in the none endpoint PR, but Brian just said he doesn't see this.

I'm not entirely sure I know what that means. But there were hundreds of comments across #381, #199, & #39 discussing removal of c_nonce* from the AS's token endpoint. The issue behind this PR #393 and an issue to get rid of c_nonce_expires_in in general #394 were broken out from #381 in an attempt to keep its scope tight and allow it to proceed. The nonce endpoint was added to #381, which was an expansion of scope, but as a condition of acceptance by a few key opponents. And even that had quite a bit of discussion.

bc-pi:
> This PR provides a nice simplification but from my understanding has 2 potential downsides:
> 
>     * adds a round trip in the corner cases where a server rejects a proof because the nonce was deemed invalid
> 
>     * removes the possibility to provide stateful nonces - the nonce endpoint is not protected and there is no possibility (I think?) for this right now
> 
> 
> The only thing that worries me a bit is losing the possibility to provide stateful nonces, but I am not entirely sure if we need them - are people using that right now? 

Stateful with respect to the access token or something contained therein. But good question. 

>If we really want/need to preserve the ability to have stateful nonces, then we could also add the option to have the nonce endpoint protected and signal that via metadata I guess?

That could be signaled by a 401 at the nonce endpoint too.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1100,7 +1092,8 @@ If the Wallet is requesting the issuance of a Credential that is not supported b
   * `invalid_credential_request`: The Credential Request is missing a required parameter, includes an unsupported parameter or parameter value, repeats the same parameter, or is otherwise malformed.
   * `unsupported_credential_type`: Requested Credential type is not supported.
   * `unsupported_credential_format`: Requested Credential Format is not supported.
-  * `invalid_proof`: The `proof` or `proofs` parameter in the Credential Request is invalid: (1) if both fields are missing, or (2) both are present simultaneously, or (3) one of the provided key proofs is invalid, or (4) if a `c_nonce` was previously provided and at least one of the key proofs is not linked to a valid `c_nonce` value (refer to (#issuer-provided-nonce)).
+  * `invalid_proof`: The `proof` or `proofs` parameter in the Credential Request is invalid: (1) if both fields are missing, or (2) both are present simultaneously, or (3) one of the provided key proofs is invalid, or (4) if at least one of the key proofs is not linked to a required `c_nonce` value (refer to (#nonce-response)).
```
nemqe, line 1095:
```suggestion
  * `invalid_proof`: The `proof` or `proofs` parameter in the Credential Request is invalid: (1) if both fields are missing, or (2) both are present simultaneously, or (3) one of the provided key proofs is invalid, or (4) if at least one of the key proofs does not contain a `c_nonce` value (refer to (#nonce-response)).
```

openid-4-verifiable-credential-issuance-1_0.md
nemqe, line 1096:
```suggestion
  * `invalid_nonce`: The `proof` or `proofs` parameter in the Credential Request uses an invalid nonce: at least one of the key proofs does not contain a valid `c_nonce` value. The wallet should retrieve a new `c_nonce` value (refer to (#nonce-endpoint)).
```

bc-pi (approved):

Sakurann:
agreed to proceed with this PR, merging once 1-2 more approvals

jogu (approved):

peppelinux (approved):

tlodderstedt (approved):

Sakurann:
@mickrau can you please add a changelog entry that this option was removed? I will merge this PR after that

c2bo (approved):

mickrau:
> @mickrau can you please add a changelog entry that this option was removed? I will merge this PR after that

@Sakurann  done

Sakurann (approved):

Issue #151, opened 2023-12-14
Remove the "claims" parameter from authorizastion_details in SD-JWT/mdoc


paulbastian:
Selective-Disclosure enabled credential formats give the inherent mechanism for data minimization.

Therefore, the need to specify the claims that shall be in a credential (done in authorization_details) adds unnecessary complexity, as issuers can issue credential containing all claims and wallets can use selective disclosure to release a subset to Verifiers.

Furthermore, I think this feature is hard to implement and hard to understand for users. Hence, I guess it will rarely be implemented for SD-JWT/mdoc.

Proposal: Delete the `claims` parameter for SD-JWT VC and mdoc from Authorization Details and Credential Request

peppelinux:
Your proposal simplifies a lot the implementations. Considering that SD-JWT has decoys I don't see any problem in having a predefined number of attributes in the credentials.

jogu:
I feel there are potential concerns we should discuss before removing it.

The claims parameter allows the wallet to avoid being exposed to PII that it does not wish to receive (in contrast to the suggested solution, which I believe is that the wallet receives all the PII then discards that which it does not want). Hence removing the claims parameter might violate the data minimisation principle of GDPR and similar regulations?

TakahikoKawasaki:
The specification will continue to undergo breaking changes until it finds a way to enable the illustrated concept.

![credential-narrowing-down](https://github.com/openid/OpenID4VCI/assets/1641166/a7ea8e39-84ac-4914-93d9-dbbd40d9971f)

peppelinux:
Starting from a real-world or user perspective I love the story of the user that obtains the physical digital identity card.
There, there's no way to decide which claims to be included, the credential is one with a predefined set of attributes.

from the data minimization perspective I wouldn't see impacts since the data minimization is related to the presentation. The resource owner is the user and the credential and the data contained in it is in possession of the user and these can be selectively disclosed during the presentation phase.

data minimization is related to the scope of the request, if the scope is obtaining data for what we're the owners I don't see any problem.

@fmarino-ipzs WDYT? ^

Another point would be reducing the feature that other implementation may require, where custom/customizable sub set of attributes might be requested.

simplification here costs reduction of feature. the discussion is: is this feature needed or interesting for the future?

paulbastian:
Agree with @peppelinux , my arguments:

- its very hard to do this in a good UX
- its very hard for users to udnerstand
- its complicated for issuers and wallets to implement
[...]

PR #405, opened 2024-10-11
remove `claims` parameter from ISO mdoc and SD-JWT VC Credential Request

paulbastian:
Partly solves #151

Remove the `claims`parameter from Credential Request for Credential Formats that support selective Disclosure anyway as this:
- is a potential privacy leak to the Credential Issuer
- hinders interoperability when users have a Credential with subset of claims in their wallet and other Verifiers want other claims (which will be unexplainable to the user)
- this has been an interop issues in LSP hackathon as particpants interpreted this differently


Sakurann:
why removing only for mdoc and not other credential formats?

paulbastian:
> why removing only for mdoc and not other credential formats?

it is removed for all credential formats that have selective disclosure themselves, as from discussions that was the only reasoning this parameter was introduced (to enable a kind of selective disclosure for jwt-vc and alike)

tplooker (approved):

nemqe (approved):

Sakurann:
noting that there is no agreement in the issue to remove this parameter. and this needs to be discussed before this can be merged.

paulbastian:
Happy to discuss this in the working group call!

bc-pi (approved):

jogu (approved):

Sakurann:
discussed in the WG, agreement to remove claims/credential_definition parameter for all credential formats from credential requests. but those parameters are kept for authorization details.

peppelinux (approved):

Sakurann:
so this PR hopefully entirely resolves #151 because there is an agreement to keep claims in authorization_details and only remove it in credential request

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2270,8 +2269,7 @@ The following is a non-normative example of an authorization details object with
 
 The following additional parameters are defined for Credential Requests and this Credential Format.
 
-* `vct`: REQUIRED when the `format` parameter is present in the Credential Request. It MUST NOT be used otherwise. It is a string as defined in (#server-metadata-sd-jwt-vc). This claim contains the type value of the Credential that the Wallet requests the Credential Issuer to issue.
-* `claims`: OPTIONAL. An object as defined in (#server-metadata-sd-jwt-vc).
+* `vct`: REQUIRED when the `format` parameter is present in the Credential Request. It MUST NOT be used otherwise. It is a string as defined in (#server-metadata-sd-jwt-vc).
```
Sakurann, line 2272:
```suggestion
* `vct`: REQUIRED when the `format` parameter is present in the Credential Request. It MUST NOT be used otherwise. It is a string as defined in (#server-metadata-sd-jwt-vc). This claim contains the type value of the Credential that the Wallet requests the Credential Issuer to issue.
```
this should probably stay

Sakurann (approved):

Sakurann:
during the hybrid WG call, agreed to remove all credential format specific claims from credential request. merging this PR since this PR was already ready to fulfill part of that agreement. There will be a follow up PR to complete what was agreed during a WG mtg - ie removing credential format specific claims from w3c vcdm format.

PR #419, opened 2024-11-20
remove c_nonce_expires_in from nonce response

### LINKED ISSUE #394: Consider removing the `c_nonce_expires_in` parameter

tplooker:
Resolves #394

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -774,7 +774,7 @@ Content-Length: 0
 
 The Credential Issuer provides a nonce value in the HTTP response with a 2xx status code and the following parameters included as top-level members in the message body of the HTTP response using the application/json media type:
 
-* `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)). This value MUST be unpredictable and unique for every response returned from the nonce endpoint.
+* `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)). This value MUST be unpredictable.
 
 Due to the temporal and contextually sensitive nature of the `c_nonce` value, the Credential Issuer MUST make the response uncacheable by adding a `Cache-Control` header field including the value `no-store`.
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -775,9 +775,10 @@ Content-Length: 0
 The Credential Issuer provides a nonce value in the HTTP response with a 2xx status code and the following parameters included as top-level members in the message body of the HTTP response using the application/json media type:
 
 * `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)).
-* `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`. This value serves only as a hint to the Client, indicating how long the Credential Issuer is likely to accept the `c_nonce` as valid.
 
 Due to the temporal and contextually sensitive nature of the `c_nonce` value, the Credential Issuer MUST make the response uncacheable by adding a `Cache-Control` header field including the value `no-store`.
+
+A wallet SHOULD assume that the returned `c_nonce` value remains valid and continue using it in credential requests (see [#credential-request]) until the credential endpoint returns an `invalid_nonce` error response.
 
 Below is a non-normative example of a Nonce Response:
 
@@ -787,8 +788,7 @@ Content-Type: application/json
 Cache-Control: no-store
 
 {
-  "c_nonce": "wKI4LT17ac15ES9bw8ac4",
-  "c_nonce_expires_in": 120
+  "c_nonce": "wKI4LT17ac15ES9bw8ac4"
 }
 ```
 
@@ -2572,6 +2572,7 @@ The technology described in this specification was made available from contribut
    * Fixed #239: Completed IANA Considerations section
    * add key attestation as additional information in a proof of possesion and new proof type
    * change credential format identifier `vc+sd-jwt` to `dc+sd-jwt` to align with the media type in draft -06 of [@I-D.ietf-oauth-sd-jwt-vc] and update `typ` accordingly in examples 
+   * removes `c_nonce_expires_in` from Nonce Endpoint
 
    -14
    
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -774,11 +774,9 @@ Content-Length: 0
 
 The Credential Issuer provides a nonce value in the HTTP response with a 2xx status code and the following parameters included as top-level members in the message body of the HTTP response using the application/json media type:
 
-* `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)).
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -775,10 +775,11 @@ Content-Length: 0
 The Credential Issuer provides a nonce value in the HTTP response with a 2xx status code and the following parameters included as top-level members in the message body of the HTTP response using the application/json media type:
 
 * `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)).
-* `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`. This value serves only as a hint to the Client, indicating how long the Credential Issuer is likely to accept the `c_nonce` as valid.
 
 Due to the temporal and contextually sensitive nature of the `c_nonce` value, the Credential Issuer MUST make the response uncacheable by adding a `Cache-Control` header field including the value `no-store`.
 
+A wallet SHOULD assume that the returned `c_nonce` value remains valid and continue using it in credential requests (see [#credential-request]) until the credential endpoint returns an `invalid_nonce` error response.
```
bc-pi, line 781: IMO this is unneeded and even suggests behavior that might be problematic with respect to efficient (from the client/wallet accessing its local or remote security thingy) processing. We should, again IMO, suggest getting a new nonce prior to each credential endpoint invocation or just say nothing. 
```suggestion
```

babisRoutis: > We should, again IMO, suggest getting a new nonce prior to each credential endpoint invocation or just say nothing.

:+1: 

Indeed, it would be much cleaner for the wallets to obtain a fresh `c_nonce` before placing a credential request.

I think a recommendation should be included to the specification, just to avoid issuers & wallets to treat `c_nonce` as an ad-hoc refreshable  `DPOP-Nonce`

peppelinux, line 781:
```suggestion
A Wallet SHOULD continue using the returned `c_nonce` value in credential requests (see [#credential-request]) until the Credential endpoint returns an `invalid_nonce` error response.
```

wallet -> Wallet

I have suggested a reformulation of the sentence, as I am not a big fan of assumptions, especially when they are not intended for sentient beings but for devices such as wallets.

Sakurann: discussed in a WG. rough consensus to remove this paragraph. and ensure that there is enough guidance around returning fresh nonces (seems to be already the case, but @tplooker to double check)
jfyi: DPoP language https://datatracker.ietf.org/doc/html/rfc9449#name-authorization-server-provid

jogu: I think this sentence came out of my fear that wallets would interpret the spec as "just fetch a fresh nonce each time" once we removed `c_nonce_expires_in`.  But it seems several people actually think wallets potentially doing that is an acceptable outcome. I'm good with dropping this sentence.

tplooker:
Thanks all, I have updated the PR to remove this sentence, I also added one sentence to the c_nonce definition to make it clear the expectations around its unpredictability and uniqueness across responses from the nonce endpoint.

leecam (approved):

babisRoutis (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -774,8 +774,7 @@ Content-Length: 0
 
 The Credential Issuer provides a nonce value in the HTTP response with a 2xx status code and the following parameters included as top-level members in the message body of the HTTP response using the application/json media type:
 
-* `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)).
-* `c_nonce_expires_in`: OPTIONAL. Number denoting the lifetime in seconds of the `c_nonce`. This value serves only as a hint to the Client, indicating how long the Credential Issuer is likely to accept the `c_nonce` as valid.
+* `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)). This value MUST be unpredictable and unique for every response returned from the nonce endpoint.
```
jogu, line 777: I'm not convinced we have consensus around "MUST be unique for every response returned from the nonce endpoint", particularly as there's no way to tie the request/response to a particular client etc. I'm sure people were talking about stateless implementations and given it's I believe an unprotected endpoint I think there are implementations that will want to limit the amount of work done in a call to the nonce endpoint.

awoie, line 777: We had something like the following in OIDC core:
```suggestion
* `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)). Sufficient entropy MUST be present in the `c_nonce` values used for its lifetime to prevent attackers from guessing values.
```

awoie: Does it not matter in the context of the entire OID4VCI flow?

jogu: In OpenID Connect the nonce is created by and send by the client to the AS, and when it gets the id_token back the client checks if it contains the nonce it sent. If an attacker obtains a nonce (and state) they could potentially start a separate flow using the same nonce/state and inject the resulting id_token back into the other flow, or things like that. (Because these requests are being sent through the browser nonce/state could potentially leak or the user phished into starting a new flow with a nonce associated with the attackers session.)

In OpenID4VCI at the credential endpoint the nonce is being fetched from the credential issuer and then included in the proof that the client is in possession of the key in the key proof - the nonce here is all about freshness and showing that it's not an old key proof. The nonces aren't actually bound to the client or session (the nonce endpoint doesn't require a client id or access token).

It probably needs sufficient entropy that it's infeasible for an attacker to pre-generate proofs for all possible values.

tplooker: I don't think that requiring the nonce returned here to be unique precludes stateless implementations, for instance one could used signed timestamps if they want to achieve stateless nonces. However, I'm ok with dropping the uniqueness requirement if we believe some implementations for other reasons would want to say re-use a nonce over a short time period. I do believe the unpredictable language needs to be kept though that was inspired by the language in the DPoP RFC.

tplooker: I agree the applicability of the nonce is different here when compared with OIDC, I'm comfortable with unpredictable being sufficient all be it a little subjective.

tplooker: Suggestion based on @jogu comment/

```suggestion
* `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)). This value MUST be unpredictable.
```

jogu: LGTM, thanks Tobias!

awoie (approved):

paulbastian (approved):

jogu (approved):

c2bo (approved):

bc-pi (approved):
The statement that this "value MUST be unpredictable" is a bit odd to me but who am I to judge?

PR #427, opened 2024-12-12
use path for mdocs too (pushing to Daniel's PR on VCI metadata)

Sakurann:
the reason why there are changes from PR #419 is because Daniel's PR's main was not up to date with remote main. so from Daniel's PR perspective, this PR effectively achieves git pull origin main, so please let's not hang up on this...


Sakurann:
daniel applied patch himself

PR #439, opened 2024-12-21
IANA considerations for key-attestation+jwt

### LINKED ISSUE #435: keyattestation+jwt looks like it needs an entry in IANA considerations section
`keyattestation+jwt` looks like it's defined by OID4VCI so we need an IANA considerations entry for it.

c2bo:
Fixes #435

I just copied the other media types entry, adjusted some text and kept Torsten as author - hope that is fine @tlodderstedt ?

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2688,6 +2688,28 @@ in the manner described in [@RFC6838].
 * Change controller: OpenID Foundation Digital Credentials Protocols Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
 * Provisional registration? No
 
+### application/keyattestation+jwt
+
+* Type name: `application`
+* Subtype name: `keyattestation+jwt`
+* Required parameters: n/a
+* Optional parameters: n/a
+* Encoding considerations: Uses JWS Compact Serialization, as specified in [@!RFC7515].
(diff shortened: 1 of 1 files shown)


paulbastian (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2688,6 +2688,28 @@ in the manner described in [@RFC6838].
 * Change controller: OpenID Foundation Digital Credentials Protocols Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
 * Provisional registration? No
 
+### application/keyattestation+jwt
+
+* Type name: `application`
+* Subtype name: `keyattestation+jwt`
```
Sakurann, line 2694:
```suggestion
* Subtype name: `key-attestation+jwt`
```

Sakurann (changes requested):
key-attestaion+jwt is clearer

openid-4-verifiable-credential-issuance-1_0.md
bc-pi: would need to update the actual text and example(s) in the document too

c2bo:
Changed it from `keyattestation+jwt` to `key-attestation+jwt` in the media type registration and key attestation section

Sakurann (approved):

bc-pi (approved):

selfissued (approved):

PR #441, opened 2025-01-09
Add clarification about client_id for wallet attestation

### LINKED ISSUE #431: `client_id` value of the wallet during issuance
The text current wallet attestation draft says:
> The JWT MUST contain a "sub" (subject) claim with a value corresponding to the "client_id" of the OAuth client.

the thing is.. when there has been no pre-existing relationship between the issuer and the wallet, which `client_id` value does the wallet uses..?

c2bo:
Closes #431 

Adds small text sections that explains how `sub` of the Wallet Attestation JWT works and should be chosen

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2558,7 +2558,7 @@ The following is a non-normative example of a Wallet Attestation:
 
 To use the Wallet Attestation towards the Authorization Server, the Wallet MUST generate a proof of possession according to Section 5.2 "Client Attestation PoP JWT" of Attestation-Based Client Authentication.
 
-The `sub` claim of the Wallet Attestation JWT is picked by the Wallet Provider and represents the `client_id` of the Wallet. This value should be shared by all Wallet Instances for privacy reasons.
+The `sub` claim of the Wallet Attestation JWT is picked by the Wallet Provider and represents the `client_id` of the Wallet. This value should be shared by all Wallet Instances from that Wallet Provider for privacy reasons.
 
 # IANA Considerations
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2558,7 +2558,7 @@ The following is a non-normative example of a Wallet Attestation:
 
 To use the Wallet Attestation towards the Authorization Server, the Wallet MUST generate a proof of possession according to Section 5.2 "Client Attestation PoP JWT" of Attestation-Based Client Authentication.
 
-The `sub` claim of the Wallet Attestation JWT is picked by the Wallet Provider and represents the `client_id` of the Wallet. This value should be shared by all Wallet Instances from that Wallet Provider for privacy reasons.
+The `sub` claim of the Wallet Attestation JWT is picked by the Wallet Provider and represents the `client_id` of the Wallet. For privacy reasons, this value is the same across Wallet Instances of that Wallet Provider.
 
 # IANA Considerations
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2558,6 +2558,8 @@ The following is a non-normative example of a Wallet Attestation:
 
 To use the Wallet Attestation towards the Authorization Server, the Wallet MUST generate a proof of possession according to Section 5.2 "Client Attestation PoP JWT" of Attestation-Based Client Authentication.
 
+The `sub` claim of the Wallet Attestation JWT is picked by the Wallet Provider and represents the `client_id` of the Wallet. This value should be shared by all Wallet Instances from that Wallet Provider for privacy reasons.
```
c2bo, line 2561: should this be normative (SHOULD)?

Sakurann:
```suggestion
The `sub` claim of the Wallet Attestation JWT is picked by the Wallet Provider and represents the `client_id` of the Wallet. For privacy reasons, this value is the same across Wallet Instances of that Wallet Provider.
```

I would honestly, add a short privacy considerations section about this and point to it from here

paulbastian (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1664,6 +1664,13 @@ for example, by including clear-text session information as a `state` parameter
 it in a `redirect_uri` parameter. A third party may observe such information through browser
 history, etc. and correlate the user's activity using it.
 
+### Wallet Attestation Subject {#walletattestation-sub}
+
+The Wallet Attestation as defined in (#wallet attestation) SHOULD NOT introduce a unique identifier specific to a single client.
```
Sakurann, line 1669:
```suggestion
The Wallet Attestation as defined in (#walletattestation) SHOULD NOT introduce a unique identifier specific to a single client.
```

Sakurann: shouldn't this be "MUST NOT"?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2558,6 +2565,8 @@ The following is a non-normative example of a Wallet Attestation:
 
 To use the Wallet Attestation towards the Authorization Server, the Wallet MUST generate a proof of possession according to Section 5.2 "Client Attestation PoP JWT" of Attestation-Based Client Authentication.
 
+The `sub` claim of the Wallet Attestation JWT is picked by the Wallet Provider and represents the `client_id` of the Wallet. For privacy reasons, this value is the same across Wallet instances of that Wallet Provider, see (#walletattestation-sub) for more details.
```
Sakurann, line 2568:
```suggestion
The `sub` claim of the Wallet Attestation JWT is picked by the Wallet Provider and represents the `client_id` of the Wallet instance. For privacy reasons, this value is the same across Wallet instances of that Wallet Provider, see (#walletattestation-sub) for more details.
```

openid-4-verifiable-credential-issuance-1_0.md
c2bo: Good question and for most systems I would agree. There might be ecosystems with other privacy properties and I wasn't sure if we should disallow that in the core spec -> I went with SHOULD, but I am happy to change

Sakurann (approved):

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: but wallet attestation format is defined in VCI, right? so i think it should be MUST NOT. i might be short sighted but what other privacy property can there be?

Sakurann: I am happy to continue the discussion in https://github.com/openid/oid4vc-haip/issues/171

PR #451, opened 2025-01-22
clarify issuer can be an AS too

### LINKED ISSUE #309: AS can act as an Issuer?
Obviously very wrong how it is, so approving.

Probably we should mention that the Credential Issuer may optionally also be an AS.

_Originally posted by @jogu in https://github.com/openid/OpenID4VCI/pull/299#pullrequestreview-2024835394_

Sakurann:
resolves #309

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -84,7 +84,7 @@ Presentation:
 : Data that is presented to a specific Verifier, derived from one or more Verifiable Credentials that can be from the same or different Credential Issuers. It can be of any Credential Format.
 
 Credential Issuer (or Issuer):
-:  An entity that issues Verifiable Credentials. In the context of this specification, the Credential Issuer acts as an OAuth 2.0 Authorization Server and/or a Resource Server (see [@!RFC6749]).
+:  An entity that issues Verifiable Credentials. In the context of this specification, the Credential Issuer acts as an OAuth 2.0 Resource Server (see [@!RFC6749]). The Credential Issuer might also act as an Authorization Server.
 
 Holder:
 :  An entity that receives Verifiable Credentials and has control over them to present them to the Verifiers as Presentations.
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -84,7 +84,7 @@ Presentation:
 : Data that is presented to a specific Verifier, derived from one or more Verifiable Credentials that can be from the same or different Credential Issuers. It can be of any Credential Format.
 
 Credential Issuer (or Issuer):
-:  An entity that issues Verifiable Credentials. In the context of this specification, the Credential Issuer acts as an OAuth 2.0 Resource Server (see [@!RFC6749]).
+:  An entity that issues Verifiable Credentials. In the context of this specification, the Credential Issuer acts as an OAuth 2.0 Authorization Server and/or a Resource Server (see [@!RFC6749]).
```
babisRoutis, line 87: IMHO, specification currently provides a clear definition that Credential Issuer is, first and foremost, a Resource Server, and it took some time (drafts) to de-associate Issuer from AS (updated sequence diagrams, removed c_nonce from token endpoint etc).

The fact that a Credential Issuer could also act a an AS (protecting itself) is an implementation detail and as such, perhaps related remark to the implementation consideration section could me add (keeping the current definition as is).

paulbastian (approved):

Sakurann:
WG discussion
- direction agreed.
- joseph to make a suggestion
- might need to make sure other sections are aligned.

openid-4-verifiable-credential-issuance-1_0.md
paulbastian, line 87:
```suggestion
:  An entity that issues Verifiable Credentials. In the context of this specification, the Credential Issuer acts as an OAuth 2.0 Resource Server, and may also act as an OAuth Authorization Server (see [@!RFC6749]).
```

jogu:
```suggestion
:  An entity that issues Verifiable Credentials. In the context of this specification, the Credential Issuer acts as an OAuth 2.0 Resource Server (see [@!RFC6749]). The Credential Issuer might also act as an Authorization Server.
```

awoie (approved):
approving if either joseph's or paul's suggestion was committed.

babisRoutis:
> approving if either joseph's or paul's suggestion was committed.

:+1:

PR #452, opened 2025-01-22
clarify batch management in the Wallet when the new batch of the credentials was received

### LINKED ISSUE #124: How to request new copies of a credential
In section [12.6](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-refreshing-issued-credentia) the refresh method is described as an update mechanism "when some values have changed".
What is not described is what will happen with the already issued credentials. Will they be revoked? Can the wallet send a value if the old ones should be revoked or not? Will the issuer give information about the handling when issuing the first credential?

Background of the question was the situation when I requested a batch of credentials because I want to use a dedicated JWT for each relying party to avoid linkability during the presentation. But during the issuance process I don't know how many copies I will need in the future so there could be the situation where the wallet asks the issuer for more copies of the credentials. For user experience I don't want the holder to manually start the process, both approaches are already covered by the specification.

So when my wallet is down to 5 credentials and it wants to fill up to 10 credentials again, does it have to query 5 credentials to fill it up or does it have to ask for 10 because all the old ones will be revoked (in case the issuer is supporting some kind of status management). Of course my wallet could check after the reissuance if the old ones are still valid and then start the issuer process again but now with 10 copies. But this seems to be a dirty solution to handle the problem.

Sakurann:
resolves #124

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1101,6 +1101,8 @@ If the Client requested an encrypted response by including the `credential_respo
 
 If the Credential Response is not encrypted, the media type of the response MUST be set to `application/json`.
 
+When the Wallet receives a new batch of the same Credential Datasets, the Wallet SHOULD delete any Credential Datasets it might have received previously.
+
 The following parameters are used in the JSON-encoded Credential Response body:
 
 * `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. This specification defines the following parameters to be used inside this object:
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1101,7 +1101,7 @@ If the Client requested an encrypted response by including the `credential_respo
 
 If the Credential Response is not encrypted, the media type of the response MUST be set to `application/json`.
 
-When the Wallet receives a new batch of Credentials containing a specific Credential Dataset, the Wallet SHOULD delete any Credentials containing the same Credential Dataset it might have received previously.
+When the Wallet receives new Credentials containing a specific Credential Dataset, the Wallet SHOULD delete any Credentials containing the same Credential Dataset it received in previous issuance responses if those Credentials are no longer valid.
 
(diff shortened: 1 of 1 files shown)


c2bo (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1101,6 +1101,8 @@ If the Client requested an encrypted response by including the `credential_respo
 
 If the Credential Response is not encrypted, the media type of the response MUST be set to `application/json`.
 
+When the Wallet receives a new batch of Credentials containing a specific Credential Dataset, the Wallet SHOULD delete any Credentials containing the same Credential Dataset it might have received previously.
```
jogu, line 1104: I wonder if this is slightly over specific - it doesn't give any advice for the case where the issuer doesn't support batch issuance, nor it is necessarily helpful in the case where the wallet wants a batch of 10 and the issuer only supports batches of 5 (the wallet would hence ask for two batches of 5 and definitely wouldn't delete the first batch of 5 when it got the second batch of 5...).

Maybe this is more generic advice?

```suggestion
When the Wallet receives new Credentials containing a specific Credential Dataset, the Wallet SHOULD delete any Credentials containing the same Credential Dataset it received in previous sessions.
```

(I'm not sure "session" is well defined but can't think of another phrase that means "delete credentials you received a while ago but not ones you received 10 seconds ago".)

tlodderstedt:
```suggestion
When the Wallet receives new Credentials containing a specific Credential Dataset, the Wallet SHOULD delete any Credentials containing the same Credential Dataset it received in previous issuance responses if those Credentials are no longer valid.
```

tplooker (approved):

tplooker (commented):
Related to #472 I think we need to merge the language there as how do you know whether the batch you have just received is materially different to the last one, if you are just obtaining new copies of a credential and the claim values are the same why should you be forced to delete the copies you have at the moment?

tplooker (changes requested):
Changing my review based on above comment, easy to resolve I just think we need to check these PR's are consistent with one and other.

Sakurann:
WG discussion:
- credential versioning has been agreed to be tackled in v1.1
- for 1.0, agreed to close this in favor of section 13.5 already saying something similar

PR #455, opened 2025-01-30
wallet-attestation+jwt media registry entry

peppelinux:
This PR adds the media type wallet-attestation+jwt in the media registry of this specification.

It cames from [OpenID HAIP](https://github.com/openid/oid4vc-haip/pull/86#issuecomment-2609703378) as it was asked me to move it to openid4vci.


Sakurann (approved):

paulbastian (changes requested):
https://drafts.oauth.net/draft-ietf-oauth-attestation-based-client-auth/draft-ietf-oauth-attestation-based-client-auth.html#name-client-attestation-jwt currently defines `oauth-client-attestation+jwt`, are we sure it makes sense to define something new here?

Sakurann:
closing since reference ietf draft already defines typ

PR #457, opened 2025-02-03
Change Cryptographic Holder Binding to Cryptographic Key Binding

### LINKED ISSUE #26: Holder Binding definition dilemma
### Original Reporter: peppelinux

Dear friends and authors, I read in the OpenID4VCI specs the definition of Holder Binding as follows

”Holder Binding: Ability of the Holder to prove legitimate possession of a Verifiable Credential.”

I assume that the statement above refers to the Proof of Possession, carried out by the Holder during the presentation stage, as an ability of the Holder of doing that.

Differently, thinking about Holder Binding, the following definition comes to mind

"characteristic or attribute of a credential, inseparable from it and resistant to tampering, which uniquely links a Verifiable Credential to the entity who requested it and of which it is the legitimate Holder"

javereec:
Fixes #26 

The issue described changing Holder Binding to Key Binding, but I don't believe that is correct as there are other ways to bind a credential to a Holder than to use key binding. I therefore updated Cryptographic Holder Binding to Cryptographic Key Binding, which might have been the intent anyway I presume.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2757,6 +2757,7 @@ The technology described in this specification was made available from contribut
    -16
 
    *
+   * Change Cryptographic Holder Binding to Cryptographic Key Binding
 
    -15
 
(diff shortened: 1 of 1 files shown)


Sakurann (approved):
thank you!

c2bo (approved):

bc-pi (approved):

jogu (approved):

PR #458, opened 2025-02-05
Move Proof Types from Credential Endpoint section to an appendix

### LINKED ISSUE #304: Consider moving Proof Type section to Annex
Proof Types section has becoming very large and it makes the Credential Endpoint section hard to read. Consider moving it to Annex or separate chapter?

javereec:
Fixes #304

I think this is self-explanatory. I chose to move it before the Key Attestation and Wallet Attestation parts as they refer to Proof Types so it felt more intuitive. I hope that's okay.


c2bo:
Just a quick reminder before I forget: This will very likely shift references from HAIP (and maybe also OpenID4VP) -> when this gets merged, we should check and fix those if necessary.

Sakurann:
You're right Christian. given that, is this worth the change..?

c2bo:
I like the change - I think it improves readability. I can make the fixes after this merge in both specs and just wanted to make sure we don't forget

c2bo (approved):

Sakurann:
what about ISO specs that reference out specs......?

Sakurann:
wg discussion - agreed we want to do this.
- move this to the latest annex, so that annex D and E for wallet and key attestations do not get new alphabets
- @paulbastian to help check

javereec:
@paulbastian @c2bo @Sakurann Merged in latest changes from main and moved proof types to appendix F as per last comments. Please review again.

paulbastian:
Oh damn, I already did that in #489

javereec:
@paulbastian Oh, I see. Would be good to close this PR before you do that next time.

javereec:
Superseeded by #489

Issue #354, opened 2024-06-25
Make `credential_signing_alg_values_supported` and `proof_signing_alg_values_supported` polymorphic or credential format-specific


awoie:
Both `credential_signing_alg_values_supported` and `proof_signing_alg_values_supported` are defined in the general section of OID4VCI Issuer Metadata and are currently defined as array of String values. While this is possible for JOSE algorithms, it is not possible for COSE algorithms.

Either make both polymorphic with the type depending on the format, or just make them format-specific parameters.

babisRoutis:
Hi @awoie ,

I believe that `proof_signing_alg_values_supported`should  be defined, as needed, for each proof type (jwt, cwt & ld_vp),  The general section is confusing (since a reader could miss paragraph 7) and perhaps should be removed.

Similarly, `credential_signing_alg_values_supported` should be format specific.

awoie:
> Hi @awoie ,
> 
> I believe that `proof_signing_alg_values_supported`should be defined, as needed, for each proof type (jwt, cwt & ld_vp), The general section is confusing (since a reader could miss paragraph 7) and perhaps should be removed.
> 
> Similarly, `credential_signing_alg_values_supported` should be format specific.

Fully agree.

Sakurann:
it should not be credential format specific. redefining the value of these parameters from array of strings to array of strings or integers is preferable (if cbor credential format people agree that integer is the only way to express alg values supported, since string is also possible I think)

jogu:
Consensus on today's WG call that we would have only the name defined in the credential-format-independent part of the spec (i.e. we would no longer say they are strings), and the credential formats would define `credential_signing_alg_values_supported` type and values, and the proof formats section would do the same for `proof_signing_alg_values_supported`.

jogu:
I believe we also got consensus that COSE algorithms should be described by integers, because even though there are string names defined as per https://www.rfc-editor.org/rfc/rfc8152#section-16.4 the integer values https://www.iana.org/assignments/cose/cose.xhtml#algorithms are the thing that should be used.

Issue #303, opened 2024-04-25
Consider using COSE algorithm identifiers for credential_signing_alg_values_supported for mso_mdoc


awoie:
ISO mdocs use COSE_Sign1 to secure the MSO using COSE algorithms, not JOSE algorithms. COSE algorithms are identified by integer values according to the IANA registry. We should change the mso_mdoc section to use those identifiers instead of JOSE algorithm identifiers.

awoie:
However, when using COSE algorithms, the issue is that some algorithm identifiers such as ES256 (-7) are polymorphic identifiers. For example, it is possible to use any brainpool curve, or NIST curve (of any length) with ES256 since ES256 only refers to ECDSA with SHA256. I know that COSE is going to deprecate polymorphic identifiers but 1) ISO will keep referring to ES256 (-7), and 2) fully defined algorithm identifiers have not been registered with IANA yet.

As an interim solution, we could say that credential_signing_alg_values_supported refer to JOSE algorithms which should then be mapped to their COSE counter part. Currently, the spec says to use one of the algorithms defined in ISO but ISO does not have a list of algorithm identifiers.

awoie:
Currently, we have:
> Cryptographic algorithm names used in the credential_signing_alg_values_supported parameter SHOULD be one of those defined in [[ISO.18013-5](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#ISO.18013-5)].

We would need to change to:
> Cryptographic algorithm names used in the credential_signing_alg_values_supported parameter SHOULD be one of those defined in [[ISO.18013-5](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#ISO.18013-5)] represented as JOSE algorithm identifiers.

Sakurann:
> We would need to change to:
> 
> > Cryptographic algorithm names used in the credential_signing_alg_values_supported parameter SHOULD be one of those defined in [[ISO.18013-5](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#ISO.18013-5)] represented as JOSE algorithm identifiers.

Sounds like we would also need a sentence saying that `JOSE algorithms for credential_signing_alg_values_supported should then be mapped to their COSE counter part`. 

[...]

PR #460, opened 2025-02-07
make credential_signing_alg_values_supported type and values format specific

awoie:
This PR does:
   * make type and values for `credential_signing_alg_values_supported` format specific
   * change algorithm identifiers for `credential_signing_alg_values_supported` to COSE algorithm values for mdocs
   * add `credential_signing_crv_values_supported` to mdoc format because of polymorphic COSE algorithm identifiers

Fixes https://github.com/openid/OpenID4VCI/issues/354 and fixes https://github.com/openid/OpenID4VCI/issues/303

Note that this is also how it is used in 23220-3 atm.


Sakurann (approved):
i think this is an improvement - thank you!

davidz25 (approved):
I think this is an improvement, thanks.

(As an aside, would be nice if we could depend on [fully specified algorithms](https://datatracker.ietf.org/doc/draft-ietf-jose-fully-specified-algorithms/) - that way we wouldn't need e.g. `credential_signing_crv_values_supported`. But I don't think we can add that dependency.)

paulbastian:
agree with David, we should use fully-specified algorithms right away to make things easier. It's close to final and apparently its not a too big of a problem to reference non-final specs.

tplooker:
> (As an aside, would be nice if we could depend on [fully specified algorithms](https://datatracker.ietf.org/doc/draft-ietf-jose-fully-specified-algorithms/) - that way we wouldn't need e.g. credential_signing_crv_values_supported. But I don't think we can add that dependency.)

If we did this too though we wouldn't be able to point at the COSE IANA registry like we are as we are because the polymorphic algorithm identifiers are already registered there.

Hopefully fully specified algs just become more common use in future thus making this metadata element redundant.

tplooker (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2190,12 +2199,14 @@ The Credential Format Identifier is `mso_mdoc`. This refers to the Mobile Securi
...
+Cryptographic algorithm identifiers used in the `credential_signing_alg_values_supported` parameter are numeric values and SHOULD be one of those COSE algorithm values defined in [@IANA.COSE].
 
 The following additional Credential Issuer metadata parameters are defined for this Credential Format for use in the `credential_configurations_supported` parameter, in addition to those defined in (#credential-issuer-parameters).
 
 * `doctype`: REQUIRED. String identifying the Credential type, as defined in [@!ISO.18013-5].
 * `claims`: OPTIONAL. An array of claims description objects as defined in (#claims-description-issuer-metadata).
+* `credential_signing_crv_values_supported`: OPTIONAL. Array of cryptographic curve identifiers that the Issuer uses with one of the algorithms listed in `credential_signing_alg_values_supported` to sign the issued Credential. Curve identifier types and values used are numeric values and SHOULD be one of those COSE eliptic curve values defined in [@IANA.COSE]. Note that `credential_signing_crv_values_supported` MAY be used if one of the algorithms in `credential_signing_alg_values_supported` supports multiple curves.
```
Sakurann, line 2208: this needs to be changed to align with https://github.com/openid/OpenID4VP/pull/553

Sakurann (changes requested):
needs to be changed to align with https://github.com/openid/OpenID4VP/pull/553

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: How does this relate to #495 ?

awoie:
Closing this PR in favor of #495

Issue #462, opened 2025-02-19
Misunderstanding of what AuthorizedNamespaces is


davidz25:
Section _A.2.4. Credential Response_ says the following

```
The value of the credential claim in the Credential Response MUST be a string
that is the base64url-encoded representation of the CBOR-encoded IssuerSigned
structure, as defined in [ISO.18013-5]. This structure SHOULD contain all
Namespaces and IssuerSignedItems that are included in the AuthorizedNamespaces
of the MobileSecurityObject.
```

According to 18013-5 `AuthorizedNamespaces` is a mechanism for the issuer to convey that `DeviceKey` is authorized to sign data elements in that name space and to be returned in DeviceSigned. So it doesn't make any sense to say "This structure SHOULD contain all Namespaces and IssuerSignedItems that are included in the AuthorizedNamespaces of the MobileSecurityObject.". (Also, if you look at MSOs being minted today across e.g. US mDL issuers, no-one is actually using DeviceSigned at all to return data elements, as far as I know.)

I also don't think it make sense to specify what the structure SHOULD contain, I mean, it's already completely specified by 18013-5 what it contains. I would just strike the entire last sentence in the quoted paragraph.

davidz25:
> DeviceSigned should be allowed as well.

So, yeah, I mean, I don't disagree. But I don't think we can say anything about the encoding of the data... because the whole point of `DeviceSigned` data elements is that the device comes up with them using multiple signals, not just what the issuer tells them, if anything.

Here are three examples:
* `age_over_18` could be a device-signed data element and in this case there are two signals, namely `birth_date` and also what the device believes to be the current time.
* `device_matched_authorized_holder` could be a device signed boolean calculated by using sensors on the device and comparing against issuer-provided biometrics provided out-of-band, set to `true` only if there's a match.
 * `holder_verified_email_address` device signed tstr which is the email address obtained from e.g. asking the user and verifying it out-of-band.

At least this is the spirit of device-signed and examples often used in ISO.

So IOW, it's hard to specify what format the data used - if any - should be transmitted by the issuer. There's just nothing we can say about it. I do think we should allow implementations to put other things in `IssuerSigned` map but I don't think it needs to be standardized.

Sakurann:
@davidz25 could you please do a small quick PR removing `This structure SHOULD contain all Namespaces and IssuerSignedItems that are included in the AuthorizedNamespaces of the MobileSecurityObject.`. ?

Sakurann:
PR merged

PR #469, opened 2025-03-17
Clarify credential response for ISO mdoc.

davidz25:
This change removes the reference to AuthorizedNamespaces of the MobileSecurityObject since this doesn't apply to issuer-signed data.

Fixes Issue #462.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2224,7 +2224,7 @@ The following is a non-normative example of an authorization details object with
 
 ### Credential Response
 
-The value of the `credential` claim in the Credential Response MUST be a string that is the base64url-encoded representation of the CBOR-encoded `IssuerSigned` structure, as defined in [@!ISO.18013-5]. This structure SHOULD contain all `Namespaces` and `IssuerSignedItems` that are included in the `AuthorizedNamespaces` of the `MobileSecurityObject`.
+The value of the `credential` claim in the Credential Response MUST be a string that is the base64url-encoded representation of the CBOR-encoded `IssuerSigned` structure, as defined in [@!ISO.18013-5].
 
 The following is a non-normative example of a Credential Response containing a Credential of format `mso_mdoc`.
 


Sakurann (approved):

c2bo (approved):

awoie (approved):

awoie:
I want to note though that `nameSpaces` is an optional element of `IssuerSigned` and therefore we should require that `nameSpaces` should be there as well, unless you want to transmit the MSO only which might be a legitimate use case. But if it is we would need some language on that as well, right? @davidz25

davidz25:
> I want to note though that `nameSpaces` is an optional element of `IssuerSigned` and therefore we should require that `nameSpaces` should be there as well, unless you want to transmit the MSO only which might be a legitimate use case. But if it is we would need some language on that as well, right? @davidz25

(Sorry for the delayed response, I missed this reply)

I'm not sure we need any language... if `nameSpaces` is empty it means there are no issuer signed data elements which is unusual but still valid, for example the mdoc could have entirely DeviceSigned data elements. The MSO is still required b/c it would indicate which DeviceSigned data elements the data can be returned in.

If we're to add language here, I think it would be to point out that the issuer may want to convey additional information in order to instruct the wallet what to put in DeviceSigned.... for the reasons described in [this comment in #462](https://github.com/openid/OpenID4VCI/issues/462#issuecomment-2701564773) this is going to be hard to standardize and probably initially specific to out-of-band agreements between the issuer and the wallet... so my gut feeling is that we shouldn't say too much about it? So, tl;dr, happy to add a comment but not sure what it should be

awoie:
> > I want to note though that `nameSpaces` is an optional element of `IssuerSigned` and therefore we should require that `nameSpaces` should be there as well, unless you want to transmit the MSO only which might be a legitimate use case. But if it is we would need some language on that as well, right? @davidz25
> 
> (Sorry for the delayed response, I missed this reply)
> 
> I'm not sure we need any language... if `nameSpaces` is empty it means there are no issuer signed data elements which is unusual but still valid, for example the mdoc could have entirely DeviceSigned data elements. The MSO is still required b/c it would indicate which DeviceSigned data elements the data can be returned in.
> 
> If we're to add language here, I think it would be to point out that the issuer may want to convey additional information in order to instruct the wallet what to put in DeviceSigned.... for the reasons described in [this comment in #462](https://github.com/openid/OpenID4VCI/issues/462#issuecomment-2701564773) this is going to be hard to standardize and probably initially specific to out-of-band agreements between the issuer and the wallet... so my gut feeling is that we shouldn't say too much about it? So, tl;dr, happy to add a comment but not sure what it should be

This makes sense. I'm now convinced we don't need additional language.

PR #476, opened 2025-03-31
Add support for the Digital Credentials API

### LINKED ISSUE #410: OID4VCI with Digital Credentials API
we added support to use OpenID4VP over Digital Credentials API, it would be beneficial to add support to use  Digital Credentials API for VCI, too, when passing a credential offer. it would greatly help with wallet selector. Not so much with cross-device security, since credential offer is not signed, but adding an origin information to the credential offer might help already?

This was discussed at IIW and since the wallet matcher needs enough information to make a decision whether the wallet can support a credential offer from a specific issuer, and wallet matcher cannot make an external call, the preferred direction seems to be to add entire issuer metadata and authorization server metadata by value to the credential offer, when it is passed over the browser API. and the wallet can make its own decision which issuer it can talk to, based on that information.

leecam:
Adds a new annex to defining support for the DC API

closes #410

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2052,6 +2052,83 @@ regulation), the Credential Issuer should properly authenticate the Wallet and e
   </front>
 </reference>
 
+<reference anchor="W3C.Digital_Credentials_API" target="https://wicg.github.io/digital-credentials/">
+        <front>
+          <title>Digital Credentials API</title>
+		  <author fullname="Marcos Caceres">
+            <organization>Apple Inc.</organization>
+          </author>
+          <author fullname="Tim Cappalli">
+            <organization>Okta</organization>
+          </author>
+          <author fullname="Sam Goto">
+            <organization>Google</organization>
+          </author>
+        </front>
+</reference>
+
+# OpenID4VCI over the Digital Credentials API {#dc_api}
+
+This section defines how to use OpenID4VCI with the Digital Credentials API.
+
+The name "Digital Credentials API" (DC API) encompasses the W3C Digital Credentials API [@!W3C.Digital_Credentials_API]
+as well as its native App Platform equivalents in operating systems (such as [Credential Manager on Android](https://developer.android.com/jetpack/androidx/releases/credentials)).
+The DC API allows web sites and native apps acting as Issuers to initiate the issuance of verifiable credentials.
+The API itself is agnostic to the Credential issuance protocol and can be used with different protocols.
+The Web Platform, working in conjunction with other layers, such as the app platform/operating system, and based on the permission of the End-User, will send the credential offer along with the Origin of the Issuer to the End-User's chosen Wallet.
+
+## Protocol
+
+To use OpenID4VCI with the Digital Credentials API (DC API), the exchange protocol value has the following format: `openid4vci-v<version>`.
+
+The value `1` MUST be used for the `<version>` field to indicate the request and response are compatible with this version of the specification.
+
+The following exchange protocol values are defined by this specification:
+
+* `openid4vci-v1`
+
+## Credential Offer {#dc_api_credential_offer}
+
+The Issuer sends the Credential Offer object to the Digital Credentials API. All of the Credential Offer parameters MAY be used. The Credential Offer object has been extended to include the following parameters:
+
+
+* `credential_issuer_metadata`: REQUIRED. The Issuer's Credential Issuer Metadata object as defined in (#credential-issuer-parameters).
+* `authorization_server_metadata`: OPTIONAL. The Authorization Server metadata object as defined by Section 2 of [@!RFC8414]. When provided the `authorization_server` parameter must not be present.
+
+The following is a non-normative example of OpenID4VCI Credential Offer over the DC API.
+
+
+```js
+navigator.credentials.create({
+  digital: {
+    requests: [
+      {
+        protocol: "openid4vci-v1",
+        data: {
+          credential_issuer: "...",
+          credential_configuration_ids: [...],
+          grants: {...}
+          credential_issuer_metadata: {...},
+          authorization_server_metadata: {...}
+        }
+      }
+    ]
+  }
+})
+```
+
+## Response {#dc_api_response}
+
+The Wallet must validate the Credential Offer and return an object with the following parameter:
+
+* `status`: REQUIRED. A string containing one of the status codes defined in this section.
+
+The following status codes are supported:
+
+`offer_accepted`:
+- The Wallet accepted the Credential Offer and will attempt to obtain the credential. It does not indicate the credential was successfully issued.
+
+
 # Credential Format Profiles {#format-profiles}
 
 This specification defines several extension points to accommodate the differences across Credential Formats. Sets of Credential Format-specific parameters or claims referred to as Credential Format Profiles are identified by the Credential Format Identifier and used at these extension points.


leecam:
Still work in progress, but can get some initial comments

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2052,6 +2052,83 @@ regulation), the Credential Issuer should properly authenticate the Wallet and e
...
+})
+```
+
+## Response {#dc_api_response}
+
+The Wallet must validate the Credential Offer and return an object with the following parameter:
+
+* `status`: REQUIRED. A string containing one of the status codes defined in this section.
+
+The following status codes are supported:
```
leecam, line 2126: TODO: Need to decide what error codes to support here.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2052,6 +2052,83 @@ regulation), the Credential Issuer should properly authenticate the Wallet and e
...
+
+The value `1` MUST be used for the `<version>` field to indicate the request and response are compatible with this version of the specification.
+
+The following exchange protocol values are defined by this specification:
+
+* `openid4vci-v1`
+
+## Credential Offer {#dc_api_credential_offer}
+
+The Issuer sends the Credential Offer object to the Digital Credentials API. All of the Credential Offer parameters MAY be used. The Credential Offer object has been extended to include the following parameters:
```
paulbastian, line 2092:
```suggestion
The Issuer sends the Credential Offer object to the Digital Credentials API. All of the Credential Offer parameters MAY be used. As some implementations of the DC API may not allow to network calls, the Credential Offer object is extended to include the following parameters:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2052,6 +2052,83 @@ regulation), the Credential Issuer should properly authenticate the Wallet and e
...
+
+
+* `credential_issuer_metadata`: REQUIRED. The Issuer's Credential Issuer Metadata object as defined in (#credential-issuer-parameters).
+* `authorization_server_metadata`: OPTIONAL. The Authorization Server metadata object as defined by Section 2 of [@!RFC8414]. When provided the `authorization_server` parameter must not be present.
```
paulbastian, line 2096:
```suggestion
* `authorization_server_metadata`: REQUIRED. The Authorization Server metadata object as defined in (#as-metadata).
```

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: probably offer_declined?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2052,6 +2052,83 @@ regulation), the Credential Issuer should properly authenticate the Wallet and e
...
+
+To use OpenID4VCI with the Digital Credentials API (DC API), the exchange protocol value has the following format: `openid4vci-v<version>`.
+
+The value `1` MUST be used for the `<version>` field to indicate the request and response are compatible with this version of the specification.
+
+The following exchange protocol values are defined by this specification:
+
+* `openid4vci-v1`
+
+## Credential Offer {#dc_api_credential_offer}
```
paulbastian, line 2090: I see potential for other requests, i.e. calling DC-API to continue a deferred issuance with `transaction_id`

paulbastian (changes requested):
Looks good so far, some minor suggestions

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2052,6 +2052,83 @@ regulation), the Credential Issuer should properly authenticate the Wallet and e
...
+
+The Issuer sends the Credential Offer object to the Digital Credentials API. All of the Credential Offer parameters MAY be used. The Credential Offer object has been extended to include the following parameters:
+
+
+* `credential_issuer_metadata`: REQUIRED. The Issuer's Credential Issuer Metadata object as defined in (#credential-issuer-parameters).
```
c2bo, line 2095: feels somewhat redundant?
```suggestion
* `credential_issuer_metadata`: REQUIRED. The Credential Issuer Metadata object as defined in (#credential-issuer-parameters).
```

openid-4-verifiable-credential-issuance-1_0.md
jogu: I think it's more "to allow all the installed wallets to figure out if they can handle the request without fetching external data that may result in data being leaked without user consent"? Not sure.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2052,6 +2052,83 @@ regulation), the Credential Issuer should properly authenticate the Wallet and e
...
+
+# OpenID4VCI over the Digital Credentials API {#dc_api}
+
+This section defines how to use OpenID4VCI with the Digital Credentials API.
+
+The name "Digital Credentials API" (DC API) encompasses the W3C Digital Credentials API [@!W3C.Digital_Credentials_API]
+as well as its native App Platform equivalents in operating systems (such as [Credential Manager on Android](https://developer.android.com/jetpack/androidx/releases/credentials)).
+The DC API allows web sites and native apps acting as Issuers to initiate the issuance of verifiable credentials.
+The API itself is agnostic to the Credential issuance protocol and can be used with different protocols.
+The Web Platform, working in conjunction with other layers, such as the app platform/operating system, and based on the permission of the End-User, will send the credential offer along with the Origin of the Issuer to the End-User's chosen Wallet.
```
Sakurann, line 2078:
```suggestion
The term "Digital Credentials API" (DC API) encompasses the W3C Digital Credentials API [@!W3C.Digital_Credentials_API]
as well as its native App Platform equivalents in operating systems (such as [Credential Manager on Android](https://developer.android.com/jetpack/androidx/releases/credentials)).
The API itself is agnostic to the Credential issuance protocol and can be used with different protocols.
When the DC API is used with this specification, it allows web sites and native apps acting as Issuers to initiate the issuance of verifiable credentials.
The Web Platform, working in conjunction with other layers, such as the app platform/operating system, and based on the permission of the End-User, will send the Credential Offer along with the Origin of the Issuer to the End-User's chosen Wallet.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2052,6 +2052,83 @@ regulation), the Credential Issuer should properly authenticate the Wallet and e
...
+
+## Protocol
+
+To use OpenID4VCI with the Digital Credentials API (DC API), the exchange protocol value has the following format: `openid4vci-v<version>`.
+
+The value `1` MUST be used for the `<version>` field to indicate the request and response are compatible with this version of the specification.
```
tplooker, line 2084: Do we need to be explicit here that the version essentially constitutes a MAJOR version as per semver and therefore implementations should adopt the robustness principle when it comes to things like ignoring unknown elements in maps/objects to support forward compatibility with minor changes?

openid-4-verifiable-credential-issuance-1_0.md
tplooker: Do you mean like using the DC API to notify the wallet that a deferred issuance is now ready for issuance?

openid-4-verifiable-credential-issuance-1_0.md
tplooker: I think thats probably right as its highlighting why some DC API implementations have this limitation.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2052,6 +2052,83 @@ regulation), the Credential Issuer should properly authenticate the Wallet and e
...
+
+## Response {#dc_api_response}
+
+The Wallet must validate the Credential Offer and return an object with the following parameter:
+
+* `status`: REQUIRED. A string containing one of the status codes defined in this section.
+
+The following status codes are supported:
+
+`offer_accepted`:
```
tlodderstedt, line 2128: Clarifying question: is the idea the issuer sends the offer and gets a confirmation from the wallet, but the actual issuance happens in a separate, parallel process within the wallet?

c2bo: That would be my understanding right now, the DC API flow basically succeeds with reaching a wallet

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: I would assume to ignore unknown parameters is the generally expected behavior? So why differentiating between minor and major?

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: As far as I understand the concept of this PR, the wallet receives the credential offer through DC API but interacts directly with the issuer afterwards. This would also include deferred issuance.

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: Perhaps we could be even more explicit like "As some implementations ..., the Credential Offer object is extended to allow to carry all necessary metadata parameters by value"

openid-4-verifiable-credential-issuance-1_0.md
tplooker: That was my understanding too @tlodderstedt but I think @paulbastian is saying perhaps there is potential to use the DC API to notify a wallet about the change in status to a deferred credential issuance flow, like the credential is now ready to be picked up, did I get that right @paulbastian? If so I think there is quite a bit of complexity to unpack there before we could support it.

openid-4-verifiable-credential-issuance-1_0.md
tplooker: @tlodderstedt to set the expectation that if this value is not a version supported by the wallet then it shouldn't attempt to process the request because it likely contains breaking changes e.g is considered a MAJOR version as per semver rules.

Sakurann:
not in scope for 1.0. will re-open if needed in 1.1

PR #478, opened 2025-04-01
Add DPoP Nonce to Nonce Endpoint

### LINKED ISSUE #412: Add DPoP-Nonce to the nonce endpoint
It may be beneficial to optionally return a DPoP-Nonce header along the Nonce Endpoint response, so that the Wallet saves a roundtrip & cryptographic operations to fetch a nonce for DPoP protected access tokens.

paulbastian:
Closes #412

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -778,7 +778,7 @@ The Credential Issuer provides a nonce value in the HTTP response with a 2xx sta
 
 Due to the temporal and contextually sensitive nature of the `c_nonce` value, the Credential Issuer MUST make the response uncacheable by adding a `Cache-Control` header field including the value `no-store`.
 
- The Credential Issuer MAY provide a DPoP nonce as an HTTP header as defined in Section 8.2 of [@!RFC9449]). In this case, the Wallet MUST use the new nonce value for the access token at the Credential Endpoint.
+ The Credential Issuer MAY provide a DPoP nonce as an HTTP header as defined in Section 8.2 of [@!RFC9449]). In this case, the Wallet MUST use the new nonce value in the DPoP proof when presenting an access token at the Credential Endpoint.
 
 Below is a non-normative example of a Nonce Response:
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -778,7 +778,7 @@ The Credential Issuer provides a nonce value in the HTTP response with a 2xx sta
 
 Due to the temporal and contextually sensitive nature of the `c_nonce` value, the Credential Issuer MUST make the response uncacheable by adding a `Cache-Control` header field including the value `no-store`.
 
- The Credential Issuer MAY provide a DPoP nonce as an HTTP header as defined in Section 8.2 of [@!RFC9449]). In this case, the Wallet MUST use the new nonce value in the DPoP proof when presenting an access token at the Credential Endpoint.
+ The Credential Issuer MAY provide a DPoP nonce as an HTTP header as defined in Section 8.2 of [@!RFC9449]). In this case, the Wallet uses the new nonce value in the DPoP proof when presenting an access token at the Credential Endpoint.
 
 Below is a non-normative example of a Nonce Response:
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -778,12 +778,15 @@ The Credential Issuer provides a nonce value in the HTTP response with a 2xx sta
 
 Due to the temporal and contextually sensitive nature of the `c_nonce` value, the Credential Issuer MUST make the response uncacheable by adding a `Cache-Control` header field including the value `no-store`.
 
+ The Credential Issuer MAY provide a DPoP nonce as an HTTP header as defined in Section 8.2 of [@!RFC9449]). In this case, the Wallet MUST use the new nonce value for the access token at the Credential Endpoint.
+
 Below is a non-normative example of a Nonce Response:
 
 ```
 HTTP/1.1 200 OK
 Content-Type: application/json
 Cache-Control: no-store
+DPoP-Nonce: eyJ7S_zG.eyJH0-Z.HX4w-7v
 
 {
   "c_nonce": "wKI4LT17ac15ES9bw8ac4"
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -778,12 +778,15 @@ The Credential Issuer provides a nonce value in the HTTP response with a 2xx sta
 
 Due to the temporal and contextually sensitive nature of the `c_nonce` value, the Credential Issuer MUST make the response uncacheable by adding a `Cache-Control` header field including the value `no-store`.
 
+ The Credential Issuer MAY provide a DPoP nonce as an HTTP header as defined in Section 8.2 of [@!RFC9449]). In this case, the Wallet MUST use the new nonce value for the access token at the Credential Endpoint.
```
bc-pi, line 781: the nonce goes in the proof along side the access token so this feels a little more accurate 

```suggestion
 The Credential Issuer MAY provide a DPoP nonce as an HTTP header as defined in Section 8.2 of [@!RFC9449]). In this case, the Wallet MUST use the new nonce value in the DPoP proof when presenting an access token at the Credential Endpoint.
```

bc-pi (approved):

c2bo (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -778,12 +778,15 @@ The Credential Issuer provides a nonce value in the HTTP response with a 2xx sta
 
 Due to the temporal and contextually sensitive nature of the `c_nonce` value, the Credential Issuer MUST make the response uncacheable by adding a `Cache-Control` header field including the value `no-store`.
 
+ The Credential Issuer MAY provide a DPoP nonce as an HTTP header as defined in Section 8.2 of [@!RFC9449]). In this case, the Wallet MUST use the new nonce value in the DPoP proof when presenting an access token at the Credential Endpoint.
```
jogu, line 781: I'm not 100% sure what the new normative behaviour we're trying to define here is.

Almost all of that sounds like it's standard DPoP behaviour, though I'm not 100% sure DPoP defines the behaviour when a DPoP-Nonce: header is returned when there was neither a dpop proof provided nor an error returned.

I'm also not clear if there's an implicit "if both the nonce endpoint and the credential endpoint are on the same server" that needs to be made more explicit as usually DPoP nonces are server specific ( https://datatracker.ietf.org/doc/html/rfc9449#section-9 ). The 'MUST' here could be read as requiring the client to use a nonce coming from the server hosting the nonce endpoint when it already had a valid one provided by the server hosting the credential endpoint, a situation that would actually make things worse.

paulbastian: The dpop standard is a little vague on the nonce handling and you may get different answers from different people on the nonce handling. Some people say you must always get a nonce through an http 400 error, therefore this PR offers a new mechanism to fetch a dpop nonce.

Furthermore, I would interpret "server" as in authorization server and credential issuer in our case, that means the wallet cannot rely on using this dpop nonce on the token endpoint unfortunately, it may work but it's issuer specific

jogu: > The dpop standard is a little vague on the nonce handling and you may get different answers from different people on the nonce handling. Some people say you must always get a nonce through an http 400 error, therefore this PR offers a new mechanism to fetch a dpop nonce.

Those people are wrong - https://datatracker.ietf.org/doc/html/rfc9449#section-8.2 seems clear to me that a new DPoP-Nonce can be issued in a successful response:

> A more efficient manner of supplying a new nonce value is also defined by including a DPoP-Nonce HTTP header in the HTTP 200 (OK) response from the previous request. The client MUST use the new nonce value supplied for the next token request and for all subsequent token requests until the authorization server supplies a new nonce.

> Furthermore, I would interpret "server" as in authorization server and credential issuer in our case, that means the wallet cannot rely on using this dpop nonce on the token endpoint unfortunately, it may work but it's issuer specific

I think it's really hard to justify any behaviour of associating DPoP-Nonce values with anything other than the hostname/port combination. Anything else feels like a layering violation to me. (But would be happy to get an opinion from the DPoP spec authors on that one :) )

bc-pi: > Almost all of that sounds like it's standard DPoP behaviour, though I'm not 100% sure DPoP defines the behaviour when a DPoP-Nonce: header is returned when there was neither a dpop proof provided nor an error returned.

It doesn't really define it specifically, which is what I understand this PR to be doing - just saying that if you get a DPoP-Nonce in your response here, use it with the credential endpoint request.

bc-pi: > I'm also not clear if there's an implicit "if both the nonce endpoint and the credential endpoint are on the same server" that needs to be made more explicit as usually DPoP nonces are server specific ( https://datatracker.ietf.org/doc/html/rfc9449#section-9 ). The 'MUST' here could be read as requiring the client to use a nonce coming from the server hosting the nonce endpoint when it already had a valid one provided by the server hosting the credential endpoint, a situation that would actually make things worse. (I can't really think why the nonce & credential endpoint would be hosted on different servers but it seems plausible it could happen.)

Does it really make things worse? A req with an invalid nonce vs one without a nonce... I suppose maybe in situations with the endpoints on different servers (which seems really unlikely but things happen) and repeated calls to the credential endpoint with the validity of the various nonce values getting out of sync, taking it as a MUST MUST ALWAYS could result in some inefficient churn. That seems like a very edge case. But what if the language was lightened just a bit? I think this would still allow/provide the benefits @paulbastian is looking to get while allaying @jogu's concerns: 

```suggestion
 The Credential Issuer MAY provide a DPoP nonce as an HTTP header as defined in Section 8.2 of [@!RFC9449]). In this case, the Wallet uses the new nonce value in the DPoP proof when presenting an access token at the Credential Endpoint.
```

jogu: Yeah, I can live with that, thanks Brian. I'd be tempted to add a 'at the same server' on the end but perhaps that could equally cause more confusion...

paulbastian: If the Issuer provides a nonce endpoint, the Wallet MUST call it anyway to get a `c_nonce`?

paulbastian: @jogu @bc-pi shall I commit your suggestion?

Sakurann:
WG discussion.
- resources potentially saved since one less round trip
- implementation feedback that this is easier

jogu (approved):

tplooker (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2790,6 +2793,7 @@ The technology described in this specification was made available from contribut
 
    -16
   
+   * Add DPoP Nonce to Nonce Endpoint
```
Sakurann, line 2796:
```suggestion
   * Adds an option to return DPoP Nonce from the Nonce Endpoint
```

PR #486, opened 2025-04-25
format list 

cre8:
Editorial change:

extra lines in the list are not rendered as a list when transformed to html (no problem with markdown)

Signed-off-by: Mirko Mollik <mirko.mollik@eudi.sprind.org>


cre8:
mmh, I see the error here:

https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-notification-request

but in the editors draft it is correct:

https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-notification-request

So it seems there was already some kind of adjustment, but to keep it aligned with the rest I think removing the empty lines would not be bad.

jogu:
I think the problem you see in the openid.net/specs version was fixed by Torsten last week:

https://github.com/openid/OpenID4VCI/pull/482/files

cre8:
> I think the problem you see in the openid.net/specs version was fixed by Torsten last week:
> 
> https://github.com/openid/OpenID4VCI/pull/482/files

Adding the empty line before the list started in line [1304](https://github.com/openid/OpenID4VCI/pull/482/files#diff-1f424614b35a9899813079f1b1f6218631a2aedd993368ccb89bb81a9eda0289R1304
) is correct and fixed it. However when you want to have it 100% correct, there should not be a space between list elements (like it's done with the other lists).

But up to you if you want to merge or close it :)

jogu (approved):

Issue #473, opened 2025-03-25
Presentation during Issuance


danielfett:
There is currently no sufficient guidance provided for doing a presentation of a credential during the issuance of another credential. In particular, there are specific advantages using the [OAuth for First-Party Applications draft](https://www.ietf.org/archive/id/draft-parecki-oauth-first-party-apps-02.html) for this.

Here's our proposal from the German EUDI wallet project: https://gitlab.opencode.de/bmi/eudi-wallet/eidas-2.0-architekturkonzept/-/blob/main/architecture-proposal/flows/Presentation-During-Issuance.md?ref_type=heads

I discussed with Torsten that VCI would probable be the right place to add this guidance.

TimoGlastra:
+1 from me

I was just thinking last week that it would be great if this got standardized in OID4VCI. We have implemented this in our wallet, and the flow was very reasonable to support, and provides great UX improvements.

Especially the fallback to PAR in the authorization challenge endpoint as defined in Firth Party Applications draft, means you can always use the authorization challenge endpoint if the authorization server supports it. If no presentation during issuance is required, the authorization challenge endpoint can just return the PAR response as if you'd done the PAR request directly.

There's now also an IETF draft: https://datatracker.ietf.org/doc/draft-ietf-oauth-first-party-apps/

Sakurann:
+1. there seems to be multiple implementers working on/interested in this in the EU and ISO communities

paulbastian:
I support this as well. The question to me is how much to rely on the OAuth for first party applications draft or whether to define something separate.

jogu:
Discussed on today's WG call:

- Restrict scope to presentation during issuance
- Have extension points so presenting ID cards etc can be defined elsewhere
- We need to think more about how this applies to DC API issuance/presentation
- This situations are mostly not going to be first party so probably we shouldn't reference the first party apps drafts (and it's not ready yet and it could be confusing for developers when they're implementing non-first party cases) so perhaps define something directly in VP

Lee suggested an alternative where a VP request is bundled into the DC API issuance call. Torsten said this has less flexibility than the above proposal based on first party apps draft, and we have implementation experience on first party apps approach (at least in non-DC API model). Plus we need a solution that works for wallet initiated issuance too.

Sakurann:
WG discussion
- this solves a real problem: want to do it in 1.0, not 1.1
    - there is quite some implementation experience (in cases without DC API)
- requirements: 
[...]

PR #487, opened 2025-05-01
Remove the Dynamic Credential Request section and associated content

### LINKED ISSUE #103: Should the parameter `wallet_issuer` be changed to `wallet_provider`?

### LINKED ISSUE #145: [AS] `wallet_issuer` and `user_hint` requires accurate explaination about their scopes and usefulness
My colleagues and I never really understood the parameters wallet_issuer and user_hint and this PR doesn't solve that either. I think the specification is missing text on this, but unsure if this should be a target of this PR

_Originally posted by @paulbastian in https://github.com/openid/OpenID4VCI/issues/142#issuecomment-1848953666_

bc-pi:
fixes #145 
closes #103


jogu (approved):

tplooker (approved):

paulbastian (approved):

Sakurann:
i would not be comfortable merging this before PR for #473 is merged - WDYT @danielfett @cre8

jogu:
> i would not be comfortable merging this before PR for #473 is merged - WDYT @danielfett @cre8

I don't think we need to create dependencies - what we have in this section as far as we could find out no one has implemented this, nor is it completely defined.

tplooker:
> I don't think we need to create dependencies - what we have in this section as far as we could find out no one has implemented this, nor is it completely defined.

Agree with this, dynamic credential requests isn't fully defined (arguable not implementable in its current state) so regardless of other issues that might have a tangential relation we should move forward with removing this text.

jogu:
Merging as per consensus on today's WG call.

PR #488, opened 2025-05-02
Various arrays can't sensibly be empty, add "non-empty" to definitons

jogu:
We already state that various arrays are optional or required, so in both cases if they're actually present they need to have at least one element. This avoids weird interoperability problems and means that conformance testing can be simpler as it removes a case we'd otherwise need to test for.

Similar to issue in VP: https://github.com/openid/OpenID4VP/issues/572


tplooker (approved):
LGTM, thanks.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -687,8 +687,8 @@ The Authorization Server might decide to authorize issuance of multiple instance
...
+* `authorization_details`: REQUIRED when the `authorization_details` parameter is used to request issuance of a Credential of a certain Credential Configuration as defined in (#authorization-details). OPTIONAL when `scope` parameter was used to request issuance of a Credential of a certain Credential Configuration. It is a non-empty array of objects, as defined in Section 7 of [@!RFC9396]. In addition to the parameters defined in (#authorization-details), this specification defines the following parameter to be used with the authorization details type `openid_credential` in the Token Response:
+  * `credential_identifiers`: REQUIRED. Non-empty array of strings, each uniquely identifying a Credential Dataset that can be issued using the Access Token returned in this response. Each of these Credential Datasets corresponds to the same Credential Configuration in the `credential_configurations_supported` parameter of the Credential Issuer metadata. The Wallet MUST use these identifiers together with an Access Token in subsequent Credential Requests. See (#identifying_credential) for the summary of the options how requested Credential(s) are identified throughout the Issuance flow.
```
TimoGlastra, line 691: Above you use "A non-empty",
For consistency

```suggestion
  * `credential_identifiers`: REQUIRED. A non-empty array of strings, each uniquely identifying a Credential Dataset that can be issued using the Access Token returned in this response. Each of these Credential Datasets corresponds to the same Credential Configuration in the `credential_configurations_supported` parameter of the Credential Issuer metadata. The Wallet MUST use these identifiers together with an Access Token in subsequent Credential Requests. See (#identifying_credential) for the summary of the options how requested Credential(s) are identified throughout the Issuance flow.
```

TimoGlastra (approved):

c2bo (approved):

PR #492, opened 2025-05-07
Move issuance pending from Deferred Credential Error Response to Deferred Credential Response and move interval parameter to Credential Response

### LINKED ISSUE #1: [ID-2] Credential Issuance: `retry_in` parameter for deferred credentials
### Original Reporter: bellebaum

Moved to here from Github.  
  
It seems unclear to me how often the client should retry to retrieve a token that could not yet be issued.  
Maybe an option for the server to specify when it expects to be able to deliver the requested credential would be helpful.  
This parameter \(let's call it `retry_in`\) would \(optionally or mandatory\) be returned with the `acceptance_token` at the credential endpoint, and at the deferred credential endpoint whenever the credential could not yet be delivered.

### LINKED ISSUE #226: Deferred Credential Error represents non-error cases
The [Deferred Credential Error Response](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-deferred-credential-error-r) is used for two primary cases: `issuance_pending` and `invalid_transaction_id`. Functionally, there is a distinction between:

1. A _valid request_ where the issuance is not yet completed (`issuance_pending`)
2. An _invalid request_ where the server cannot response (`invalid_transaction_id`)

The text points to [section 7.3.1.2.](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#credential-request-errors) which states:

> HTTP response MUST use the HTTP status code 400 (Bad Request) 

This is not an ideal practice, since in the case of (1) above (a valid request) we are requiring a 400 which signifies a Bad Request.

Instead it would be preferable to send a 200 with a response stating `issuance_pending`.

paulbastian:
Closes #1 
Closes #226 

- I moved the pending_issuance Deferred Credential Response Error to the regular Deferred Credential Response
- I moved the `interval` parameter from Deferred Credential Response Error to the Credential Response
- I changed the interval from OPTIONAL to REQUIRED (see thread discussion)
- I added an example of a Deferred Credential Response containg no Credentials but only new interval


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1269,9 +1276,8 @@ Content-Type: application/json
 
 When the Deferred Credential Request is invalid or the Credential is not available yet, the Credential Issuer constructs the error response as defined in (#credential-error-response).
```
TimoGlastra, line 1277:
```suggestion
When the Deferred Credential Request is invalid, the Credential Issuer constructs the error response as defined in (#credential-error-response).
```

charsleysa (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1109,6 +1109,7 @@ The following parameters are used in the JSON-encoded Credential Response body:
...
    * `credential`: REQUIRED. Contains one issued Credential. It MAY be a string or an object, depending on the Credential Format. See Appendix A for the Credential Format-specific encoding requirements.
 * `transaction_id`: OPTIONAL. String identifying a Deferred Issuance transaction. This parameter is contained in the response if the Credential Issuer cannot immediately issue the Credential. The value is subsequently used to obtain the respective Credential with the Deferred Credential Endpoint (see (#deferred-credential-issuance)). It MUST not be used if the `credentials` parameter is present. It MUST be invalidated after the Credential for which it was meant has been obtained by the Wallet.
+* `interval`: OPTIONAL. Contains a number that represents the minimum amount of time in seconds that the Wallet needs to wait before providing a new request to the Deferred Credential Endpoint. It MUST NOT be used if the `credentials` parameter is present. If not present, the Wallet MUST use `86400` as the default value.
```
tplooker, line 1112:
```suggestion
* `interval`: OPTIONAL. Contains a positive number that represents the minimum amount of time in seconds that the Wallet needs to wait before providing a new request to the Deferred Credential Endpoint. It MUST NOT be used if the `credentials` parameter is present. If not present, the Wallet MUST use `86400` as the default value.
```

tplooker: Are we comfortable with implying the default value of 1 day as the interval here, I would rather we make the interval required in the response so the implementation is always explicit about this interval.

paulbastian: I would be ok making it mandatory, anybody has objections?

c2bo: interval becomes kind of redundant if we have push notifications -> I think keeping it optional is the cleaner approach to allow for that

awoie: You would still need polling even with push, hence I think making it mandatory would be better to reduce/optimize server load.

awoie (commented):
generally fine with the direction but we should consider `interval` being mandatory.

openid-4-verifiable-credential-issuance-1_0.md
jogu: I think mandatory is good. Slightly simplifies the client code and slightly increases interoperability / reduces the number of variants that have to be conformance tested.

paulbastian: added the comment and made it mandatory

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1098,6 +1098,7 @@ The following parameters are used in the JSON-encoded Credential Response body:
...
    * `credential`: REQUIRED. Contains one issued Credential. It MAY be a string or an object, depending on the Credential Format. See Appendix A for the Credential Format-specific encoding requirements.
 * `transaction_id`: OPTIONAL. String identifying a Deferred Issuance transaction. This parameter is contained in the response if the Credential Issuer cannot immediately issue the Credential. The value is subsequently used to obtain the respective Credential with the Deferred Credential Endpoint (see (#deferred-credential-issuance)). It MUST not be used if the `credentials` parameter is present. It MUST be invalidated after the Credential for which it was meant has been obtained by the Wallet.
+* `interval`: REQUIRED if `transaction_id` is present. Contains a positive number that represents the minimum amount of time in seconds that the Wallet SHOULD wait after receiving the response before sending a new request to the Deferred Credential Endpoint. It MUST NOT be used if the `credentials` parameter is present.```
```
tplooker, line 1101:
```suggestion
* `interval`: REQUIRED if `transaction_id` is present. Contains a positive number that represents the minimum amount of time in seconds that the Wallet SHOULD wait after receiving the response before sending a new request to the Deferred Credential Endpoint. It MUST NOT be used if the `credentials` parameter is present.
```

jogu, line 1101: remove the ```:

```suggestion
* `interval`: REQUIRED if `transaction_id` is present. Contains a positive number that represents the minimum amount of time in seconds that the Wallet SHOULD wait after receiving the response before sending a new request to the Deferred Credential Endpoint. It MUST NOT be used if the `credentials` parameter is present.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1228,7 +1230,12 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 
 ## Deferred Credential Response {#deferred-credential-response}
 
-The Deferred Credential Response uses the `credentials` and `notification_id` parameters as defined in (#credential-response).
+Deferred Credential Response may either contain the requested Credentials or further defer the issuance:
+
+* If the Credential Issuer is able to issue the requested Credentials, the Deferred Credential Response MUST use the `credentials` parameter as defined in (#credential-response) and MUST respond with the HTTP status code 200 (see Section 15.3.3 of [@!RFC9110]).
+* If the Credential Issuer still requires more time, the Deferred Credential Response MAY use the `interval` parameter as defined in (#credential-response) and MUST respond with the HTTP status code 202 (see Section 15.3.3 of [@!RFC9110]).
+
+The Deferred Credential Response MAY use the `notification_id` parameter as defined in (#credential-response).
```
jogu, line 1238: I think this should be in the "is able to issue" paragraph?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1228,7 +1230,12 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 
 ## Deferred Credential Response {#deferred-credential-response}
 
-The Deferred Credential Response uses the `credentials` and `notification_id` parameters as defined in (#credential-response).
+Deferred Credential Response may either contain the requested Credentials or further defer the issuance:
+
+* If the Credential Issuer is able to issue the requested Credentials, the Deferred Credential Response MUST use the `credentials` parameter as defined in (#credential-response) and MUST respond with the HTTP status code 200 (see Section 15.3.3 of [@!RFC9110]).
+* If the Credential Issuer still requires more time, the Deferred Credential Response MAY use the `interval` parameter as defined in (#credential-response) and MUST respond with the HTTP status code 202 (see Section 15.3.3 of [@!RFC9110]).
```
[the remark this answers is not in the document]
jogu: I think we're significantly under defining the "still requires more time" response now, it must include interval right?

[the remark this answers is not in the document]
paulbastian: yeah, I caught the same in the meantime :) However my suggestion was to require `interval` in the new response. WDYT?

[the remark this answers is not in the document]
jogu: @paulbastian I'm good with that.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1228,14 +1230,19 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 
 ## Deferred Credential Response {#deferred-credential-response}
 
-The Deferred Credential Response uses the `credentials` and `notification_id` parameters as defined in (#credential-response).
+A Deferred Credential Response may either contain the requested Credentials or further defer the issuance:
```
sloops77, line 1233: The response can also indicate that there are no credentials for the holder (for a variety of reasons) and that the wallet should not poll any longer. I think it's worth adding normative description of what this response looks like 

- 204
- 200 with an empty response
- 200 & and empty `credentials` array?

sloops77:
@paulbastian This PR brings even closer alignment between the credential and deferred endpoints. I think WG should reconsider merging them. 

- It would mean `transaction_id` is acceptable on the credential request, along with the corresponding extra error response. 
- Additionally is there a benefit in allowing `proofs` to be resent on each request?

paulbastian:
> @paulbastian This PR brings even closer alignment between the credential and deferred endpoints. I think WG should reconsider merging them. 
> 
> - It would mean `transaction_id` is acceptable on the credential request, along with the corresponding extra error response. 
> - Additionally is there a benefit in allowing `proofs` to be resent on each request?

Yes. I'm pushing in the same direction. Consider it a preparation step.

openid-4-verifiable-credential-issuance-1_0.md
jogu: I think this is out of scope for this PR. I'll convert it into a new issue instead.

PR #495, opened 2025-05-08
fix signing alg values supported for ISO mdos in metadata

### LINKED ISSUE #303: Consider using COSE algorithm identifiers for credential_signing_alg_values_supported for mso_mdoc

### LINKED ISSUE #354: Make `credential_signing_alg_values_supported` and `proof_signing_alg_values_supported` polymorphic or credential format-specific

awoie:
This PR fixes signing alg values supported for ISO mdos in metadata. Aligned with OID4VP language.

Fixes #303; fixes #354

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2214,6 +2214,10 @@ Cryptographic algorithm identifiers used in the `credential_signing_alg_values_s
 * The value exactly matches the `alg` value in the `IssuerAuth` COSE header;
 * Or, the value is a fully specified algorithm, as defined in [@!I-D.ietf-jose-fully-specified-algorithms], and the combination of the `alg` value and the signing key's curve in the `IssuerAuth` COSE structure matches the combination specified by the fully specified algorithm.
 
+Consequently, depending on the approach taken by the Issuer as described above, the `alg` value in the `IssuerAuth` COSE header MAY NOT directly correspond to the identifier listed in the `credential_signing_alg_values_supported` metadata parameter. This metadata paramter in is primarily informational for the Wallet.
+
+Note that these `alg` values are informational to the wallet and depending on the approach the issuers chose, the `alg` value in the `IssueAuth` COSE header MAY NOT correspond to the identifier in the `credential_signing_alg_values_supported` metadta parameter.
+
 Example:
 If the `IssuerAuth` structure contains an `alg` header value of `-7` (ECDSA with SHA-256, per [@!IANA.COSE]) and is signed using a P-256 key, it matches both `-7` and `-9` in `credential_signing_alg_values_supported`. The latter (`-9`) corresponds to ECDSA with P-256 and SHA-256, as defined in [@!I-D.ietf-jose-fully-specified-algorithms].
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2814,7 +2814,8 @@ The technology described in this specification was made available from contribut
    * add missing request for media type registration of key-attestation+jwt in IANA Considerations
    * rename keyattestation+jwt to key-attestation+jwt
    * make type and values for credential_signing_alg_values_supported format specific
-   * change algorithm identifiers for credential_signing_alg_values_supported to COSE algorithm values for mdocs   
+   * make type and values for proof_signing_alg_values_supported proof type specific
+   * change algorithm identifiers for credential_signing_alg_values_supported to COSE algorithm values for mdocs
 
    -15
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -943,7 +943,7 @@ The Credential Issuer MUST validate that the JWT used as a proof is actually sig
 
 The Credential Issuer SHOULD issue a Credential for each cryptographic public key specified in the `attested_keys` claim within the `key_attestation` parameter.
 
-Cryptographic algorithm identifiers used in the `proof_signing_alg_values_supported` Credential Issuer metadata parameter for this proof type are case sensitive strings and SHOULD be one of those defined in [@IANA.JOSE].
+Cryptographic algorithm identifiers used in the `proof_signing_alg_values_supported` Credential Issuer metadata parameter for this proof type are case sensitive strings and SHOULD be one of the Algorithm Names defined in [@IANA.JOSE].
 If Credential Issuer metadata is provided, the `alg` JWT header of the key proof, and if present, the `alg` JOSE headers of both `key_attestation` and `trust_chain`, MUST match one of the values listed in the `proof_signing_alg_values_supported` metadata parameter.
 
 Below is a non-normative example of a `proofs` parameter (with line breaks within values for display purposes only):
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2200,7 +2200,13 @@ The Credential Format Identifier is `mso_mdoc`. This refers to the Mobile Securi
 
 ### Credential Issuer Metadata {#server-metadata-mso-mdoc}
 
-Cryptographic algorithm names used in the `credential_signing_alg_values_supported` parameter SHOULD be one of those defined in [@!ISO.18013-5].
+Cryptographic algorithm names used in the `credential_signing_alg_values_supported` parameter correspond to the algorithm identifiers used to secure the COSE IssuerAuth structure, as defined in [@!ISO.18013-5]. The values SHOULD be chosen from the following:
+
+* The value exactly matches the `alg` value in the IssuerAuth COSE header.
+* The value is a fully specified algorithm, as defined in [I-D.ietf-jose-fully-specified-algorithms], and the combination of the `alg` value and the signing key's curve in the `IssuerAuth` COSE structure matches the combination specified by the fully specified algorithm.
+
+Example:
+If the `IssuerAuth` structure contains an `alg` header value of `-7` (ECDSA with SHA-256, per [IANA.COSE]) and is signed using a P-256 key, it matches both `-7` and `-9` in `credential_signing_alg_values_supported`. The latter (`-9`) corresponds to ECDSA with P-256 and SHA-256, as defined in [I-D.ietf-jose-fully-specified-algorithms].
 
 The following additional Credential Issuer metadata parameters are defined for this Credential Format for use in the `credential_configurations_supported` parameter, in addition to those defined in (#credential-issuer-parameters).
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2211,13 +2211,13 @@ The Credential Format Identifier is `mso_mdoc`. This refers to the Mobile Securi
 
 ### Credential Issuer Metadata {#server-metadata-mso-mdoc}
 
-Cryptographic algorithm identifiers used in the `credential_signing_alg_values_supported` parameter correspond to the numeric algorithm identifiers used to secure the COSE `IssuerAuth` structure, as defined in [@!ISO.18013-5]. The values SHOULD be chosen from the following:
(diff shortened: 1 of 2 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2200,7 +2200,13 @@ The Credential Format Identifier is `mso_mdoc`. This refers to the Mobile Securi
 
 ### Credential Issuer Metadata {#server-metadata-mso-mdoc}
 
-Cryptographic algorithm names used in the `credential_signing_alg_values_supported` parameter SHOULD be one of those defined in [@!ISO.18013-5].
+Cryptographic algorithm names used in the `credential_signing_alg_values_supported` parameter correspond to the algorithm identifiers used to secure the COSE `IssuerAuth` structure, as defined in [@!ISO.18013-5]. The values SHOULD be chosen from the following:
```
tplooker, line 2203:
```suggestion
Cryptographic algorithm identifiers used in the `credential_signing_alg_values_supported` parameter correspond to the algorithm identifiers used to secure the COSE `IssuerAuth` structure, as defined in [@!ISO.18013-5]. The values SHOULD be chosen from the following:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2200,7 +2200,13 @@ The Credential Format Identifier is `mso_mdoc`. This refers to the Mobile Securi
...
+
+* The value exactly matches the `alg` value in the IssuerAuth COSE header.
+* The value is a fully specified algorithm, as defined in [I-D.ietf-jose-fully-specified-algorithms], and the combination of the `alg` value and the signing key's curve in the `IssuerAuth` COSE structure matches the combination specified by the fully specified algorithm.
```
tplooker, line 2206: As I stated in this PR https://github.com/openid/OpenID4VP/pull/553 im concerned about the complexity of this logic, rather than just having a `credential_signing_crv_values_supported` parameter which would mean the wallet nor the issuer would need to maintain mapping logic here as there would always be an explicit match.

tplooker (changes requested):
Not an objection to this PR outright but I've tried to express my concerns in https://github.com/openid/OpenID4VP/pull/553 which is related and I've not had any clarification to that, so until we do I'm going to request changes here.

openid-4-verifiable-credential-issuance-1_0.md
awoie: While I agree that a `_crv` parameter would be slightly less complex, the DCP WG decided not to introduce a `_crv` parameter for OID4VP, and opted for consistency by making the same decision for OID4VCI.

awoie: I think we should actually require fully specified algorithm for this, otherwise the wallet does not know whether the curve is supported before getting the credential.

paulbastian (approved):

openid-4-verifiable-credential-issuance-1_0.md
awoie: Discussed in the WG and it makes no sense to require fully specified algorithm only.

awoie:
@martijnharing could you please review.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2200,7 +2211,13 @@ The Credential Format Identifier is `mso_mdoc`. This refers to the Mobile Securi
 
 ### Credential Issuer Metadata {#server-metadata-mso-mdoc}
 
-Cryptographic algorithm names used in the `credential_signing_alg_values_supported` parameter SHOULD be one of those defined in [@!ISO.18013-5].
+Cryptographic algorithm identifiers used in the `credential_signing_alg_values_supported` parameter correspond to the numeric COSE algorithm identifiers used to secure the `IssuerAuth` COSE structure, as defined in [@!ISO.18013-5]. The values SHOULD be chosen from the following:
+
+* The value exactly matches the `alg` value in the `IssuerAuth` COSE header;
+* Or, the value is a fully specified algorithm, as defined in [@!I-D.ietf-jose-fully-specified-algorithms], and the combination of the `alg` value and the signing key's curve in the `IssuerAuth` COSE structure matches the combination specified by the fully specified algorithm.
```
tplooker, line 2217: I'm still concerned (and this applies to OpenID4VP too) that we are undermining referencing the COSE IANA registry with this language, which should be the primary source of algorithm identifiers. For example if guidance were to change with the IANA registry again in future, this language would make it super complicated to understand what an implementation should do.

martijnharing: I tend to agree, (especially in VCI) I think there is a reasonable chance that we are aligning things at the cost of ambiguity. An alternative would be to use arrays of crv/alg parameters, which doesn't have any risk of being misinterpreted.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2200,7 +2209,13 @@ The Credential Format Identifier is `mso_mdoc`. This refers to the Mobile Securi
...
+
+Example:
+If the `IssuerAuth` structure contains an `alg` header value of `-7` (ECDSA with SHA-256, per [@!IANA.COSE]) and is signed using a P-256 key, it matches both `-7` and `-9` in `credential_signing_alg_values_supported`. The latter (`-9`) corresponds to ECDSA with P-256 and SHA-256, as defined in [@!I-D.ietf-jose-fully-specified-algorithms].
```
martijnharing, line 2218: But this means that there is no mechanism for the issuer to say:
- I support issuance of mdocs with -9, but not issuance of -7.
This can be important to know, since receiving an mdoc with -9 may not work for the wallet, while -7 does.

awoie: @martijnharing Yes, that is not possible today. I would argue that even if we split the parameter into dedicated alg/crv parameters, the wallet would either receive -7 or -9 depending on the issuer's choice. What solution would you suggest instead?

selfissued (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -960,7 +960,7 @@ The Credential Issuer MUST validate that the JWT used as a proof is actually sig
 
 The Credential Issuer SHOULD issue a Credential for each cryptographic public key specified in the `attested_keys` claim within the `key_attestation` parameter.
 
-Cryptographic algorithm names used in the `proof_signing_alg_values_supported` Credential Issuer metadata parameter for this proof type SHOULD be one of those defined in [@IANA.JOSE].
+Cryptographic algorithm identifiers used in the `proof_signing_alg_values_supported` Credential Issuer metadata parameter for this proof type are case sensitive strings and SHOULD be one of those defined in [@IANA.JOSE].
```
jogu: https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms describes them as "Algorithm Name", so just to be 100% clear here I think we could say:

```suggestion
Cryptographic algorithm identifiers used in the `proof_signing_alg_values_supported` Credential Issuer metadata parameter for this proof type are case sensitive strings and SHOULD be one of the Algorithm Names defined in [@IANA.JOSE].
```

jogu (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2123,7 +2131,7 @@ When the `format` value is `jwt_vc_json`, the entire Credential Offer, Authoriza
 
 #### Credential Issuer Metadata {#server-metadata-jwt-vc-json}
 
-Cryptographic algorithm names used in the `credential_signing_alg_values_supported` parameter SHOULD be one of those defined in [@IANA.JOSE].
+Cryptographic algorithm identifiers used in the `credential_signing_alg_values_supported` parameter are case sensitive strings and SHOULD be one of those JWS algorithm names defined in [@IANA.JOSE].
```
awoie, line 2134:
```suggestion
Cryptographic algorithm identifiers used in the `credential_signing_alg_values_supported` parameter are case sensitive strings and SHOULD be one of those JWS Algorithm Names defined in [@IANA.JOSE].
```

```suggestion
Cryptographic algorithm identifiers used in the `credential_signing_alg_values_supported` parameter are case sensitive strings and SHOULD be one of those JWS Algorithm Names defined in [@IANA.JOSE].
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2263,7 +2279,7 @@ The Credential Format Identifier is `dc+sd-jwt`.
 
 ### Credential Issuer Metadata {#server-metadata-sd-jwt-vc}
 
-Cryptographic algorithm names used in the `credential_signing_alg_values_supported` parameter SHOULD be one of those defined in [@IANA.JOSE].
+Cryptographic algorithm identifiers used in the `credential_signing_alg_values_supported` parameter are case sensitive strings and SHOULD be one of those JWS algorithm names defined in [@IANA.JOSE].
```
```suggestion
Cryptographic algorithm identifiers used in the `credential_signing_alg_values_supported` parameter are case sensitive strings and SHOULD be one of those JWS Algorithm Names defined in [@IANA.JOSE].
```

openid-4-verifiable-credential-issuance-1_0.md
awoie: I added the suggestion you raised in the last DCP WG call, so I believe this should be addressed.

openid-4-verifiable-credential-issuance-1_0.md
awoie: I added the suggestion raised in the last DCP WG call, so I believe this should be addressed.

openid-4-verifiable-credential-issuance-1_0.md
awoie: I added the suggestion raised in the last DCP WG call, so I believe this should be addressed.

openid-4-verifiable-credential-issuance-1_0.md
martijnharing: “MAY NOT” could be interpreted as “is not allowed to”. Maybe something like “therefore xx MAY be different”
If this use of MAY NOT is common, no changes necessary.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2228,7 +2236,15 @@ The Credential Format Identifier is `mso_mdoc`. This refers to the Mobile Securi
 
 ### Credential Issuer Metadata {#server-metadata-mso-mdoc}
 
-Cryptographic algorithm names used in the `credential_signing_alg_values_supported` parameter SHOULD be one of those defined in [@!ISO.18013-5].
+Cryptographic algorithm identifiers used in the `credential_signing_alg_values_supported` parameter correspond to the numeric COSE algorithm identifiers used to secure the `IssuerAuth` COSE structure, as defined in [@!ISO.18013-5]. The values SHOULD be chosen from the following:
+
+* The value exactly matches the `alg` value in the `IssuerAuth` COSE header;
+* Or, the value is a fully specified algorithm, as defined in [@!I-D.ietf-jose-fully-specified-algorithms], and the combination of the `alg` value and the signing key's curve in the `IssuerAuth` COSE structure matches the combination specified by the fully specified algorithm.
+
+Consequently, depending on the approach taken by the Issuer as described above, the `alg` value in the `IssuerAuth` COSE header MAY NOT directly correspond to the identifier listed in the `credential_signing_alg_values_supported` metadata parameter. This metadata parameter is primarily informational for the Wallet.
```
```suggestion
Consequently, depending on the approach taken by the Issuer as described above, the `alg` value in the `IssuerAuth` COSE header MAY differ from the identifier listed in the `credential_signing_alg_values_supported` metadata parameter. This metadata parameter is primarily informational for the Wallet.
```

Issue #421, opened 2024-11-26
Add display metadata to the CredentialResponse


leecam:
When receiving a credential from the Credential Endpoint we'd like an Issuer to be able to provide optional "card art" and a description.

For example an CredentialResponse maybe extended as follows:
```
{
    "display": [
        {
             "locale": "en-US",
             "name": "Lee's Tribank Card"
             "description": "5423 4343 2334 XXXX"
             "logo": {
                    "uri": "data:image/png;base64,F00==",
                    "alt_text": "credit card image"
             },
        }
    ]
    "credentials": [
    {
      "credential": "LUpixVCWJk0eOt4CXQe1NXK....WZwmhmn9OQp6YxX0a2L"
    }
  ]
}
```

This allows the wallet to display the credential with personalized cardart and descriptions in a standardized way, regardless of credential format or doctype. See this screenshot for an example of customized logo, title and description. 

![image](https://github.com/user-attachments/assets/cc7fc3ec-4461-4141-a850-79e36cf0d2f2)

c2bo:
That would mean that display information inside the credential response would overrule the generic display information inside the issuer metadata if I understand correctly?

javereec:
Good idea. Some thoughts:

- It very likely that this information would be useful beyond this point (display in wallet, in a presentation flow). Doing it this way would probably mean that this information would be retained (... and passed on?)
- If this is just for this interaction, I feel there needs to be a link between this and the format specific display metadata (from the issuer metadata).

nemqe:
In general not against the idea but have some questions/thoughts:
- Would we do full override where we do not even look at the structure in the issuer medatada, or would we do partial overlays where we merge the returned structure with the one in the issuer medatada (we only replace the background image but logo should be the default one)
- Would we deliver the full set of display data with regards to the locale or just parts. If we deliver the 'en-US' locale and the user changes the locale on their device what would be the cardart displayed to the user, and could the user request the cardart for the new locale, or would we always send all locales to the wallet. Even in the case where we send all of the currently known locales how would the user/wallet fetch a new locale that has been added (edge case)
- Would we need mechanisms to update this personalized cardart in case of mistakes or changes in branding

timbl-ont:
Another option is to include attribute values, from the credential, to overlay on top of the card view in the credential display metadata. The background_image could be an SVG with parameterized values, if precise layout was required.

fkj:
After doing pilot testing we have run into a similar problem with the UX.
[...]

PR #500, opened 2025-05-15
decouple credential metadata

### LINKED ISSUE #99: decoupling issuer metadata from the supported credentials metadata
bringing the conversation that started in SD-JWT VC https://github.com/oauth-wg/oauth-sd-jwt-vc/issues/181 to VCI, because potentially moving `credentials_supported` part out of issuer metadata has implications and is related to some of the discussions in issue #82. Let's keep the conversation on this topic in SD-JWT VC issue tracker and not this one.

c2bo:
Closes #99 

This is a bit different from what we discussed at last WG, thus marking it as draft for the time being.


c2bo:
TODO: change all the examples

Sakurann:
WG discussion:
- direction seems to make sense.

sloops77:
@c2bo I'm not a fan of adding a `credential_metadata` claim. It's too generic sounding and makes me wonder why many other credential configuration isnt in there

Some options to consider:
1. rename the property to `display_metadata`, `display_descriptor`. Is there any purpose for this property than for wallet display?
2. `type_metadata` is an alternative to the above and shows alignment with sd-jwt-vc type metadata
3. nest `claims` inside of `display`, and update the sd-jwt-vc spec

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1458,17 +1458,19 @@ This specification defines the following Credential Issuer Metadata parameters:
...
       * `key_storage`: OPTIONAL. Array defining values specified in (#keyattestation-apr) accepted by the Credential Issuer.
       * `user_authentication`: OPTIONAL. Array defining values specified in (#keyattestation-apr) accepted by the Credential Issuer.
-  * `display`: OPTIONAL. A non-empty array of objects, where each object contains the display properties of the supported Credential for a certain language. Below is a non-exhaustive list of parameters that MAY be included.
+  * `credential_metadata`: OPTIONAL. Object containing information relevant to the usage and display of issued Credentials. Credential Format-specific mechanisms can overwrite the information in this object to convey Credential metadata. Format-specific mechanisms are always preferred over the information in this object, which serves as the default fallback. Below is a non-exhaustive list of parameters that MAY be included:
```
Sakurann: i think we need to add specific example of sd-jwt vc display metadata. and be explicit for whom this is a fallback etc (ie wallet). btw talking to some EUDI implementers, it sounds like they would prefer using the same issuer metadata display for both mdoc and sd-jwt...

```suggestion
  * `credential_metadata`: OPTIONAL. Object containing information relevant to the usage and display of issued Credentials. Credential Format-specific mechanisms can overwrite the information in this object to convey Credential metadata. Format-specific mechanisms, such as SD-JWT VC display metadata are always preferred by the Wallet over the information in this object, which serves as the default fallback. Below is a non-exhaustive list of parameters that MAY be included:
```

charsleysa (approved):
The `credential_metadata` name still feels weird but I don't have a strong enough opinion on it for me to not approve this PR.

jogu (approved):

leecam (approved):

lchacha (approved):

c2bo:
> * this change feels quite sd-jwt specific, and I don't think it is clear why properties of the credential configuration such as `format` and `credential_definition` are defined outside of a property called `credential_metadata`, but `display` and `claims` are in there unless one is familiar with the sd-jwt spec.

My rationale was that things like `format` and `credential_definition` are necessary for choices the wallet makes during the issuance process/flow, whereas everything in `credential_metadata` is about longterm information about the credential, mainly for display purposes right now.

It would imho also be a clear path forward for #421 to allow for another display override via the credential response (which would then also be scoped on this object).

> * I think defining override mechanisms is really worthwhile, but it could encompass more than this `credential_metadata`. Our ecosystem uses w3c VCs and has a multitude of issuers all whom would benfit from delegating a default set of credential configurations to an ecosystem endpoint, and allow issuers to optionally override these default settings. This kind of override would be perhaps accepting a particular proof type or using a particular signing algs, or it could even be a whole credential configuration unique to the issuer.

The idea was to initially define the override for information that is not critical to the issuance process itself (like format, signing algs etc.), but information like display that in itself does not create an interop problem and the wallet could always fall back to the "default" information in the metadata.

An override mechanism that modifies the issuance flow would have a lot more implications and we need to be really careful to not create interop issues here.

PR #501, opened 2025-05-15
set nonce in key attestation to c_nonce value for issuance

### LINKED ISSUE #438: clarify that nonce in key attestation is c_nonce
in this line https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#appendix-D.1-4.1.2.7

awoie:
This PR fixes #438. 

The language was added to the more specific proof type section since the key attestation annex might be referenced by other specs that may or may not use c_nonce. If it is unlikely that other specs would borrow the key attestation annex, I could change the PR, so the c_nonce text is added to the annex instead of the proof types section.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -933,6 +933,8 @@ There are two ways to convey key attestations (as defined in (#keyattestation)) 
 - The Wallet uses the `jwt` proof type in the Credential Request to create a proof of possession of the key and adds the key attestation in the JOSE header.
 - The Wallet uses the `attestation` proof type in the Credential Request with the key attestation without a proof of possession of the key itself.
 
+In both cases, the `nonce` value in the key attestation is set to the `c_nonce` value provided by the Issuer.
+
 Depending on the Wallet's implementation, the `attestation` may avoid unnecessary End-User interaction during Credential issuance, as the key itself does not necessarily need to perform signature operations.
 
 Additional proof types MAY be defined and used.
@@ -2796,6 +2798,7 @@ The technology described in this specification was made available from contribut
    * deprecate the proof paramter in the credential request
    * add missing request for media type registration of key-attestation+jwt in IANA Considerations
    * rename keyattestation+jwt to key-attestation+jwt
+   * set key attestation nonce to c_nonce value for proof types with key attestations
 
    -15
 


c2bo (approved):

tplooker (approved):

tlodderstedt (approved):

paulbastian (changes requested):
We already have: https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-8.2.1.3-2

So for key_attestation proof type, this seems to be good already. For jwt proof type including key_attestation, it is more vague, but there the nonce is also not mandatory. If we want to fix this, I would rather put it here: https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-8.2.1.1-2.1.2.6

jogu (approved):

Sakurann:
> We already have: https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-8.2.1.3-2
> 
> So for key_attestation proof type, this seems to be good already. For jwt proof type including key_attestation, it is more vague, but there the nonce is also not mandatory. If we want to fix this, I would rather put it here: https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-8.2.1.1-2.1.2.6

I would rather remove the text in https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-8.2.1.3-2 and keep Oliver's current text.

awoie:
> > We already have: https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-8.2.1.3-2
> > So for key_attestation proof type, this seems to be good already. For jwt proof type including key_attestation, it is more vague, but there the nonce is also not mandatory. If we want to fix this, I would rather put it here: https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-8.2.1.1-2.1.2.6
> 
> @paulbastian I would rather remove the text in https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-8.2.1.3-2 and keep Oliver's current text.
> 
> we basically need clarity on the nonce in the following scenarios:
> 
> * jwt proof **without** key_atterstation
> * jwt proof **with** key_atterstation
> * key attestation proof type
> 
> and i don't think current spec text is clear enough

I think I agree with @paulbastian and I implemented the change in the PR.

With this PR, the following should be clear:
> * jwt proof **without** key_atterstation
> * jwt proof **with** key_atterstation
> * key attestation proof type

Issue #339, opened 2024-05-31
Support Credential Request application-layer encryption


GarethCOliver:
While not relevant when making a direct request form a Wallet to an Issuer endpoint, when going through Wallet Server between the Wallet and the Issuer, application-layer encryption provides a positive privacy benefit to the user. 

In particular, for a credential format like mdoc, if the Wallet Server sees the device keys being sent to the issuer, they are able to collude with RPs to track all presentations of the issued credentials. If the credential request is instead encrypted, (and using single-use MSOs) this sort of tracking would only be possible through issue-RP collusion.

jogu:
I'm not sure if we explicitly say it, but I think there is a current assumption in the spec that calls to the issuer endpoint are made from the wallet app.

I guess the first thing we need to agree on is why there may be a need for these calls to go through a wallet's backend?

GarethCOliver:
Oh, sure. From prior conversation with others on the spec my understanding was that OpenID4VCI was intended to support both going through some wallet sever infrastructure as well as directly connecting to the issuer. Clearly, there are privacy implications to doing so, and that e2ee was the intended mechanism for closing some of these gap (and why it is available for credential response). 

At its heart the reason for going through a wallet server infrastructure is reliability and usability at scale. Functionally, it should be assumed that any Wallet application can never be updated, as in practice users update slowly and inconsistently. Similarly any analytics/monitoring is slow and unreliable, making detection of issues challenging. 

Wallet server infrastructure does not face these issues, and allows for working around client version skew issues that would simply be infeasible for issuers to deal with. 

My assumption here is that the Wallet in question is distributed to millions or billions of devices, and is intending to contact 100s of different issuers. Similarly, Issuers would assume to interact with multiple wallets that they do not own, control or understand.

If you'd like, I can open a different issue to  ask if using wallet server infrastructure is a valid way to use OpenId4VCI? For me to want to use this would be a hard requirement.

jogu:
Thanks Gareth! I think that's probably fair - I'm not sure it's necessarily intended to support this architecture but it's certainly not prevented. We don't have any privacy considerations for that kind of operation so potentially we might need to add them.

The credential response encryption unfortunately isn't great at the moment as currently the wallet backend could gain access to the encrypted data (unless there's extra checks in the issuer on the key used for encryption), but adding the kind of encryption you suggest in this issue might solve that some of that issue.

peppelinux:
[...]

Issue #502, opened 2025-05-15
Partial response encryption


martijnharing:
As part of the discussion of https://github.com/openid/OpenID4VCI/issues/339 we discussed whether there is a need partial response encryption and to have a separate discussion item for that to give reasons for having partial response encryption.

Conceptually partial response encryption is important when there are different components involved in provisioning that don't all have access to the key used for encryption, but they need access to part of the response to provide functionality. Examples of such situations are:
- Within the mobile device there are areas on the mobile device that not all have access to the key used for response encryption, but they may need access to part of the response.
- The mobile device has secure areas outside of the mobile device that contains the key used for response encryption, but components on the mobile device need access to part of the response.
- Issuance is mediated by a wallet server, where the wallet server needs access to part of the response but doesn't have access to the key used for response encryption.

c2bo:
I've been wondering since this was brought up: Do you think there will some form of generally applicable rule that fits all of those use-cases which part would be encrypted by which party or would that need to be fully dynamic and the wallet and whatever system in-between need to figure that part out?

Fully dynamic sounds a lot more complex and probably also more prone to errors?

Sakurann:
i think @GarethCOliver also put some thoughts in this issue thread https://github.com/openid/OpenID4VCI/issues/339#issuecomment-2877238722

Sakurann:
WG discussion:
- clarification: only applies to credential response
- what needs to be non-encrypted is outside credentials parameter; ie what does not need ot be encrypted is not credentials parameter (@martijnharing please confirm)
- overall concern that this adds a lot of complexity..
    - isn't this wallet/use-case specific
    - probably can't standardize which parameters to encrypt and which don't
- any specific technical solutions?
    - please think of some

tlodderstedt:
I don't feel we have enough information to properly design a solution. 

I'm concerned this feature will make the spec much more complex and I don't fully understand the benefit yet. I think we need to understand the use cases, especially what data shall not be encrypted.

tplooker:
Conceptually I understand the desire here but agree with @tlodderstedt it feels a bit premature to design a solution unless we have some clear usecases we can model the feature against.

Sakurann:
I think we can add this feature in a non-breaking manner in 1.1 if needed?

jogu:
Moved to 1.1 as per above comments. Please continue to discuss and add the details of the use cases / requirements.

PR #505, opened 2025-05-16
Add support for request encryption

GarethCOliver:
Resolves Issue #339

The adds in a new credential issuer metadata parameter to support Credential Request encryption. 

Merging this with the existing credential response encryption was considered, but given we allow all permutations of mandatory/optional/not supported for both request and response independently, there isn't a great way to do this. 

If we required issuers to be consistent on support or not then these would be cleaner as single parameter (but perhaps something we should leave to HAIP). 

To be consistent with HAIP, `alg` is now required in the `jwk`, a common section on how to perform selection is included and `kid` must be included if available. 

The removal of alg is a breaking change but IMO is worth while to be consistent across the specs.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -806,14 +806,19 @@ For Cryptographic Key Binding, the Client has the following options defined in (
 
 ## Credential Request {#credential-request}
 
-A Client makes a Credential Request to the Credential Endpoint by sending the following parameters in the entity-body of an HTTP POST request using the `application/json` media type.
+A Client makes a Credential Request to the Credential Endpoint by sending the following parameters in the entity-body of an HTTP POST request.
+
+The Credential Issuer indicates support for encrypted requests by including the `credential_request_encryption` parameter in the Credential Issuer Metadata. The Client MAY encrypt the request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`. 
+
+When performing Credential Request encryption, the Client MUST encode the information in the Credential Request in a JWT as specified by [#encrypted-messages], using the parameters in from `credential_request_encryption` object. 
+
+If the Credential Request is not encrypted, the media type of the request MUST be set to `application/json`.
 
 * `credential_identifier`: REQUIRED when an Authorization Details of type `openid_credential` was returned from the Token Response. It MUST NOT be used otherwise. A string that identifies a Credential Dataset that is requested for issuance. When this parameter is used, the `credential_configuration_id` MUST NOT be present.
 * `credential_configuration_id`: REQUIRED if a `credential_identifiers` parameter was not returned from the Token Response as part of the `authorization_details` parameter. It MUST NOT be used otherwise. String that uniquely identifies one of the keys in the name/value pairs stored in the `credential_configurations_supported` Credential Issuer metadata. The corresponding object in the `credential_configurations_supported` map MUST contain one of the value(s) used in the `scope` parameter in the Authorization Request. When this parameter is used, the `credential_identifier` MUST NOT be present.
 * `proofs`: OPTIONAL. Object providing one or more proof of possessions of the cryptographic key material to which the issued Credential instances will be bound to. The `proofs` parameter contains exactly one parameter named as the proof type in (#proof-types), the value set for this parameter is a non-empty array containing parameters as defined by the corresponding proof type.
 * `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If this request element is not present, the corresponding credential response returned is not encrypted.
     * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
-    * `alg`: REQUIRED. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses.
     * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
 
 See (#identifying_credential) for the summary of the options how requested Credential(s) are identified throughout the Issuance flow.
@@ -905,9 +910,7 @@ Authorization: BEARER czZCaGRSa3F0MzpnWDFmQmF0M2JW
 }
 ```
 
-The Client MAY request encrypted responses by providing its encryption parameters in the Credential Request.
-
-The Credential Issuer indicates support for encrypted responses by including the `credential_response_encryption` parameter in the Credential Issuer Metadata.
+The Credential Issuer indicates support for encrypted responses by including the `credential_response_encryption` parameter in the Credential Issuer Metadata. The Client MAY request encrypted responses by providing its encryption parameters in the Credential Request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`. 
 
 ### Proof Types {#proof-types}
 
@@ -1089,7 +1092,7 @@ Credential Response can be immediate or deferred and can contain one or more Cre
 
 The HTTP status code MUST be 202 (see Section 15.3.3 of [@!RFC9110]).
 
-If the Client requested an encrypted response by including the `credential_response_encryption` object in the request, the Credential Issuer MUST encode the information in the Credential Response as a JWT using the  parameters from the `credential_response_encryption` object. If the Credential Response is encrypted, the media type of the response MUST be set to `application/jwt`. If encryption was requested in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
+If the Client requested an encrypted response by including the `credential_response_encryption` object in the request, the Credential Issuer MUST encode the information in the Credential Response as specified by [#encrypted-messages], using the parameters from the `credential_response_encryption` object. 
 
 If the Credential Response is not encrypted, the media type of the response MUST be set to `application/json`.
 
@@ -1275,6 +1278,16 @@ Cache-Control: no-store
 }
 ```
 
+## Encypted Messages {#encrypted-messages}
+Encryption of Request and Response Messages is performed as follows:
+
+The contents of the message MUST be encoded as a JWT as described in [@!RFC7519]. The media type MUST be set to `application/jwt`.
+
+The Public Key to which to encrypt the message is selected based on the context. In the case where multiple public keys are available, any may be selected based on the information about each key, such as the `kty` (Key Type), `use` (Public Key Use), `alg` (Algorithm), and other JWK parameters. The `alg` parameter MUST be present. The JWE `alg` algorithm used MUST be equal to the `alg` value of the chosen `jwk`. If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in [@!RFC7516, Section 4.1.6]) of the encrypted message. This enables the easy identification of the specific public key that was used to encrypt the message. The JWE `enc` content encryption algorithm used is obtained based on context.
+
+When encryption of a message was required but the received message is unencrypted, it SHOULD be rejected.
+
+
 # Notification Endpoint {#notification-endpoint}
 
 This endpoint is used by the Wallet to notify the Credential Issuer of certain events for issued Credentials. These events enable the Credential Issuer to take subsequent actions after issuance. The Credential Issuer needs to return one `notification_id` parameter per Credential Response or Deferred Credential Response for the Wallet to be able to use this endpoint. Support for this endpoint is OPTIONAL. The Issuer cannot assume that a notification will be sent for every issued Credential since the use of this Endpoint is not mandatory for the Wallet.
@@ -1422,6 +1435,10 @@ This specification defines the following Credential Issuer Metadata parameters:
 * `nonce_endpoint`: OPTIONAL. URL of the Credential Issuer's Nonce Endpoint, as defined in (#nonce-endpoint). This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Credential Issuer does not support the Nonce Endpoint.
 * `deferred_credential_endpoint`: OPTIONAL. URL of the Credential Issuer's Deferred Credential Endpoint, as defined in (#deferred-credential-issuance). This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Credential Issuer does not support the Deferred Credential Endpoint.
 * `notification_endpoint`: OPTIONAL. URL of the Credential Issuer's Notification Endpoint, as defined in (#notification-endpoint). This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Credential Issuer does not support the Notification Endpoint.
+* `credential_request_encryption`: OPTIONAL. Object containing information about whether the Credential Issuer supports encryption of the Credential Request on top of TLS.
+  * `jwks`: REQUIRED. A JSON Web Key Set, as defined in [@!RFC7591], that contains one or more public keys, to be used by the Wallet as an input to a key agreement for encryption of the Credential Request. Each JWK in the set MUST have a kid (Key ID) parameter that uniquely identifies the key.
+  * `enc_values_supported`: REQUIRED. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential Endpoint to decode the Credential Request from a JWT [@!RFC7519].
+  * `encryption_required`: REQUIRED. Boolean value specifying whether the Credential Issuer requires the additional encryption on top of TLS for the Credential Requests. If the value is `true`, the Credential Issuer requires encryption for every Credential Request. If the value is `false`, the Wallet MAY choose whether it encrypts the request or not.
 * `credential_response_encryption`: OPTIONAL. Object containing information about whether the Credential Issuer supports encryption of the Credential Response on top of TLS.
   * `alg_values_supported`: REQUIRED. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`alg` values) [@!RFC7518] supported by the Credential Endpoint to encode the Credential Response in a JWT [@!RFC7519].
   * `enc_values_supported`: REQUIRED. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential Endpoint to encode the Credential Response in a JWT [@!RFC7519].
@@ -2758,6 +2775,7 @@ The technology described in this specification was made available from contribut
    * add missing request for media type registration of key-attestation+jwt in IANA Considerations
    * rename keyattestation+jwt to key-attestation+jwt
    * Remove the Dynamic Credential Request section and associated content
+   * Add Credential Request encryption
 
    -15
 


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1422,9 +1447,15 @@ This specification defines the following Credential Issuer Metadata parameters:
...
+* `credential_request_encryption`: OPTIONAL. Object containing information about whether the Credential Issuer supports encryption of the Credential Request on top of TLS.
+  * `jwks`: REQUIRED. A JSON Web Key Set, as defined in [@!RFC7591], that contains one or more public keys, to be used by the Wallet as an input to a key agreement for encryption of the Credential Request. Each JWK in the set MUST have a kid (Key ID) parameter that uniquely identifies the key.
+  * `enc_values_supported`: REQUIRED. Array containing a list of the JWE [@!RFC7516] encryption algorithms (`enc` values) [@!RFC7518] supported by the Credential Endpoint to decode the Credential Request from a JWT [@!RFC7519].
+  * `zip_values_supported`: OPTIONAL. Array containing a list of the JWE [@!RFC7516] compression algorithms (`zip` values) [@!RFC7518] supported by the Credential Endpoint to compress the Credential Request prior to encryption. If absent then compression is not supported.
```
bc-pi, line 1453: I don't think the Credential Endpoint is the entity that would/could compress the Credential Request. Honestly, I'm getting turned around in my understanding here but maybe this fits what was intended? 

```suggestion
  * `zip_values_supported`: OPTIONAL. Array containing a list of the JWE [@!RFC7516] compression algorithms (`zip` values) [@!RFC7518] supported by the Credential Endpoint that the Wallet can use to compress the Credential Request prior to encryption. If absent then compression is not supported. 
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -806,15 +806,15 @@ For Cryptographic Key Binding, the Client has the following options defined in (
...
 * `proofs`: OPTIONAL. Object providing one or more proof of possessions of the cryptographic key material to which the issued Credential instances will be bound to. The `proofs` parameter contains exactly one parameter named as the proof type in (#proof-types), the value set for this parameter is a non-empty array containing parameters as defined by the corresponding proof type.
 * `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If this request element is not present, the corresponding credential response returned is not encrypted.
     * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
-    * `alg`: REQUIRED. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses.
     * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
+    * `zip`: OPTIONAL. JWE [@!RFC7516] `zip` algorithm [@!RFC7518] for compressing Credential Responses prior to encryption. If absent then compression MUST not be used.
```
bc-pi, line 817: Someone will probably mention (or maybe not, I don't know) that the JWT BCP recommends against compressing prior to encryption: https://datatracker.ietf.org/doc/html/rfc8725#section-3.6

As I understand things, the BCP's guidance is well-intentioned but somewhat overreaching and lacking in nuance. The kind of nuance that would suggest it's fine in this kind of usage. 

That's not really actionable, sorry, but seemed worth mentioning.

GarethCOliver: Good to know. Is there anything we can/should mention here why this sort of usage is fine? (Or why it is advised against normally)?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1226,14 +1237,18 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
 }
 ```
 
+The Credential Issuer indicates support for encrypted responses by including the `credential_response_encryption` parameter in the Credential Issuer Metadata. The Client MAY request encrypted responses by providing its encryption parameters in the Deferred Credential Request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`.
```
GarethCOliver:
```suggestion
The Credential Issuer indicates support for encrypted responses by including the `credential_response_encryption` parameter in the Credential Issuer Metadata. The Client MAY request encrypted responses by providing its encryption parameters in the Deferred Credential Request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`. Note that this object will be used for encrypting the response, regardless of what was sent in the initial Credential Request. If it is not included encryption will not be performed.
```

GarethCOliver:
> I think we should have some security & privacy considerations added too, albeit I don't mind if we open an issue to do that in another PR so long as it gets into 1.0 - the effectiveness of this mechanism seems to be highly dependent on the wallet frontend fetching the issuer metadata directly rather than through the backend.
> 
> I think it would also be worthwhile to actually state the intended advantages of the encryption mechanism in the specification itself.

Makes sense. Probably two sections: One for what encryption can protect against and what it doesn't (particular in regard to TLS already existing). The second is the one christian is starting for wallet backend considerations. The problem you are describing is definitely true for retrieving the keys (you have to have some way, on the device, to trust them (the most  obvious of which is fetching them directly). I think you would have the same issue with any other client metadata you are trusting (i.e. if it isn't fetched directly you can't trust it, e.g. if I pulled the endpoint from a wallet backend, even if I directly connected to it, it'd be a problem). 

Filed two issues on these.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1275,6 +1290,16 @@ Cache-Control: no-store
 }
 ```
 
+## Encypted Messages {#encrypted-messages}
+Encryption of Request and Response Messages is performed as follows:
+
+The contents of the message MUST be encoded as a JWT as described in [@!RFC7519]. The media type MUST be set to `application/jwt`.
+
+The Public Key to which to encrypt the message is selected based on the context. In the case where multiple public keys are available, any may be selected based on the information about each key, such as the `kty` (Key Type), `use` (Public Key Use), `alg` (Algorithm), and other JWK parameters. The `alg` parameter MUST be present. The JWE `alg` algorithm used MUST be equal to the `alg` value of the chosen `jwk`. If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in [@!RFC7516, Section 4.1.6]) of the encrypted message. This enables the easy identification of the specific public key that was used to encrypt the message. The JWE `enc` content encryption algorithm used is obtained based on context.
```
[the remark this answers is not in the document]
GarethCOliver: Appreciate any suggestions there. I wanted to avoid re-writing the exact same text in all the sections where we advise about encryption (so how to get the parameters is from the context of the encryption and the text that references it provides that information). Didn't like this wording but having one set of text describing how to do this seemed better than duplicating it.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -829,6 +829,12 @@ The `c_nonce` value can be retrieved from the Nonce Endpoint as defined in (#non
 Additional Credential Request parameters MAY be defined and used.
 The Credential Issuer MUST ignore any unrecognized parameters.
 
+The Credential Issuer indicates support for encrypted requests by including the `credential_request_encryption` parameter in the Credential Issuer Metadata. The Client MAY encrypt the request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`.
```
Sakurann, line 832:
```suggestion
The Credential Issuer indicates support for encrypted requests by including the `credential_request_encryption` parameter in the Credential Issuer Metadata. The Wallet MAY encrypt the request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`. 
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -829,6 +829,12 @@ The `c_nonce` value can be retrieved from the Nonce Endpoint as defined in (#non
 Additional Credential Request parameters MAY be defined and used.
 The Credential Issuer MUST ignore any unrecognized parameters.
 
+The Credential Issuer indicates support for encrypted requests by including the `credential_request_encryption` parameter in the Credential Issuer Metadata. The Client MAY encrypt the request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`. 
+
+When performing Credential Request encryption, the Client MUST encode the information in the Credential Request in a JWT as specified by [#encrypted-messages], using the parameters in from `credential_request_encryption` object.
```
```suggestion
When performing Credential Request encryption, the Client MUST encode the information in the Credential Request in a JWT as specified by (#encrypted-messages), using the parameters in from `credential_request_encryption` object. 
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -829,6 +829,12 @@ The `c_nonce` value can be retrieved from the Nonce Endpoint as defined in (#non
 Additional Credential Request parameters MAY be defined and used.
 The Credential Issuer MUST ignore any unrecognized parameters.
 
+The Credential Issuer indicates support for encrypted requests by including the `credential_request_encryption` parameter in the Credential Issuer Metadata. The Client MAY encrypt the request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`. 
+
+When performing Credential Request encryption, the Client MUST encode the information in the Credential Request in a JWT as specified by [#encrypted-messages], using the parameters in from `credential_request_encryption` object. 
+
+If the Credential Request is not encrypted, the media type of the request MUST be set to `application/json`.
```
Sakurann, line 830: it's kind of weird having media type for non-encrypted case all across the spec and for encrypted case in one section...

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -806,15 +806,15 @@ For Cryptographic Key Binding, the Client has the following options defined in (
...
 * `proofs`: OPTIONAL. Object providing one or more proof of possessions of the cryptographic key material to which the issued Credential instances will be bound to. The `proofs` parameter contains exactly one parameter named as the proof type in (#proof-types), the value set for this parameter is a non-empty array containing parameters as defined by the corresponding proof type.
 * `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If this request element is not present, the corresponding credential response returned is not encrypted.
     * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
-    * `alg`: REQUIRED. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses.
```
Sakurann, line 810: why is alg being removed..? alg and zip are not mutually exclusive, no?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1089,7 +1093,7 @@ Credential Response can be immediate or deferred and can contain one or more Cre
 
 The HTTP status code MUST be 202 (see Section 15.3.3 of [@!RFC9110]).
 
-If the Client requested an encrypted response by including the `credential_response_encryption` object in the request, the Credential Issuer MUST encode the information in the Credential Response as a JWT using the  parameters from the `credential_response_encryption` object. If the Credential Response is encrypted, the media type of the response MUST be set to `application/jwt`. If encryption was requested in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
+If the Client requested an encrypted response by including the `credential_response_encryption` object in the request, the Credential Issuer MUST encode the information in the Credential Response as specified by [#encrypted-messages], using the parameters from the `credential_response_encryption` object.
```
```suggestion
If the Client requested an encrypted response by including the `credential_response_encryption` object in the request, the Credential Issuer MUST encode the information in the Credential Response as specified by (#encrypted-messages), using the parameters from the `credential_response_encryption` object. 
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1226,14 +1237,18 @@ Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
...
+
 ## Deferred Credential Response {#deferred-credential-response}
 
 The Deferred Credential Response uses the `credentials` and `notification_id` parameters as defined in (#credential-response).
 
 Additional Deferred Credential Response parameters MAY be defined and used.
 The Wallet MUST ignore any unrecognized parameters.
 
-The Deferred Credential Response MUST be sent using the `application/json` media type.
+If the Client requested an encrypted response by including the `credential_response_encryption` object in the request, the Credential Issuer MUST encode the information in the Deferred Credential Response as specified by [#encrypted-messages], using the parameters from the `credential_response_encryption` object. The `credential_response_encryption` object may be different from the one included in the initial Credential Request so the Credential Issuer MUST use the newly provided one. This is to simplify key management in the case of longer deferred issuance.
```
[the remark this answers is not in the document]
Sakurann: @GarethCOliver can you please resolve conflicts by pulling the main as joseph suggests? (i also want to see html generated once the pipeline runs to confirm that [#---] reference is not compiling correctly)

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: I can pull it out if you like, I just liked having one spot to read what you needed to do for creating encrypted messages.

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: `alg` is being removed for the same reason we did in openid4vp: It is present in the key and this can only lead to conflicting definitions. For consistency cross-spec (and for the same argument) I removed it (mentioned in PR description)

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: WG discussion: rough agreement that JWT BCP recommendation against compressing prior to encryption does not apply here. so proceeding with compressing prior to encryption :)

selfissued: This precondition for the attack in https://datatracker.ietf.org/doc/html/rfc8725#section-2.4 is not met in this case:
> Compression attacks are particularly powerful when there is attacker-controlled data in the same compression space as secret data

Therefore, I believe the use of `zip` is OK

Sakurann:
WG discussion:
- Gareth to pull the main, folks to re-review after that.
- nothing technically blocking the PR

babisRoutis (approved):

openid-4-verifiable-credential-issuance-1_0.md
jogu: I think this would be solved by https://github.com/openid/OpenID4VCI/issues/523 so we can defer for now?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1201,15 +1205,22 @@ Communication with the Deferred Credential Endpoint MUST utilize TLS.
...
+* `credential_response_encryption`: OPTIONAL. as defined in (#credential-request).
 
 The Credential Issuer MUST invalidate the `transaction_id` after the Credential for which it was meant has been obtained by the Wallet.
 
 Additional Deferred Credential Request parameters MAY be defined and used.
 The Credential Issuer MUST ignore any unrecognized parameters.
 
+The Credential Issuer indicates support for encrypted requests by including the `credential_request_encryption` parameter in the Credential Issuer Metadata. The Client MAY encrypt the request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`.
+
+When performing Deferred Credential Request encryption, the Client MUST encode the information in the Deferred Credential Request in a JWT as specified by [#encrypted-messages], using the parameters in from `credential_request_encryption` object.
```
jogu: grammar?

```suggestion
When performing Deferred Credential Request encryption, the Client MUST encode the information in the Deferred Credential Request in a JWT as specified by (#encrypted-messages), using the parameters from the `credential_request_encryption` object in the Credential Issuer Metadata. 
```

openid-4-verifiable-credential-issuance-1_0.md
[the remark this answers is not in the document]
jogu: Gareth updated it now, thanks!

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1235,7 +1248,9 @@ The Deferred Credential Response MAY use the `notification_id` parameter as defi
 Additional Deferred Credential Response parameters MAY be defined and used.
 The Wallet MUST ignore any unrecognized parameters.
 
-The Deferred Credential Response MUST be sent using the `application/json` media type.
+If the Client requested an encrypted response by including the `credential_response_encryption` object in the request, the Credential Issuer MUST encode the information in the Deferred Credential Response as specified by [#encrypted-messages], using the parameters from the `credential_response_encryption` object. The `credential_response_encryption` object may be different from the one included in the initial Credential Request so the Credential Issuer MUST use the newly provided one. This is to simplify key management in the case of longer deferred issuance.
```
jogu: It feels kind of pointless to encrypt the "I still need more time" response ( `{ "interval": 30 }` ) but I guess also mostly harmless to do so.

Assuming it is the intention that we encrypt it, I would prefer that we very explicitly state that it is encrypted. I think we should also explicitly state that the error response is (I believe) not encrypted. I think If we don't explicitly state both those things we'll see different implementations doing different things which might be bad for interoperability.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -828,6 +828,12 @@ The `c_nonce` value can be retrieved from the Nonce Endpoint as defined in (#non
 Additional Credential Request parameters MAY be defined and used.
 The Credential Issuer MUST ignore any unrecognized parameters.
 
+The Credential Issuer indicates support for encrypted requests by including the `credential_request_encryption` parameter in the Credential Issuer Metadata. The Client MAY encrypt the request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`. 
+
+When performing Credential Request encryption, the Client MUST encode the information in the Credential Request in a JWT as specified by (#encrypted-messages), using the parameters in from `credential_request_encryption` object.
```
jogu: Grammar:

```suggestion
When performing Credential Request encryption, the Client MUST encode the information in the Credential Request in a JWT as specified by (#encrypted-messages), using the parameters from the `credential_request_encryption` object in the Credential Issuer Metadata.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1287,6 +1302,16 @@ Cache-Control: no-store
 }
 ```
 
+## Encrypted Messages {#encrypted-messages}
+Encryption of Request and Response Messages is performed as follows:
+
+The contents of the message MUST be encoded as a JWT as described in [@!RFC7519]. The media type MUST be set to `application/jwt`.
+
+The Public Key used to encrypt the message is selected based on the context. In the case where multiple public keys are available, any may be selected based on the information about each key, such as the `kty` (Key Type), `use` (Public Key Use), `alg` (Algorithm), and other JWK parameters. The `alg` parameter MUST be present. The JWE `alg` algorithm used MUST be equal to the `alg` value of the chosen JWK. If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in [@!RFC7516, Section 4.1.6]) of the encrypted message. This enables the easy identification of the specific public key that was used to encrypt the message. The JWE `enc` content encryption algorithm used is obtained based on context.
```
jogu, line 1313: The two "based on context" I think might be overly vague. Although I'm struggling to come up with less vague suggestions.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1287,6 +1302,16 @@ Cache-Control: no-store
...
+
+When encryption of a message was required but the received message is unencrypted, it SHOULD be rejected.
```
jogu, line 1317: This kind of suggests that actually the error response needs to be encrypted. Unless, of course, the encryption parameters passed are invalid :)

jogu (commented):
Couple of suggestions, and a couple of comments where I couldn't come up with suggestions (sorry!).

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: Yeah, I think this starts to get down to partial-encryption questions (i.e. what is valid to encrypt, and what isn't). Our initial WG conversation was to keep this as simple as possible, and so always encrypt successful credential responses. 

We could change that (to, for example, only encrypt credential responses that contain credentials, and the others are in plaintext). Will update the text to reflect the current intent (credential response is *always* encrypted, credential error responses are not encrypted), and leave any more nuanced encryption to if/when we decide to resolve #502.

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: I agree, I did not like the wording but could not come up with anything better. When read in the context it is referenced I feel like it's clear, at least?

openid-4-verifiable-credential-issuance-1_0.md
jogu: Discussed on WG today:

Agreement on Gareth's suggestion to always encrypt responses (if encryption requested) but not to ever encrypt errors.

Gareth to update today/tomorrow.

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: clarified that they are not to be.

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: done

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1089,7 +1093,7 @@ The Credential Response can either be returned immediately or in a deferred mann
...
 
-If the Client requested an encrypted response by including the `credential_response_encryption` object in the request, the Credential Issuer MUST encode the JSON-encoded Credential Response body as a JWT using the parameters from the `credential_response_encryption` object. If the Credential Response is encrypted, the media type of the response MUST be set to `application/jwt`. If encryption was requested in the Credential Request and the Credential Response is not encrypted, the Client SHOULD reject the Credential Response.
+If the Client requested an encrypted response by including the `credential_response_encryption` object in the request, the Credential Issuer MUST encode the information in the Credential Response as specified by (#encrypted-messages), using the parameters from the `credential_response_encryption` object. Note that this is be done regardless of the content.
```
jogu: Grammar suggestion:

```suggestion
If the Client requested an encrypted response by including the `credential_response_encryption` object in the request, the Credential Issuer MUST encode the information in the Credential Response as specified by (#encrypted-messages), using the parameters from the `credential_response_encryption` object. Note that this is done regardless of the content. 
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2797,6 +2830,7 @@ The technology described in this specification was made available from contribut
    * Remove the Dynamic Credential Request section and associated content
    * Align claims path query for ISO mdocs with JSON-based credentials
    * define `proof_signing_alg_values_supported` for attestation proof type
+   * Add Credential Request encryption and Zip support
```
jogu, line 2877: Think we should call out the change to response encryption too:

```suggestion
   * Add Credential Request encryption and Zip support
   * Request encryption is now required when response encryption is used
```

openid-4-verifiable-credential-issuance-1_0.md
jogu: Yeah, fair enough, I think the intent is relatively easy to figure out. We can always make editorial improvements later.

jogu (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -805,15 +805,15 @@ For Cryptographic Key Binding, the Client has the following options defined in (
 
 ## Credential Request {#credential-request}
 
-A Client makes a Credential Request to the Credential Endpoint by sending the following parameters in the entity-body of an HTTP POST request using the `application/json` media type.
+A Client makes a Credential Request to the Credential Endpoint by sending the following parameters in the entity-body of an HTTP POST request.
```
tlodderstedt: While reviewing, I was wondering the while time where the response encryption is described. To me it seems it is in the combination of the section about message encryption and metadata. To improve readability and help the understanding of the readers, I suggest to add text here pointing out credential requests can be encrypted (on top of TLS) and refer to the section about encrypted messages and metadata sections.

tlodderstedt (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: makes sense. Done.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1287,6 +1304,16 @@ Cache-Control: no-store
 }
 ```
 
+## Encrypted Messages {#encrypted-messages}
+Encryption of Request and Response Messages is performed as follows:
+
+The contents of the message MUST be encoded as a JWT as described in [@!RFC7519]. The media type MUST be set to `application/jwt`.
+
+The Public Key used to encrypt the message is selected based on the context. In the case where multiple public keys are available, any may be selected based on the information about each key, such as the `kty` (Key Type), `use` (Public Key Use), `alg` (Algorithm), and other JWK parameters. The `alg` parameter MUST be present. The JWE `alg` algorithm used MUST be equal to the `alg` value of the chosen JWK. If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in [@!RFC7516, Section 4.1.6]) of the encrypted message. This enables the easy identification of the specific public key that was used to encrypt the message. The JWE `enc` content encryption algorithm used is obtained based on context.
```
c2bo, line 1313: Shouldn't we mention compression here? We currently introduce compression for request and response, but do not mention here that/when it is used for building the JWE

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -828,6 +828,12 @@ The `c_nonce` value can be retrieved from the Nonce Endpoint as defined in (#non
 Additional Credential Request parameters MAY be defined and used.
 The Credential Issuer MUST ignore any unrecognized parameters.
 
+The Credential Issuer indicates support for encrypted requests by including the `credential_request_encryption` parameter in the Credential Issuer Metadata. The Client MAY encrypt the request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`.
```
paulbastian, line 826: If request encryption is bound to response encryption, we should only have a single issuer metadata parameter?

paulbastian (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
jogu: Discussed on WG call: This would tie the two features together, we may make changes in the future that make you can use response encryption without request encryption, so better to keep them separate.

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: done

c2bo (approved):

paulbastian:
retracting request changes (as GH doesn't allow that)

PR #509, opened 2025-05-19
Presentation During Issuance

danielfett:
First, minimal version of Presentation During Issuance (Issue #473) . Work in Progress.

There are four open todos:

- [x] Determine whether we a way to show a website. The first-part apps draft right now only contains a fallback to the regular web flow, not a flexible way to open an in-app browser tab. If we add something like this, we may need to include privacy considerations.
- [x] The original draft contains a `presentation_during_issuance_session` parameter, for binding the presentation response to the issuance process. Do we need this, and if yes, how do we define it?
- [x] Determine if we want to make the extension point ("define your own responses here for scanning an ID card") explicit or not.
- [x] Add new endpoint / OAuth error / etc to IANA considerations

PR #511, opened 2025-05-20
Editorial changes to multiple/batch credential issuance section

### LINKED ISSUE #508: Cleanup references to batch
Section 3.3.2 introduces "multiple credential issuance", but the configuration in section 11.2.3, and error notification in 10.1 still refers to batch issuance. For an implementor, the 3.3.2 is not clearly correlated to these extra requirements. The simplest change is to reintroduce the term batch within section 3.3.2.

sloops77:
resolves #508


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -170,11 +170,9 @@ Credentials can therefore vary in the following dimensions:
 - Credential Format
 - Cryptographic Data
 
-In the context of a single request, all issued Credentials MUST share the same Credential Format and Credential Dataset. However, they MAY contain different Cryptographic Data, such as being bound to different cryptographic keys (e.g., to ensure unlinkability between the Credentials).
+In the context of a single request, all issued Credentials belonging to the batch MUST share the same Credential Format and Credential Dataset. However, they MAY contain different Cryptographic Data, such as being bound to different cryptographic keys (e.g., to ensure unlinkability between the Credentials).
 To issue multiple Verifiable Credentials with differing Credential Formats or Credential Datasets, multiple requests MUST be sent to the Credential Endpoint.
 
-In the course of the authorization process, the Credential Issuer MAY also request Credential presentation as a means to authenticate or identify the End-User during the issuance flow, as described in (#use-case-5).
```
Sakurann, line 176: why was this deleted?

sloops77: @Sakurann I was editing the batch credential section. I didn't understand why a reference to presentation based authentication was here. I think it needs to be moved but maybe it makes more sense for me to restore it  and get this merged asap, and then it can be sorted out in https://github.com/openid/OpenID4VCI/pull/509

paulbastian: I agree it seems to be placed in the wrong section. Can we move it somewhere more sensible than deleting it or is it already mentioned elsewhere?

sloops77: @paulbastian @Sakurann i moved the section to the introduction to authorization code flow.

sloops77:
@jogu I feel that your comments are coming from the perspective that batches means 2 or more case, whereas the way i wrote this is that the batches can be 1 or more. 

I think i still think I prefer to describe the credential endpoint issuing batches of 1 or more, because the response properties are the same in all cases. However I see that the spec is written as if batch is 2 or more, and therefore these inconsistencies should be fixed in this PR.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -158,10 +158,10 @@ To this end, extension points to add Credential Format specific parameters in th
 Credential Format Profiles for IETF SD-JWT VC [@I-D.ietf-oauth-sd-jwt-vc], ISO mDL [@ISO.18013-5], and W3C VCDM [@VC_DATA] are specified in (#format-profiles).
 Other specifications or deployments can define their own Credential Format Profiles using the above-mentioned extension points.
 
-### Multiple Credential Issuance
+### Batch Credential Issuance
 
 This specification enables the issuance of Verifiable Credentials through the Credential Endpoint. 
-A single request message to this endpoint may request the issuance of one or more Verifiable Credentials.
+A single request message to this endpoint may request the issuance of one or more Verifiable Credentials in a batch.
```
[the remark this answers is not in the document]
jogu: withdrawing this suggestion as per https://github.com/openid/OpenID4VCI/pull/511#issuecomment-2932162091

jogu (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1290,7 +1290,7 @@ Communication with the Notification Endpoint MUST utilize TLS.
...
+* `event`: REQUIRED. Type of the notification event. It MUST be a case sensitive string whose value is either `credential_accepted`, `credential_failure`, or `credential_deleted`. `credential_accepted` is to be used when the Credentials were successfully stored in the Wallet, with or without user action. `credential_deleted` is to be used when the unsuccessful Credential issuance was caused by a user action. In all other unsuccessful cases, `credential_failure` is to be used. Partial errors during issuance multiple Credentials as a batch (e.g., one of the Credentials in the batch could not be stored) MUST be treated as the overall issuance flow failing.
```
```suggestion
* `event`: REQUIRED. Type of the notification event. It MUST be a case sensitive string whose value is either `credential_accepted`, `credential_failure`, or `credential_deleted`. `credential_accepted` is to be used when the Credentials were successfully stored in the Wallet, with or without user action. `credential_deleted` is to be used when the unsuccessful Credential issuance was caused by a user action. In all other unsuccessful cases, `credential_failure` is to be used. Partial errors during issuance of multiple Credentials as a batch (e.g., one of the Credentials in the batch could not be stored) MUST be treated as the overall issuance flow failing.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1426,8 +1424,8 @@ This specification defines the following Credential Issuer Metadata parameters:
...
-* `batch_credential_issuance`: OPTIONAL. Object containing information about the Credential Issuer's supports for batch issuance of Credentials on the Credential Endpoint. The presence of this parameter means that the issuer supports more than one key proof in the `proofs` parameter in the Credential Request so can issue more than one Verifiable Credential for the same Credential Dataset in a single request/response.
-  * `batch_size`: REQUIRED. Integer value specifying the maximum array size for the `proofs` parameter in a Credential Request.
+* `batch_credential_issuance`: OPTIONAL. Object containing information about the Credential Issuer's support for issuance of multiple Credentials in a batch in the Credential Endpoint. The presence of this parameter means that the issuer supports more than one key proof in the `proofs` parameter in the Credential Request so can issue more than one Verifiable Credential for the same Credential Dataset in a single request/response.
+  * `batch_size`: REQUIRED. Integer value specifying the maximum array size for the `proofs` parameter in a Credential Request. It MUST be 2 or greater.
```
Sakurann, line 1429: > I asked some of the other editors and everyone seemed happy if we use a definition of "batch" that means "1 or more" so I'll withdraw one of my suggestions.

if this is true, should this not be `1 or greater`?

jogu: I think because of the "The presence of this parameter means that the issuer supports more than one key proof in the `proofs` parameter" on the parent object, it can only be legal to have batch_size of at least 2.

sloops77: @Sakurann logically it should be 2 or more, but practically we could drop the statement as it _probably_ doesnt harm interop

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -158,10 +158,10 @@ To this end, extension points to add Credential Format specific parameters in th
 Credential Format Profiles for IETF SD-JWT VC [@I-D.ietf-oauth-sd-jwt-vc], ISO mDL [@ISO.18013-5], and W3C VCDM [@VC_DATA] are specified in (#format-profiles).
 Other specifications or deployments can define their own Credential Format Profiles using the above-mentioned extension points.
 
-### Multiple Credential Issuance
+### Batch Credential Issuance
 
 This specification enables the issuance of Verifiable Credentials through the Credential Endpoint. 
-A single request message to this endpoint may request the issuance of one or more Verifiable Credentials.
+A single request message to this endpoint may request the issuance of a batch of one or more Verifiable Credentials.
```
```suggestion
A single request to this endpoint may request the issuance of a batch of one or more Verifiable Credentials.
```

I know this change wasn't introduced by you but re-reading this sentence I think `message` isn't required in this sentence.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -170,11 +170,9 @@ Credentials can therefore vary in the following dimensions:
 - Credential Format
 - Cryptographic Data
 
-In the context of a single request, all issued Credentials MUST share the same Credential Format and Credential Dataset. However, they MAY contain different Cryptographic Data, such as being bound to different cryptographic keys (e.g., to ensure unlinkability between the Credentials).
+In the context of a single request, the batch of issued Credentials MUST share the same Credential Format and Credential Dataset. However, they MAY contain different Cryptographic Data, such as being bound to different cryptographic keys (e.g., to ensure unlinkability between the Credentials).
```
tplooker: Again not a change introduced by this PR but I think we should consider making the MAY here even stronger, for example when would we want to allow multiple instances of a credential to be issued to the same keys? If we dont change this it conflicts with section 10.2 of SD-JWT which states

> A batch of credentials based on the
   same claims is issued to the Holder instead of just a single
   credential.  The Holder can then use a different credential for each
   Verifier or even for each session with a Verifier.  New key binding
   keys and salts MUST be used for each credential in the batch to
   ensure that the Verifiers cannot link the credentials using these
   values

openid-4-verifiable-credential-issuance-1_0.md
tplooker: I'm happy with this change.

openid-4-verifiable-credential-issuance-1_0.md
sloops77: @tplooker pls re-review

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -158,22 +158,21 @@ To this end, extension points to add Credential Format specific parameters in th
...
 Credentials can therefore vary in the following dimensions:
 
-- Credential Dataset
 - Credential Format
+- Credential Dataset
 - Cryptographic Data
 
-In the context of a single request, all issued Credentials MUST share the same Credential Format and Credential Dataset. However, they MAY contain different Cryptographic Data, such as being bound to different cryptographic keys (e.g., to ensure unlinkability between the Credentials).
-To issue multiple Verifiable Credentials with differing Credential Formats or Credential Datasets, multiple requests MUST be sent to the Credential Endpoint.
+In the context of a single request, the batch of issued Credentials MUST share the same Credential Format and Credential Dataset, but SHOULD contain different Cryptographic Data. For example to achieve unlinkability between the Credentials, each credential should be bound to different cryptographic keys.
```
```suggestion
In the context of a single request, the batch of issued Credentials sent in response MUST share the same Credential Format and Credential Dataset, but SHOULD contain different Cryptographic Data. For example to achieve unlinkability between the Credentials, each credential should be bound to different cryptographic keys.
```

tplooker: Thanks @Sakurann one more minor suggestion.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: ok. let's keep it as-is then

Sakurann (approved):

PR #513, opened 2025-05-20
Align claims path query for ISO mdocs with JSON-based credentials

### LINKED ISSUE #468: Rationale for limiting claims description path for mdoc to only namespace and element?
mDOC supports nested claims (e.g. driving_privileges), however the claims description path as defined in Appendix B & C only allows the path array to have a length of two.

It would be useful to also enable setting claim descriptions for nested entries in mDOC credentials, especially since there is no `vct` with additional metadata that can be fetched for mDOC.

There's probably some complexity involved in mapping CBOR structure to a claims path, but for straighforward CBOR types such as object/map/array it would be quite easy to define.

I tried looking into the discussion of PR https://github.com/openid/OpenID4VCI/pull/276, but it's quite big and couldn't directly find a rationale for limiting the mDOC type.

awoie:
Fixes #468

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2371,22 +2371,31 @@ The result of the processing is the set of selected JSON elements.
 This section defines the semantics of a claims path pointer when applied to a
 credential in ISO mdoc format.
 
-A claims path pointer into an mdoc contains two elements of type string. The
-first element refers to a namespace and the second element refers to a data
-element identifier.
+* A string or integer value indicates that the respective key is to be selected from a map.
+* A null value indicates that all elements of the currently selected array(s) are to be selected.
+* A non-negative integer may also indicate a specific index within an array.
+
+Whether a non-negative integer is interpreted as a map key or an array index depends on the type of the currently selected data structure.
+
+The path is formed as follows:
+
+* To address a particular namespace, append the namespace identifier (as a string) as the first element of the path.
+* To address a claim (i.e., data element) within a namespace, or a claim within nested data structure, append the corresponding key (string or integer).
+* To address an element within an array, append the index (as a non-negative, 0-based integer).
+* To address all elements of an array, append a null value.
 
 ### Processing
 
 In detail, the array is processed as follows:
 
-1. If the claims path pointer does not contain exactly two components or 
-   one of the components is not a string abort processing and return an error.
-2. Select the namespace referenced by the first component. If the namespace does
-   not exist in the mdoc abort processing and return an error.
-3. Select the data element referenced by the second component. If the data element does not exist
-   in the credential abort processing and return an error.
+1. Select the namespace referenced by the first component. If the namespace does not exist in the mdoc, abort processing and return an error.
+2. Process each subsequent component as follows:
+   1. If the currently selected element(s) are arrays, and the component is a non-negative integer, select the element at the given index in each array. If the index is out of bounds in a selected array, remove that array from the selection.
+   2. If the currently selected element(s) are maps, and the component is a string or integer, select the element(s) associated with the key. If a selected element is not a map, abort processing. If the key does not exist in a selected map, remove that map from the selection.
+   3. If the component is null, select all elements in each currently selected array. If any selected element is not an array, abort processing.
+3. If the set of elements currently selected is empty, abort processing and return an error.
 
-The result of the processing is the selected data element value as CBOR data item.
+The result of the processing is the set of selected elements contained within the selected namespace.
 
 ## Claims Path Pointer Example {#claims_path_pointer_example}
 


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2371,22 +2371,31 @@ The result of the processing is the set of selected JSON elements.
 This section defines the semantics of a claims path pointer when applied to a
 credential in ISO mdoc format.
```
GarethCOliver, line 2373: I think this is clearer if we keep the first two always referencing the namespace element identifier. So: 

- A claims path pointer into an mdoc always contains at least two elements. The first two elements are of type string. - The first element refers to a namespace and the second element refers to a data
element identifier.
- The remaining elements are used to provide a path into the value of the element. 

<continue with a description of how to do arbitrary reference into the cbor structure>

Sakurann: +1

Sakurann: made a suggestion below

awoie: @Sakurann @GarethCOliver pls review again. I applied Kristina's suggestion. But I want to note that we don't have that new MUST for JSON-based credentials.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2371,22 +2371,29 @@ The result of the processing is the set of selected JSON elements.
...
 
 In detail, the array is processed as follows:
 
-1. If the claims path pointer does not contain exactly two components or 
-   one of the components is not a string abort processing and return an error.
-2. Select the namespace referenced by the first component. If the namespace does
-   not exist in the mdoc abort processing and return an error.
-3. Select the data element referenced by the second component. If the data element does not exist
-   in the credential abort processing and return an error.
+1. Select the namespace referenced by the first component. If the namespace does not exist in the mdoc, abort processing and return an error.
```
GarethCOliver, line 2389: Step 2 should explicitly be to select the value of the data element with the id that matches the second component (then subsequent one from the one after that). This makes the processing logic match the description above.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2371,22 +2371,29 @@ The result of the processing is the set of selected JSON elements.
...
+2. Process each subsequent component as follows:
+   1. If the currently selected element(s) are arrays, and the component is a non-negative integer, select the element at the given index in each array. If the index is out of bounds in a selected array, remove that array from the selection.
+   2. If the currently selected element(s) are maps, and the component is a string or integer, select the element(s) associated with the key. If a selected element is not a map, abort processing and return an error. If the key does not exist in a selected map, remove that map from the selection.
+   3. If the component is null, select all elements in each currently selected array. If any selected element is not an array, abort processing and return an error.
+3. If the set of elements currently selected is empty, abort processing and return an error.
 
-The result of the processing is the selected data element value as CBOR data item.
+The result of the processing is the set of selected elements contained within the selected namespace.
```
GarethCOliver, line 2396:
```suggestion
The result of the processing is the set of selected elements contained within the selected data element value, or the value itself.
```

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: We should be consistent between the two, no strong opinion.

openid-4-verifiable-credential-issuance-1_0.md
awoie: I agree, this is missing.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2371,22 +2373,31 @@ The result of the processing is the set of selected JSON elements.
...
+2. Select the namespace referenced by the first path component. If the namespace does not exist in the mdoc, abort processing and return an error.
+3. Select the data element with the data element identifier that matches the second path component. If the data element does not exist in the mdoc, abort processing and return an error.
+4. Process each subsequent path components as follows:
+   1. If the currently selected element(s) are arrays, and the path component is a non-negative integer, select the element at the given index in each array. If the index is out of bounds in a selected array, remove that array from the selection.
+   2. If the currently selected element(s) are maps, and the path component is a string or integer, select the element(s) associated with the key. If a selected element is not a map, abort processing and return an error. If the key does not exist in a selected map, remove that map from the selection.
+   3. If the path component is null, select all elements in each currently selected array. If any selected element is not an array, abort processing and return an error.
```
jogu, line 2397: I think something has gone wrong here when adapting the language - in the array case we're not dealing with the "what is not all the elements are an array" case, and the for maps the "If a selected element is not a map, abort processing and return an error." kind of feels like it can't happen because it's inside an "if all the current selected element(s) are maps".

```suggestion
   1. If all the currently selected element(s) are arrays, and the path component is a non-negative integer, select the element at the given index in each array. If the index is out of bounds in a selected array, remove that array from the selection.
   2. If all the currently selected element(s) are maps, and the path component is a string or integer, select the element(s) associated with the key. If the key does not exist in a selected map, remove that map from the selection.
   3. If the path component is null, select all elements in each currently selected array. If any selected element is not an array, abort processing and return an error.
   4. If none of the above applied, abort processing and return an error.
```

awoie: Yep, that makes sense.

jogu (approved):

tplooker (approved):

GarethCOliver (approved):

TimoGlastra (approved):
LGTM

PR #520, opened 2025-05-27
add option to have signed Credential Issuer metadata, remove `signed_metadata` from Credential Issuer metadata

### LINKED ISSUE #202: redesign signed Issuer Metadata
I see a potential security issue as some parameters in the Credential Issuer Metadata may be self-asserted, in particular:
`display`.`name`, `credentials_supported`.`display`.`name` and `credentials_supported`.`display`.`logo`. These values are today taken by Wallets and shown to the user. This might give a false sense of security, as attackers easily can create a fake issuer and fake their identity with this metadata.

The security issue originates in my opinion to the fact that OpenID4vc does not mandate specific trust mechanisms similar as being open to any credential format. While this is a strength in general, it is not clear enough to me from the security and privacy considerations that this gap MUST be filled by ecosystems relying on OpenID4vc.

Proposal: Add a section about the significance of an underlying trust framework ans what resources of the OpenID4VCI protocol must rely on it, i.e. which things must be protected by that trust mechanisms.

Same applies to OpenID4VP probably, but I didn't cross check the considerations section there.

paulbastian:
Closes #202 

- [ ] discuss whether to use JWS instead of JWT for multiple trust frameworks

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1433,7 +1433,7 @@ The signed metadata MUST be secured using a JSON Web Signature (JWS) [@!RFC7515]
   * `iat`: REQUIRED. Integer for the time at which the Credential Issuer Metadata was issued using the syntax defined in [@!RFC7519].
   * `exp`: OPTIONAL. Integer for the time at which the Credential Issuer Metadata is expiring, using the syntax defined in [@!RFC7519].
 
-The Wallet SHOULD establish trust in the signer of the metadata, and obtain the keys to validate the signature before processing the metadata, e.g. using JOSE headers like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanism how to do that is out of scope of this specification. 
+The Wallet SHOULD establish trust in the signer of the metadata, and obtain the keys to validate the signature before processing the metadata, e.g. using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification. 
 
 ### Credential Issuer Metadata Parameters {#credential-issuer-parameters}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1263,7 +1263,7 @@ The Credential Issuer MUST respond with HTTP Status Code 200 and return the Cred
 * an unsigned JSON document using the media type `application/json`
 * a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`
 
-The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type`. It is RECOMMENDED to respond with a `Content-Type` matching to the Wallet's requested `Accept` Header. However, the Credential Issuer MAY ignore the `Accept` Header.
+The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` Header. However, the Credential Issuer MAY ignore the `Accept` Header.
 
 The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the language(s) preferred for display. It is up to the Credential Issuer whether to:
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1417,7 +1417,7 @@ The Credential Issuer MUST respond with HTTP Status Code 200 and return the Cred
 * an unsigned JSON document using the media type `application/json`
 * a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`
 
-The Credential Issuer is RECOMMENDED to respond with a `Content-Type` matching to the Wallet's requested `Accept` Header. However, the Credential Issuer MAY ignore the `Accept` Header. In any way, it MUST indicate the type of the returned Credential Issuer Metadata using the HTTP `Content-Type`.
+The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type`. It is RECOMMENDED to respond with a `Content-Type` matching to the Wallet's requested `Accept` Header. However, the Credential Issuer MAY ignore the `Accept` Header.
 
 The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the language(s) preferred for display. It is up to the Credential Issuer whether to:
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1280,7 +1280,7 @@ Host: server.example.com
 Accept-Language: fr-ch, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
 ```
 
-### Signed Metadata
+### Signed Metadata {#credential-issuer-signed-metadata}
 
 The signed metadata MUST be secured using a JSON Web Signature (JWS) [@!RFC7515] and contain the following elements:
 
@@ -2791,6 +2791,28 @@ in the manner described in [@RFC6838].
 * Interoperability considerations: n/a
 * Published specification: (#keyattestation-jwt) of this specification
 * Applications that use this media type: Applications that use the key attestation format defined in this specification
+* Additional information:
+  - Magic number(s): n/a
+  - File extension(s): n/a
+  - Macintosh file type code(s): n/a
+* Person & email address to contact for further information: Torsten Lodderstedt, torsten@lodderstedt.net
+* Intended usage: COMMON
+* Restrictions on usage: none
+* Author: Torsten Lodderstedt, torsten@lodderstedt.net
+* Change controller: OpenID Foundation Digital Credentials Protocols Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
+* Provisional registration? No
+
+### application/openidvci-issuer-metadata+jwt
+
+* Type name: `application`
+* Subtype name: `openidvci-issuer-metadata+jwt`
+* Required parameters: n/a
+* Optional parameters: n/a
+* Encoding considerations: Uses JWS Compact Serialization, as specified in [@!RFC7515].
+* Security considerations: See the Security Considerations in [@!RFC7519].
+* Interoperability considerations: n/a
+* Published specification: (#credential-issuer-signed-metadata) of this specification
+* Applications that use this media type: Applications that use signed metadata format defined in this specification
 * Additional information:
   - Magic number(s): n/a
   - File extension(s): n/a
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1294,7 +1294,7 @@ The signed metadata MUST be secured using a JSON Web Signature (JWS) [@!RFC7515]
   * `iat`: REQUIRED. Integer for the time at which the Credential Issuer Metadata was issued using the syntax defined in [@!RFC7519].
   * `exp`: OPTIONAL. Integer for the time at which the Credential Issuer Metadata is expiring, using the syntax defined in [@!RFC7519].
 
-The Wallet MUST establish trust in the signer of the metadata. Otherwise, the Wallet MUST reject the signed metadata. When validating the signature, the Wallet obtains the keys to validate the signature before processing the metadata, e.g. using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification.
+When requesting signed metadata, the Wallet MUST establish trust in the signer of the metadata. Otherwise, the Wallet MUST reject the signed metadata. When validating the signature, the Wallet obtains the keys to validate the signature before processing the metadata, e.g. using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification.
 
 ### Credential Issuer Metadata Parameters {#credential-issuer-parameters}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1410,7 +1410,7 @@ For example, the metadata for the Credential Issuer Identifier `https://issuer.e
 
 Communication with the Credential Issuer Metadata Endpoint MUST utilize TLS.
 
-To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate whether it supports signed metadata.
+To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate the Content Type(s) it supports, and by doing so, signaling whether it supports signed metadata.
 
 The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the [parameters defined in this specification](#credential-issuer-parameters) as either:
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1263,7 +1263,7 @@ The Credential Issuer MUST respond with HTTP Status Code 200 and return the Cred
 * an unsigned JSON document using the media type `application/json`, or
 * a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`.
 
-The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` header. However, the Credential Issuer MAY ignore the `Accept` header, e.g. if it doesn't support signed metadata or has another preference.
+The Credential Issuer MUST support returning metadata in an unsigned form 'application/json' and MAY support returning it in a signed form 'application/jwt'. In all cases the Credential Issuer MUST indicate the media type of the returned Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` header when the requested content type is supported.
 
 The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the language(s) preferred for display. It is up to the Credential Issuer whether to:
 
@@ -1294,7 +1294,7 @@ The signed metadata MUST be secured using a JSON Web Signature (JWS) [@!RFC7515]
   * `iat`: REQUIRED. Integer for the time at which the Credential Issuer Metadata was issued using the syntax defined in [@!RFC7519].
   * `exp`: OPTIONAL. Integer for the time at which the Credential Issuer Metadata is expiring, using the syntax defined in [@!RFC7519].
 
-The Wallet SHOULD establish trust in the signer of the metadata, and obtain the keys to validate the signature before processing the metadata, e.g. using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification. 
+The Wallet MUST establish trust in the signer of the metadata. Otherwise, the Wallet MUST reject the signed metadata. When validating the signature, the Wallet obtains the keys to validate the signature before processing the metadata, e.g. using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification.
 
 ### Credential Issuer Metadata Parameters {#credential-issuer-parameters}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1440,7 +1440,7 @@ The signed metadata MUST be secured using a JSON Web Signature (JWS) [@!RFC7515]
 
 * in the JOSE header,
   * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry [@IANA.JOSE]. It MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
-  * `typ`: REQUIRED. MUST be `key-attestation+jwt`, which explicitly types the key proof JWT as recommended in Section 3.11 of [@!RFC8725].
+  * `typ`: REQUIRED. MUST be `openidvci-issuer-metadata+jwt`, which explicitly types the key proof JWT as recommended in Section 3.11 of [@!RFC8725].
 
 * in the JWT body,
   * `iss`: OPTIONAL. String denoting the party attesting to the claims in the signed metadata
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1263,7 +1263,7 @@ The Credential Issuer MUST respond with HTTP Status Code 200 and return the Cred
 * an unsigned JSON document using the media type `application/json`, or
 * a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`.
 
-The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` header. However, the Credential Issuer MAY ignore the `Accept` header, e.g. if he doesn't support signed metadata or has another preference.
+The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` header. However, the Credential Issuer MAY ignore the `Accept` header, e.g. if it doesn't support signed metadata or has another preference.
 
 The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the language(s) preferred for display. It is up to the Credential Issuer whether to:
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1395,7 +1395,7 @@ Credential Issuers publishing metadata MUST make a JSON document available at th
 
 Communication with the Credential Issuer Metadata Endpoint MUST utilize TLS.
 
-To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept-Content` Header in the HTTP GET request to indicate whether it supports signed metadata.
+To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate whether it supports signed metadata.
 
 The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the [parameters defined in this specification](#credential-issuer-parameters) as either:
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1421,7 +1421,7 @@ Accept-Language: fr-ch, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
 
 ### Signed Metadata
 
-The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain the following elements:
+The signed metadata MUST be secured using a JSON Web Signature (JWS) [@!RFC7515] and contain the following elements:
 
 * in the JOSE header,
   * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry [@IANA.JOSE]. It MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1258,12 +1258,12 @@ Communication with the Credential Issuer Metadata Endpoint MUST utilize TLS.
 
 To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate the Content Type(s) it supports, and by doing so, signaling whether it supports signed metadata.
 
-The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the [parameters defined in this specification](#credential-issuer-parameters) as either:
+The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the parameters defined in (#credential-issuer-parameters) as either
 
-* an unsigned JSON document using the media type `application/json`
-* a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`
+* an unsigned JSON document using the media type `application/json`, or
+* a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`.
 
-The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` Header. However, the Credential Issuer MAY ignore the `Accept` Header.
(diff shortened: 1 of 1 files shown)


c2bo:
> * [ ]  discuss whether to use JWS instead of JWT for multiple trust frameworks

To give a bit more context on this: Paul and I had a bit of a discussion if we need to support multiple signatures similar to what was introduced in OpenID4VP. We came to the conclusion that starting with a JWT seems to be the more pragmatic choice and if we realise we need to support that feature, we could probably introduce it rather simple with a new media type (jose+json based). 

Using JWT and different Issuer deployments for different trust ecosystems seems reasonable to me.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1395,7 +1395,14 @@ Credential Issuers publishing metadata MUST make a JSON document available at th
...
 
-To fetch the Credential Issuer Metadata, the requester MUST send an HTTP request using the GET method and the path formed following the steps above. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and the HTTP Status Code 200.
+To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept-Content` Header in the HTTP GET request to indicate whether it supports signed metadata.
+
+The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the [parameters defined in this specification](#credential-issuer-parameters) as either:
+
+* an unsigned JSON document using the media type `application/json`
+* a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`
+
+The Credential Issuer MAY ignore the `Accept-Content` Header. It MUST indicate the type of the returned Credential Issuer Metadata using the HTTP `Content-Type`.
```
c2bo, line 1405: I know i kinda convinced you to remove a similar sentence, but it felt weird reading without any comment on what to do.

```suggestion
The Credential Issuer MAY ignore the `Accept-Content` Header. It MUST indicate the type of the returned Credential Issuer Metadata using the HTTP `Content-Type`.
Issuer and Wallet policies regarding signed or unsigned metadata are out of scope of this document and should be defined on a ecosystem level (e.g. via a Profile).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1412,6 +1419,22 @@ Host: server.example.com
 Accept-Language: fr-ch, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
 ```
 
+### Signed Metadata
+
+The signed metadata MUST be secured using JSON Web Signature (JWS) [@!RFC7515] and MUST contain the following elements:
```
c2bo:
```suggestion
The signed metadata MUST be secured using a JSON Web Signature (JWS) [@!RFC7515] and contain the following elements:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1412,6 +1419,22 @@ Host: server.example.com
...
+  * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry [@IANA.JOSE]. It MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
+  * `typ`: REQUIRED. MUST be `key-attestation+jwt`, which explicitly types the key proof JWT as recommended in Section 3.11 of [@!RFC8725].
+
+* in the JWT body,
+  * `iss`: OPTIONAL. String denoting the party attesting to the claims in the signed metadata
+  * `sub`: REQUIRED. String matching the Credential Issuer identifier
+  * `iat`: REQUIRED. Integer for the time at which the Credential Issuer Metadata was issued using the syntax defined in [@!RFC7519].
+  * `exp`: OPTIONAL. Integer for the time at which the Credential Issuer Metadata is expiring, using the syntax defined in [@!RFC7519].
+
+The Wallet SHOULD establish trust in the signer of the metadata, and obtain the keys to validate the signature before processing the metadata, e.g. using JOSE headers like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanism how to do that is out of scope of this specification.
```
c2bo, line 1436:
```suggestion
The Wallet SHOULD establish trust in the signer of the metadata, and obtain the keys to validate the signature before processing the metadata, e.g. using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification. 
```

openid-4-verifiable-credential-issuance-1_0.md
tplooker:
```suggestion
The Credential Issuer MAY ignore the `Accept` Header. It MUST indicate the type of the returned Credential Issuer Metadata using the HTTP `Content-Type`.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1395,7 +1395,14 @@ Credential Issuers publishing metadata MUST make a JSON document available at th
 
 Communication with the Credential Issuer Metadata Endpoint MUST utilize TLS.
 
-To fetch the Credential Issuer Metadata, the requester MUST send an HTTP request using the GET method and the path formed following the steps above. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and the HTTP Status Code 200.
+To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept-Content` Header in the HTTP GET request to indicate whether it supports signed metadata.
+
+The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the [parameters defined in this specification](#credential-issuer-parameters) as either:
```
tplooker: For interoperability, what are we normatively implying implementations must support, it would appear we are implying that all credential issuer implementations MUST support signed metadata or are we saying that only plain JSON MUST be supported and the client/wallet gets to decide what it wants to support?

paulbastian: The either to me implies that the issuer can choose. In the next section, it says the issuer may ignite the accept header and send whatever he likes, so to me it seems signed metadata is optional. That means wallets must support both.

tplooker: Im not sure I support making signed metadata required for wallet/client implementations I'd prefer we only require them to support unsigned which is consistent with the equivalent feature in OAuth and OpenID connect.

Sakurann: wg discussion. 
- no normative requirement for the wallet to support this.
- HAIP can mandate it.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1395,7 +1395,14 @@ Credential Issuers publishing metadata MUST make a JSON document available at th
 
 Communication with the Credential Issuer Metadata Endpoint MUST utilize TLS.
 
-To fetch the Credential Issuer Metadata, the requester MUST send an HTTP request using the GET method and the path formed following the steps above. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and the HTTP Status Code 200.
+To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate whether it supports signed metadata.
```
```suggestion
To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate the Content Type(s) it supports.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1395,7 +1395,14 @@ Credential Issuers publishing metadata MUST make a JSON document available at th
...
+
+The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the [parameters defined in this specification](#credential-issuer-parameters) as either:
+
+* an unsigned JSON document using the media type `application/json`
+* a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`
+
+The Credential Issuer MAY ignore the `Accept` Header. It MUST indicate the type of the returned Credential Issuer Metadata using the HTTP `Content-Type`.
```
```suggestion
The Credential Issuer MAY ignore the `Accept` Header. If the Credential Issuer receives a request with an `Accept` header set to both media types then it SHOULD respond with a signed metadata format if possible (ie. `application/jwt`).  It MUST indicate the type of the returned Credential Issuer Metadata using the HTTP `Content-Type`.
```

sloops77 (changes requested):
Using the Accept header allows the wallet to indicate support for multiple content types as well. These changes clarify how that should be handled

openid-4-verifiable-credential-issuance-1_0.md
lj-raidiam:
```suggestion
To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate the Content Type(s) it supports, and by doing so, signaling whether it supports signed metadata.
```

@sloops77 I let myself improve it further. Saying straight that the mime type is used only determine wallet support for metadata is not ideal, but there's a value in keeping it explicit too.

tplooker:
```suggestion
To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate the Content Type(s) it supports, and by doing so, signalling whether it supports signed metadata. If the `Accept` header features in the request, the Credential Issuer is RECOMMENDED to return a response with the requested Content Type. 
```

I think we need to be explicit here about what the credential issuer can do with the accept header, which ultimately they should reserve the right to ignore. This does raise an important question though for interoperability purposes I would propose as a common baseline that all credential issuers MUST support unsigned metadata if they support metadata at all, otherwise this mechanism wont allow implementations to reliably interoperate.

tplooker: I'll also point out that the way metadata and specifically signed metadata is handled in RFC8414 avoids this possible interoperability issue by requiring the response to always be application/json with a signed metadata element within it. Thus making plain JSON required if supporting the feature by AS's and signed metadata being optional.

paulbastian: @tplooker to your first comment: I moved the sentence further to the back, the first paragraph is about the Wallet's request and the Accept header

openid-4-verifiable-credential-issuance-1_0.md
paulbastian:
```suggestion
The Credential Issuer is RECOMMENDED to respond with a `Content-Type` matching to the Wallet's requested `Accept` Header. However, the Credential Issuer MAY ignore the `Accept` Header. In any way, it MUST indicate the type of the returned Credential Issuer Metadata using the HTTP `Content-Type`.
```

paulbastian:
WG discussion:
- to not make a breaking change, we would need to make unsigned metadata support mandatory for Credential Issuers supporting metadata
- different implementations between Credential Issuer and AS, discussion around the motivation from deviating from RFC8414 -> security concerns about not having a complete signed metadata document
- please comment and review about Accept/content-Type approach vs RFC8414 approach

charsleysa:
As I understand, one of the reasons behind this change is because of a concern that `signed_metadata` may not have all metadata included.

I think that would be easily solvable by mandating that all metadata (except `signed_metadata`) be present in the signed metadata and that wallets should use either the unsigned or the signed metadata and not mix the two.

If unsigned metadata support will be mandatory anyway, I don't see the need to have different methods of requesting signed metadata.

UBaggeler:
I like the Accept/content-Type approach. In addition to the aforementioned arguments, as discussed in https://github.com/openid/OpenID4VCI/issues/528#issuecomment-2958466264, this would allow for wallet apps and issuers to support multiple trust frameworks at the same time. E.g. using https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-jws-json-serialization as in OID4VP.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1427,6 +1434,22 @@ Host: server.example.com
 Accept-Language: fr-ch, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
 ```
 
+### Signed Metadata
+
+The signed metadata MUST be secured using a JSON Web Signature (JWS) [@!RFC7515] and contain the following elements:
+
+* in the JOSE header,
+  * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry [@IANA.JOSE]. It MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
+  * `typ`: REQUIRED. MUST be `key-attestation+jwt`, which explicitly types the key proof JWT as recommended in Section 3.11 of [@!RFC8725].
```
paulbastian:
```suggestion
  * `typ`: REQUIRED. MUST be `openidvci-issuer-metadata+jwt`, which explicitly types the key proof JWT as recommended in Section 3.11 of [@!RFC8725].
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1410,14 +1410,14 @@ For example, the metadata for the Credential Issuer Identifier `https://issuer.e
...
+To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate the Content Type(s) it supports, and by doing so, signaling whether it supports signed metadata.
 
 The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the [parameters defined in this specification](#credential-issuer-parameters) as either:
 
 * an unsigned JSON document using the media type `application/json`
 * a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`
 
-The Credential Issuer MAY ignore the `Accept` Header. It MUST indicate the type of the returned Credential Issuer Metadata using the HTTP `Content-Type`.
+The Credential Issuer is RECOMMENDED to respond with a `Content-Type` matching to the Wallet's requested `Accept` Header. However, the Credential Issuer MAY ignore the `Accept` Header. In any way, it MUST indicate the type of the returned Credential Issuer Metadata using the HTTP `Content-Type`.
```
```suggestion
The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type`. It is RECOMMENDED to respond with a `Content-Type` matching to the Wallet's requested `Accept` Header. However, the Credential Issuer MAY ignore the `Accept` Header.
```

sloops77 (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1273,6 +1280,22 @@ Host: server.example.com
 Accept-Language: fr-ch, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
 ```
 
+### Signed Metadata
+
+The signed metadata MUST be secured using a JSON Web Signature (JWS) [@!RFC7515] and contain the following elements:
+
+* in the JOSE header,
+  * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry [@IANA.JOSE]. It MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
+  * `typ`: REQUIRED. MUST be `openidvci-issuer-metadata+jwt`, which explicitly types the key proof JWT as recommended in Section 3.11 of [@!RFC8725].
```
c2bo, line 1289: We are missing the IANA registration part for this typ aren't we?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1273,6 +1280,22 @@ Host: server.example.com
...
+
+* in the JWT body,
+  * `iss`: OPTIONAL. String denoting the party attesting to the claims in the signed metadata
+  * `sub`: REQUIRED. String matching the Credential Issuer identifier
+  * `iat`: REQUIRED. Integer for the time at which the Credential Issuer Metadata was issued using the syntax defined in [@!RFC7519].
+  * `exp`: OPTIONAL. Integer for the time at which the Credential Issuer Metadata is expiring, using the syntax defined in [@!RFC7519].
+
+The Wallet SHOULD establish trust in the signer of the metadata, and obtain the keys to validate the signature before processing the metadata, e.g. using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification.
```
c2bo: Would HAIP then define how exactly the binding works for e.g., x5c?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1256,7 +1256,14 @@ For example, the metadata for the Credential Issuer Identifier `https://issuer.e
...
+To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate the Content Type(s) it supports, and by doing so, signaling whether it supports signed metadata.
+
+The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the [parameters defined in this specification](#credential-issuer-parameters) as either:
+
+* an unsigned JSON document using the media type `application/json`
+* a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`
+
+The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type`. It is RECOMMENDED to respond with a `Content-Type` matching to the Wallet's requested `Accept` Header. However, the Credential Issuer MAY ignore the `Accept` Header.
```
```suggestion
The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` Header. However, the Credential Issuer MAY ignore the `Accept` Header.
```

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: I added it

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1256,7 +1256,14 @@ For example, the metadata for the Credential Issuer Identifier `https://issuer.e
 
 Communication with the Credential Issuer Metadata Endpoint MUST utilize TLS.
 
-To fetch the Credential Issuer Metadata, the requester MUST send an HTTP request using the GET method and the path formed following the steps above. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and the HTTP Status Code 200.
+To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate the Content Type(s) it supports, and by doing so, signaling whether it supports signed metadata.
+
+The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the [parameters defined in this specification](#credential-issuer-parameters) as either:
```
danielfett:
```suggestion
The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the parameters defined in (#credential-issuer-parameters) as either
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1256,7 +1256,14 @@ For example, the metadata for the Credential Issuer Identifier `https://issuer.e
...
+
+* an unsigned JSON document using the media type `application/json`
+* a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`
```
danielfett:
```suggestion
* an unsigned JSON document using the media type `application/json`, or
* a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1256,7 +1256,14 @@ For example, the metadata for the Credential Issuer Identifier `https://issuer.e
...
+
+The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` Header. However, the Credential Issuer MAY ignore the `Accept` Header.
```
danielfett:
```suggestion
The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` header. However, the Credential Issuer MAY ignore the `Accept` header, e.g. if he doesn't support signed metadata or has another preference.
```

danielfett: Under which conditions may the header be ignored?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1273,6 +1280,22 @@ Host: server.example.com
...
+  * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry [@IANA.JOSE]. It MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
+  * `typ`: REQUIRED. MUST be `openidvci-issuer-metadata+jwt`, which explicitly types the key proof JWT as recommended in Section 3.11 of [@!RFC8725].
+
+* in the JWT body,
+  * `iss`: OPTIONAL. String denoting the party attesting to the claims in the signed metadata
+  * `sub`: REQUIRED. String matching the Credential Issuer identifier
+  * `iat`: REQUIRED. Integer for the time at which the Credential Issuer Metadata was issued using the syntax defined in [@!RFC7519].
+  * `exp`: OPTIONAL. Integer for the time at which the Credential Issuer Metadata is expiring, using the syntax defined in [@!RFC7519].
+
+The Wallet SHOULD establish trust in the signer of the metadata, and obtain the keys to validate the signature before processing the metadata, e.g. using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification.
```
danielfett: Why not "MUST establish trust"?

danielfett:
```suggestion
The Wallet SHOULD establish trust in the signer of the metadata, and obtain the keys to validate the signature before processing the metadata, e.g., using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification. 
```

openid-4-verifiable-credential-issuance-1_0.md
c2bo: We are also using upper-case Header below - but I agree that it should probably be lower case?

openid-4-verifiable-credential-issuance-1_0.md
c2bo: I wondered about the same question - at the end of the day it probably comes down to ecosystem/wallet policy similar to the formulation in VP? Would something like this work?

```suggestion
The Wallet SHOULD verify the signature of the JWT and establish trust in the signer of the metadata, and obtain the keys to validate the signature before processing the metadata, e.g., by using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification. 

Factors that influence the Wallet's decision whether to establish trust in the signed metadata and verify the signature include, but are not limited to, trust framework the Wallet supports, specific policies defined by the ecosystem, and profiles of this specification.
```

paulbastian: To me MUST would feel to limiting in OpenID4VCI, could be mandated by profiles. I would be fine with suggestion from @c2bo

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1256,7 +1256,14 @@ For example, the metadata for the Credential Issuer Identifier `https://issuer.e
...
+To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate the Content Type(s) it supports, and by doing so, signaling whether it supports signed metadata.
+
+The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the parameters defined in (#credential-issuer-parameters) as either
+
+* an unsigned JSON document using the media type `application/json`, or
+* a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`.
+
+The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` header. However, the Credential Issuer MAY ignore the `Accept` header, e.g. if he doesn't support signed metadata or has another preference.
```
```suggestion
The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` header. However, the Credential Issuer MAY ignore the `Accept` header, e.g. if it doesn't support signed metadata or has another preference.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1256,7 +1256,14 @@ For example, the metadata for the Credential Issuer Identifier `https://issuer.e
...
+To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate the Content Type(s) it supports, and by doing so, signaling whether it supports signed metadata.
+
+The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the parameters defined in (#credential-issuer-parameters) as either
+
+* an unsigned JSON document using the media type `application/json`, or
+* a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`.
+
+The Credential Issuer MUST indicate the media type of the returned Credential Issuer Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` header. However, the Credential Issuer MAY ignore the `Accept` header, e.g. if it doesn't support signed metadata or has another preference.
```
```suggestion
The Credential Issuer MUST support returning metadata in an unsigned form 'application/json' and MAY support returning it in a signed form 'application/jwt'. In all cases the Credential Issuer MUST indicate the media type of the returned Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` header when the requested content type is supported.
```

tlodderstedt: lgtm

openid-4-verifiable-credential-issuance-1_0.md
tlodderstedt: I suggest to make it a MUST. If the wallet does not know how to process, it MUST fail.

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: This changes imo the use case where an Issuer mandates a trust framework but the Wallet does not. In this case the Wallet will fail a little bit later but still fail. I think it doesn't change much in the end.

openid-4-verifiable-credential-issuance-1_0.md
Sakurann: wg discussion

```suggestion
The Wallet MUST establish trust in the signer of the metadata. Otherwise, the Wallet MUST reject the signed metadata. When validating the signature, the Wallet obtains the keys to validate the signature before processing the metadata, e.g. using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1273,6 +1280,22 @@ Host: server.example.com
...
+  * `typ`: REQUIRED. MUST be `openidvci-issuer-metadata+jwt`, which explicitly types the key proof JWT as recommended in Section 3.11 of [@!RFC8725].
+
+* in the JWT body,
+  * `iss`: OPTIONAL. String denoting the party attesting to the claims in the signed metadata
+  * `sub`: REQUIRED. String matching the Credential Issuer identifier
+  * `iat`: REQUIRED. Integer for the time at which the Credential Issuer Metadata was issued using the syntax defined in [@!RFC7519].
+  * `exp`: OPTIONAL. Integer for the time at which the Credential Issuer Metadata is expiring, using the syntax defined in [@!RFC7519].
+
+The Wallet MUST establish trust in the signer of the metadata. Otherwise, the Wallet MUST reject the signed metadata. When validating the signature, the Wallet obtains the keys to validate the signature before processing the metadata, e.g. using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification.
```
```suggestion
When requesting signed metadata, the Wallet MUST establish trust in the signer of the metadata. Otherwise, the Wallet MUST reject the signed metadata. When validating the signature, the Wallet obtains the keys to validate the signature before processing the metadata, e.g. using JOSE header parameters like `x5c`, `kid` or `trust_chain` to convey the public key. The concrete mechanisms how to do that are out of scope of this specification.
```

tlodderstedt (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1273,6 +1280,22 @@ Host: server.example.com
...
+
+The signed metadata MUST be secured using a JSON Web Signature (JWS) [@!RFC7515] and contain the following elements:
+
+* in the JOSE header,
+  * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry [@IANA.JOSE]. It MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
+  * `typ`: REQUIRED. MUST be `openidvci-issuer-metadata+jwt`, which explicitly types the key proof JWT as recommended in Section 3.11 of [@!RFC8725].
+
+* in the JWT body,
+  * `iss`: OPTIONAL. String denoting the party attesting to the claims in the signed metadata
+  * `sub`: REQUIRED. String matching the Credential Issuer identifier
```
dastoikov:
```suggestion
  * `sub`: REQUIRED. String matching the Credential Issuer Identifier
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1256,7 +1256,14 @@ For example, the metadata for the Credential Issuer Identifier `https://issuer.e
 
 Communication with the Credential Issuer Metadata Endpoint MUST utilize TLS.
 
-To fetch the Credential Issuer Metadata, the requester MUST send an HTTP request using the GET method and the path formed following the steps above. The Credential Issuer MUST return a JSON document compliant with this specification using the `application/json` media type and the HTTP Status Code 200.
+To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` Header in the HTTP GET request to indicate the Content Type(s) it supports, and by doing so, signaling whether it supports signed metadata.
```
c2bo:
```suggestion
To fetch the Credential Issuer Metadata, the Wallet MUST send an HTTP request using the GET method and the path formed following the steps above. The Wallet is RECOMMENDED to send an `Accept` header in the HTTP GET request to indicate the Content Type(s) it supports, and by doing so, signaling whether it supports signed metadata.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1256,7 +1256,14 @@ For example, the metadata for the Credential Issuer Identifier `https://issuer.e
...
+
+The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the parameters defined in (#credential-issuer-parameters) as either
+
+* an unsigned JSON document using the media type `application/json`, or
+* a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type `application/jwt`.
+
+The Credential Issuer MUST support returning metadata in an unsigned form 'application/json' and MAY support returning it in a signed form 'application/jwt'. In all cases the Credential Issuer MUST indicate the media type of the returned Metadata using the HTTP `Content-Type` header. It is RECOMMENDED for Credential Issuers to respond with a `Content-Type` matching to the Wallet's requested `Accept` header when the requested content type is supported.
 
 The Wallet is RECOMMENDED to send an `Accept-Language` Header in the HTTP GET request to indicate the language(s) preferred for display. It is up to the Credential Issuer whether to:
```
c2bo:
```suggestion
The Wallet is RECOMMENDED to send an `Accept-Language` header in the HTTP GET request to indicate the language(s) preferred for display. It is up to the Credential Issuer whether to:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1273,6 +1280,22 @@ Host: server.example.com
 Accept-Language: fr-ch, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5
 ```
 
+### Signed Metadata {#credential-issuer-signed-metadata}
+
+The signed metadata MUST be secured using a JSON Web Signature (JWS) [@!RFC7515] and contain the following elements:
+
+* in the JOSE header,
+  * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry [@IANA.JOSE]. It MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
```
dastoikov, line 1288: Question: is there a way for a client requesting signed metadata to check if the server’s supported signing algorithms would match the former’s capabilities, and potentially negotiate preferences?

c2bo (approved):

openid-4-verifiable-credential-issuance-1_0.md
jogu: I think we wanted to keep it simple to start with (that's how the oauth signed metadata and federation work too I believe) - it is possible that we could define a future mechanism using a header or where the issuer returns a json serialised response with multiple signatures.

dastoikov: Thank you, @jogu!

PR #524, opened 2025-05-28
Attempt to clarify intro text about key attestations

### LINKED ISSUE #463: Clarification about use of key attestations in JWT proofs
Need some clarification about the usage of key attestations in JWT proof type.

Spec mentions the following text in section [jwt Proof Type](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#section-8.2.1.1)

> The Credential Issuer SHOULD issue a Credential for each cryptographic public key specified in the attested_keys claim within the key_attestation parameter.

I am bit confuzed on what is the intention behind this. My understanding is that in the case of JWT proofs issuer should issue one credential per jwt proof sent by the wallet. Why should this change if key attestation is used in the header of the JWT proof?

jogu:
The singular "key" was being used in a way that could be confusing given a single key attestation can be used to attest multiple keys.

closes #463

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -922,7 +922,7 @@ There are two ways to convey key attestations (as defined in (#keyattestation)) 
 - The Wallet uses the `jwt` proof type in the Credential Request to create a proof of possession for one of the attested keys and adds the key attestation in the JOSE header.
 - The Wallet uses the `attestation` proof type in the Credential Request to provide a key attestation without a proof of possession of any of the keys.
 
-Depending on the Wallet's implementation, the `attestation` may avoid unnecessary End-User interaction during Credential issuance, as the key to which the Credential will be bound does not necessarily need to perform signature operations.
+Depending on the Wallet's implementation, the `attestation` may avoid unnecessary End-User interaction during Credential issuance, as the key(s) to which the Credential(s) will be bound does not necessarily need to perform signature operations, and one key attestation can be used to attest multiple keys.
 
 Additional proof types MAY be defined and used.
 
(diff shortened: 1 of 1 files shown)


c2bo (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -919,10 +919,10 @@ This specification defines the following proof types:
 
 There are two ways to convey key attestations (as defined in (#keyattestation)) of the cryptographic key material during Credential issuance:
 
-- The Wallet uses the `jwt` proof type in the Credential Request to create a proof of possession of the key and adds the key attestation in the JOSE header.
-- The Wallet uses the `attestation` proof type in the Credential Request with the key attestation without a proof of possession of the key itself.
+- The Wallet uses the `jwt` proof type in the Credential Request to create a proof of possession for one of the attested keys and adds the key attestation in the JOSE header.
```
Sakurann, line 922:
```suggestion
- The Wallet uses the `jwt` proof type in the Credential Request to create a proof of possession for one of the attested keys and adds the key attestation with all of the attested keys in the JOSE header.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -919,10 +919,10 @@ This specification defines the following proof types:
...
+- The Wallet uses the `attestation` proof type in the Credential Request to provide a key attestation without a proof of possession of any of the keys.
 
-Depending on the Wallet's implementation, the `attestation` may avoid unnecessary End-User interaction during Credential issuance, as the key itself does not necessarily need to perform signature operations.
+Depending on the Wallet's implementation, the `attestation` may avoid unnecessary End-User interaction during Credential issuance, as the key to which the credential will be bound does not necessarily need to perform signature operations.
```
Sakurann, line 925:
```suggestion
Depending on the Wallet's implementation, the `attestation` may avoid unnecessary End-User interaction during Credential issuance, as the key to which the Credential will be bound does not necessarily need to perform signature operations.
```

openid-4-verifiable-credential-issuance-1_0.md
tplooker: I agree with @jogu, that suggestion is clearer than this proposed language.

Issue #496, opened 2025-05-08
is c_nonce a nonce or a challenge?


Sakurann:
"or clarify that c_nonce stands for challenge nonce?"
"if it's not used once, it should not be called nonce" 
-@panva 

personally, don't think this is worth a breaking change of renaming `c_nonce` but we probably should add a clarification that this is closer to a challenge or something. since this has historically caused a tremendous confusion

patatoid:
According to my understanding, a nonce is to ensure the requested resource correspond to the actual request.

I would also see the described c_nonce as a `code_challenge` / `code_verifier` with the plain method even it is not linked to the authorization request but to the verifier.

Sakurann:
no one seem to want it

Issue #491, opened 2025-05-07
Nonce Endpoint for Authorization Server


paulbastian:
Given our latest proposal for [nonce fetching for Wallet Attestation](https://github.com/oauth-wg/draft-ietf-oauth-attestation-based-client-auth/issues/110), we propose to have a nonce endpoint for the Authorization Server in OpenID4VCI. Our reasoing:
- we already have a nonce endpoint for Credential Issuer
- we can use the nonce endpoint at AS to get nonces for wallet attestation before calling the PAR endpoint
- we can fetch DPoP nonces also at AS, which would bring us on par with pending PR #478 
- if AS and Credential Issuer are the same, they can have a merged nonce endpoint
- in the long run extract the ideas from OpenID4VCI and Attestation-Based Client Authentication into a dedicated OAuth Draft in backwards-compatible fashion?

tplooker:
Im supportive of this, having multiple nonce endpoints creates quite a lot of implementation difficulty.

Sakurann:
WG discussion:
- Current situation: 
    - all the nonces in the flow: wallet attestation, DPoP in AS and c_nonce (proof and key attestation) and DPoP in RS
    - existing mechanisms: nonce endpoint for c_nonce in RS; the same endpoint can be used for DPoP nonce at RS
- what remains to be specced: two nonces at AS
   - want to also improve how DPoP nonces work in general (and more deterministic) (wallet would still have to deal with DPoP nonce in the error)
- direction
    - want two separate nonce endpoints for the case where AS and RS are different entities. nonce endpoint preferred to http OPTIONS
    - have a symmetrical design between AS and RS: both have a nonce endpoint that is capable of returning DPoP-nonce header
- **next steps:**
    - VCI relies on IETF attestation based client authn draft for wallet attestation. that IETF draft will add a mechanism to get a challenge. VCI to reference that mechanism and add an option to add DPoP-nonce header in VCI. consider calling nonce a challenge in IETF
    - #496 as a spill over
    - Add a section that clarifies what nonces are in VCI and how to fetch them and what to be concerned about.

Sakurann:
marking 1.1, since we can add this in a non-breaking manner in 1.1

GarethCOliver:
Poking this issue: attestation based client authn draft does now have the Challenge endpoint specified. It'd be good to incorporate this in either 1.1 or failing that 1.2

c2bo:
[...]

Issue #461, opened 2025-02-17
Protect the nonce endpoint


andprian:
Having this endpoint unprotected is not a good idea:

- If the Issuer returns the same nonce every time this makes the nonce publicly available to anyone, which makes it completely useless. Moreover, clause 7.2 states that "This value MUST be unpredictable." so I am not sure if it is therefore required to have a new nonce with every call.
- If the Issuer returns a different nonce with every call this makes it an attack vector because the nonce database would grow continuously and the service could be vulnerable to a DoS. This would also make it difficult to search for a nonce when receiving it in the credential request.

This endpoint should be protected with the access token that the wallet just obtained, just like the credential endpoint. Thus, the nonce would be also linked to a specific wallet which would make it simple for the Issuer to match the nonce when receiving it in the subsequent credential request.
 
Also, it would make more sense for this endpoint to implement a GET method instead of a POST.

c2bo:
There was some discussion on this in the prior issue/PR: #404

bc-pi:
Also some discussion in #381 - in particular this part https://github.com/openid/OpenID4VCI/pull/381#discussion_r1777355851 about GET vs POST and https://github.com/openid/OpenID4VCI/pull/381#issuecomment-2361382383 on protection of the endpoint.

oriolcanades:
My 2 cents. I checked the comments in #381 and the current draft. If the only way to use a nonce endpoint—where a fresh c_nonce value can be obtained—is for proof of possession of key material in a subsequent request to the Credential Endpoint, and this call is always made after the Token request, then why don’t we secure the nonce endpoint? Are there other use cases where we need a nonce before acquiring the Token?

andprian:
@oriolcanades I do not see the need to ask for a nonce if you don't have a previously obtained access token.

Sakurann:
WG discussion:
- nonce endpoint does not mean an exponentially growing nonce data bases (mechanisms can be self-contained nonces or fully fledged database or etc.)
- nonce is used only for the freshness according to the requirement to the servers (which can be very strict) as opposed to being unique to a session. better name would have been "challenge" (probably ship has sailed.. ) 
- can't protect nonce endpoint with a sender-constrained access token with DPoP since that would require a nonce
- for the above reasons, rough consensus is still that there is no need to protect this endpoint

andprian:
I am not sure I understand the reasons described.

[...]

PR #525, opened 2025-05-28
Cleanup language around c_nonce

jogu:
We still had text that assumed the c_nonce was proactively sent to the wallet, rather than (as is now the case) fetched by the wallet.

Also clarify that the nonce is a challenge rather than a nonce (without changing any parameter names!), and that 'unpredictable' does not mean that the same value cannot be used for a short time.

Some of the language is quite similar to the language in DPoP around the DPoP nonces, as there is a lot of similarity - both mechanisms do not (fully) prevent replay but do ensure freshness.

relates to some of the points raised in #461

relates the confusion that prompted #496

adds some clarify around the 'what nonces are' part of #491

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -814,7 +814,7 @@ See (#identifying_credential) for the summary of the options how requested Crede
 
 The proof type contained in the `proofs` parameter is an extension point that enables the use of different types of proofs for different cryptographic schemes.
 
-The proof(s) in the `proofs` parameter MUST incorporate the Credential Issuer Identifier (audience) and, if the Credential has a Nonce Endpoint, a `c_nonce` value to allow the Credential Issuer to detect freshness. The way that data is incorporated depends on the key proof type. In a JWT, for example, the `c_nonce` value is conveyed in the `nonce` claim, whereas the audience is conveyed in the `aud` claim. In a Linked Data proof, for example, the `c_nonce` is included as the `challenge` element in the key proof object and the Credential Issuer (the intended audience) is included as the `domain` element.
+The proof(s) in the `proofs` parameter MUST incorporate the Credential Issuer Identifier (audience) and, if the Credential Issuer has a Nonce Endpoint, a `c_nonce` value to allow the Credential Issuer to detect freshness. The way that data is incorporated depends on the key proof type. In a JWT, for example, the `c_nonce` value is conveyed in the `nonce` claim, whereas the audience is conveyed in the `aud` claim. In a Linked Data proof, for example, the `c_nonce` is included as the `challenge` element in the key proof object and the Credential Issuer (the intended audience) is included as the `domain` element.
 
 The `proofs` parameter MUST be present if the `proof_types_supported` parameter is present in the `credential_configurations_supported` parameter of the Issuer metadata for the requested Credential.
 


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1053,7 +1053,7 @@ Below is a non-normative example of a `proofs` parameter:
 
 A key attestation in JWT format as defined in (#keyattestation-jwt).
 
-When a key attestation is used as a proof type, it MUST contain the `c_nonce` value provided by the Credential Issuer in its `nonce` parameter.
+If the Credential Issuer has a Nonce Endpoint (as defined in (#nonce-endpoint)), the `c_nonce` value provided by the Credential Issuer MUST be provided in the key attestation's `nonce` parameter.
```
paulbastian, line 1056: This is a normative change. Should we mandate nonce endpoint in this case?

tplooker: I agree its better to be consistent here unless there is compelling reason not to be.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -760,9 +760,9 @@ Content-Length: 0
 
 The Credential Issuer provides a nonce value in the HTTP response with a 2xx status code and the following parameters included as top-level members in the message body of the HTTP response using the application/json media type:
 
-* `c_nonce`: REQUIRED. String containing a nonce to be used when creating a proof of possession of the key proof (see (#credential-request)). This value MUST be unpredictable.
+* `c_nonce`: REQUIRED. String containing a challenge to be used when creating a proof of possession of the key proof (see (#credential-request)). This value MUST be unpredictable, though the Credential Issue MAY use the same value for a time period of it's choosing.
```
andprian: For me there is a contradiction here. Since the nonce endpoint is unprotected and open to everyone, and the same nonce can be reused for any period of time, this is clearly not unpredictable. On the contrary, this is a publicly known value, possibly for a long period of time.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1419,7 +1419,7 @@ This specification defines the following Credential Issuer Metadata parameters:
...
 * `credential_endpoint`: REQUIRED. URL of the Credential Issuer's Credential Endpoint, as defined in (#credential-request). This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components.
-* `nonce_endpoint`: OPTIONAL. URL of the Credential Issuer's Nonce Endpoint, as defined in (#nonce-endpoint). This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Credential Issuer does not support the Nonce Endpoint.
+* `nonce_endpoint`: OPTIONAL. URL of the Credential Issuer's Nonce Endpoint, as defined in (#nonce-endpoint). This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Credential Issuer does not require the use of `c_nonce`.
```
andprian, line 1438: Maybe state that if this is present, the wallet must use this endpoint. I see indirect requirements in various places, for the wallet to use the same value as given in the nonce endpoint but I think we should clearly state that this is a mandatory feature for the wallet, when the Issuer supports it.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -820,11 +820,11 @@ See (#identifying_credential) for the summary of the options how requested Crede
...
+The proof(s) in the `proofs` parameter MUST incorporate the Credential Issuer Identifier (audience), and optionally a `c_nonce` value generated by the Credential Issuer to allow the Credential Issuer to detect freshness. The way that data is incorporated depends on the key proof type. In a JWT, for example, the `c_nonce` value is conveyed in the `nonce` claim, whereas the audience is conveyed in the `aud` claim. In a Linked Data proof, for example, the `c_nonce` is included as the `challenge` element in the key proof object and the Credential Issuer (the intended audience) is included as the `domain` element.
```
andprian: Having a sentence saying "MUST ... optionally" is a bit strange. I suggest to split in 2 phrases, one with the mandatory audience and one with an optional nonce.

openid-4-verifiable-credential-issuance-1_0.md
andprian: I think we can delete the word "proof" after "key".
"proof of possession of the key" would suffice

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1531,7 +1531,7 @@ The Wallet is supposed to detect signs of fraudulent behavior related to the Cre
...
+If an adversary obtains a key proof (as outlined in #proof-types), they could potentially have a Credential issued that is linked to a key pair controlled by the victim. The `c_nonce` parameter serves as the main defense against pre-generation of key proofs, whilst also limiting the time window for replay attacks. It is RECOMMENDED that Credential Issuers utilize the Nonce Endpoint as specified in (#nonce-endpoint). A Wallet can continue using a given nonce until it is rejected by the Credential Issuer. The Credential Issuer determines for how long a particular nonce can be used. Servers MUST establish a clear policy on whether the same key proof can be reused and for how long, or if each Credential Request requires a new key proof.
```
andprian: `serves as the main defense against pre-generation of key proofs, whilst also limiting the time window for replay attacks.`

Since there is not limitation on reusing the same nonce, on responding with the same nonce no matter the caller, or even a time limitation mandated, this is not necessarily true. It could be, depending on the implementation. But it is not true as a general statement.

openid-4-verifiable-credential-issuance-1_0.md
jogu: Can you suggest a better wording?

tplooker: I think the focus on the value being unpredictable is accurate albeit it a little difficult to understand initially. [DPoP](https://datatracker.ietf.org/doc/html/rfc9449) also describes nonces in the same way which is conceptually similar, what about about the following revision?

```suggestion
* `c_nonce`: REQUIRED. String containing a challenge to be used when creating a proof of possession of the key proof (see (#credential-request)). This value MUST be unpredictable and it is up to the Credential Issuer to decide when to return a new challenge value as oppose to the one returned in the previous request.
```

openid-4-verifiable-credential-issuance-1_0.md
```suggestion
```

I suggest we just remove this paragraph entirely as each concrete proof type already defines this explicitly and this prevents a situation in future where a new proof type is defined that perhaps couldn't satisfy these constraints (albeit unlikely).

openid-4-verifiable-credential-issuance-1_0.md
tplooker:
```suggestion
If an adversary obtains a key proof (as outlined in #proof-types), they could potentially replay such proof to a Credential Issuer to obtain a duplicate Credential or multiple duplicates issued to the same key. The `c_nonce` parameter serves as the main defense against this, by providing Credential Issuers with a mechanism to determine freshness of a generated proof and thus reject a proof when required. It is RECOMMENDED that Credential Issuers utilize the Nonce Endpoint as specified in (#nonce-endpoint) to protect against this. A Wallet can continue using a given nonce until it is rejected by the Credential Issuer. The Credential Issuer determines for how long a particular nonce can be used.
```

tplooker: Few suggestions here, the risk isn't key substitution AFAIK, its replaying the key proof and getting multiple duplicates back. Also removed the last sentence as generally the re-use risk is about the nonce/challenge, not the key proof and that is already covered earlier in the paragraph.

openid-4-verifiable-credential-issuance-1_0.md
andprian: @tplooker are you saying that the Credential Issuer can decide to return the nonce returned in a previous response and at the same time the nonce is unpredictable?

jogu: > are you saying that the Credential Issuer can decide to return the nonce returned in a previous response and at the same time the nonce is unpredictable?

@andprian Yes. When you pick a new value, the new value selected must not be predictable. How long you use that new value for is (currently) a matter of issuer policy.

Is the expectation now clear to you? Do you have any suggestions to improve the wording?

This is the current situation and this PR only aims to more clearly document the current situation without normative changes. If we need normative changes I think those will most likely come about as a result of further discussion of #461, though if you see other issues please do open an issue for them.

jogu: Another suggestion building slightly on Tobias's (and also removing the superfluous 'proof' andprian spotted below):

```suggestion
* `c_nonce`: REQUIRED. String containing a challenge to be used when creating a proof of possession of the key (see (#credential-request)). It is at the discretion of the Credential Issuer when to return a new challenge value as opposed to the one returned in the previous request. New challenge values MUST be unpredictable.
```

paulbastian: I would agree here, unpredictable means that nobody should be able to generate c_nonce for the future, therefore it guarantees freshness. If the c_nonce is also used to create strong replay attack prevention is up to the issuer. To be honest, the stakes are not too high for proof in the Credential Request imo

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: We don't have any freshness guarantee with this change. Yet again, it's also optional for jwt proof type. So I think I'm okay with this. 

cc @c2bo

openid-4-verifiable-credential-issuance-1_0.md
jogu: Removed the optionally:

```suggestion
The proof(s) in the `proofs` parameter MUST incorporate the Credential Issuer Identifier (audience) and, if the Credential has a `nonce_endpoint`, a `c_nonce` value to allow the Credential Issuer to detect freshness. The way that data is incorporated depends on the key proof type. In a JWT, for example, the `c_nonce` value is conveyed in the `nonce` claim, whereas the audience is conveyed in the `aud` claim. In a Linked Data proof, for example, the `c_nonce` is included as the `challenge` element in the key proof object and the Credential Issuer (the intended audience) is included as the `domain` element.
```

jogu: I also wondered about that whilst I was editing the paragraph, but in the end I think it's setting criteria for people that want to define new proof types, and "incorporating" (which doesn't necessarily mean "has a field that contains exactly") both the audience and potentially a freshness seem fairly straightforward and sensible requirements. Not a hill I'm going to die on though :)

openid-4-verifiable-credential-issuance-1_0.md
andprian: > > are you saying that the Credential Issuer can decide to return the nonce returned in a previous response and at the same time the nonce is unpredictable?
> 
> @andprian Yes. When you pick a new value, the new value selected must not be predictable. How long you use that new value for is (currently) a matter of issuer policy.
> 
> Is the expectation now clear to you? Do you have any suggestions to improve the wording?
> 
> This is the current situation and this PR only aims to more clearly document the current situation without normative changes. If we need normative changes I think those will most likely come about as a result of further discussion of #461, though if you see other issues please do open an issue for them.

@jogu , I was not requesting normative changes, I was requesting to remove the contradiction. This should be clear not just for us, but for all the people that will read this spec, whether familiar with it or not.
I am ok with the text you suggested, because now it states _in which cases_ the value must be unpredictable.

jogu: Great, thank you @andprian ! I'll apply the above suggestion then.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -820,11 +820,11 @@ See (#identifying_credential) for the summary of the options how requested Crede
...
+The proof(s) in the `proofs` parameter MUST incorporate the Credential Issuer Identifier (audience) and, if the Credential has a `nonce_endpoint`, a `c_nonce` value to allow the Credential Issuer to detect freshness. The way that data is incorporated depends on the key proof type. In a JWT, for example, the `c_nonce` value is conveyed in the `nonce` claim, whereas the audience is conveyed in the `aud` claim. In a Linked Data proof, for example, the `c_nonce` is included as the `challenge` element in the key proof object and the Credential Issuer (the intended audience) is included as the `domain` element.
```
jogu: for consistency:

```suggestion
The proof(s) in the `proofs` parameter MUST incorporate the Credential Issuer Identifier (audience) and, if the Credential has a Nonce Endpoint, a `c_nonce` value to allow the Credential Issuer to detect freshness. The way that data is incorporated depends on the key proof type. In a JWT, for example, the `c_nonce` value is conveyed in the `nonce` claim, whereas the audience is conveyed in the `aud` claim. In a Linked Data proof, for example, the `c_nonce` is included as the `challenge` element in the key proof object and the Credential Issuer (the intended audience) is included as the `domain` element.
```

charsleysa (approved):

openid-4-verifiable-credential-issuance-1_0.md
tplooker: Yeah I'm ok with the restructured language you have offered and therefore keeping the paragraph in now.

tplooker (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -814,11 +814,11 @@ See (#identifying_credential) for the summary of the options how requested Crede
...
+The proof(s) in the `proofs` parameter MUST incorporate the Credential Issuer Identifier (audience) and, if the Credential has a Nonce Endpoint, a `c_nonce` value to allow the Credential Issuer to detect freshness. The way that data is incorporated depends on the key proof type. In a JWT, for example, the `c_nonce` value is conveyed in the `nonce` claim, whereas the audience is conveyed in the `aud` claim. In a Linked Data proof, for example, the `c_nonce` is included as the `challenge` element in the key proof object and the Credential Issuer (the intended audience) is included as the `domain` element.
```
```suggestion
The proof(s) in the `proofs` parameter MUST incorporate the Credential Issuer Identifier (audience) and, if the Credential Issuer has a Nonce Endpoint, a `c_nonce` value to allow the Credential Issuer to detect freshness. The way that data is incorporated depends on the key proof type. In a JWT, for example, the `c_nonce` value is conveyed in the `nonce` claim, whereas the audience is conveyed in the `aud` claim. In a Linked Data proof, for example, the `c_nonce` is included as the `challenge` element in the key proof object and the Credential Issuer (the intended audience) is included as the `domain` element.
```

openid-4-verifiable-credential-issuance-1_0.md
jogu: I don't think the only risk is duplicate credentials - the key proof isn't bound to the credential identifier nor even the session. So I suggest keeping the first sentence as is:

```suggestion
If an adversary obtains a key proof (as outlined in #proof-types), they could potentially replay such proof to have a Credential issued that is linked to a key pair controlled by the victim. The `c_nonce` parameter serves as the main defense against this, by providing Credential Issuers with a mechanism to determine freshness of a generated proof and thus reject a proof when required. It is RECOMMENDED that Credential Issuers utilize the Nonce Endpoint as specified in (#nonce-endpoint) to protect against this. A Wallet can continue using a given nonce until it is rejected by the Credential Issuer. The Credential Issuer determines for how long a particular nonce can be used.
```

jogu:
3 approvals one this one (third one is here: https://github.com/openid/OpenID4VCI/pull/525#issuecomment-2981647016 ), we've had input from at least 2 more people, and been discussed in a few WG calls and all comments have been resolved - merging. Feel free to open a new PR/issue if you can see further improvements we could make.

PR #533, opened 2025-05-31
Clarify that number of issued credentials is related to number of keys

### LINKED ISSUE #510: Batch size configuration per credential type
Its possible for the issuer to configure his max batch size that he supports on the credential endpoint.
This configuration is configured for the whole issuer. What happens if i as an issuer would like to issue two different credential types (e.g. gym card and concert ticket) and i only want to batch issue one of them? (or i want different batch sizes)

From my point of view its currently specified that the issuer sets it for everything. So if i support batch issuing for the gym card i should also support it for the concert ticket.

Would it make sense to configure the batch size for every credential configuration? (same as for e.g. cryptographic_binding_methods_supported)

jogu:
closes #510


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1096,7 +1096,7 @@ If the Credential Response is not encrypted, the media type of the response MUST
 
 The following parameters are used in the JSON-encoded Credential Response body:
 
-* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. This specification defines the following parameters to be used inside this object:
+* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. The number of elements in the array is the number of keys for Credentials to be bound to that the Wallet has provided via the `proofs` parameter in the Credential Request, unless the Issuer decides to issue fewer Credentials. This specification defines the following parameters to be used inside this object:
```
paulbastian: It reads a little weird to me. How about:
```suggestion
* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. The number of elements in the `credentials` array matches the number of keys that the Wallet has provided in the `proofs` parameter of the Credential Request, unless the Issuer decides to issue fewer Credentials. This specification defines the following parameters to be used inside this object:
```

jogu: Thanks Paul! I think that is better, though I think "via" is better than "in". "proofs" is an extension point so people could define mechanisms where keys are provided but the keys technically aren't "in" the proofs parameter (a set of keys being derived from an initial key being one case):

```suggestion
* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. The number of elements in the `credentials` array matches the number of keys that the Wallet has provided via the `proofs` parameter of the Credential Request, unless the Issuer decides to issue fewer Credentials. Each key provided by the Wallet is used to bind, at most, one Credential. This specification defines the following parameters to be used inside this object:
```

```suggestion
* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. The number of elements in this array corresponds to the number of keys the Wallet has provided via the `proofs` parameter in the Credential Request, unless the Issuer decides to issue fewer Credentials. Where each credential returned is bound to a unique key that has been provided by the wallet in the credential request. This specification defines the following parameters to be used inside this object:
```

Another possible suggestion for the first part and for the last sentence I added I was trying to weave in explicitly that its not just the number of credentials returned will match the number of proofs provided but also that each credential returned is bound to a unique key. Otherwise plausibly an issuer implementation could for example if it receives 10 proofs in a request return 10 credentials bound to the first key. I'm not sure my language is sufficient yet, but I think we need to clarify that.

jogu: Thanks. I tried adding "Each key provided by the Wallet is used to bind, at most, one Credential." above. Let's see what people think of that...

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1096,7 +1096,7 @@ If the Credential Response is not encrypted, the media type of the response MUST
 
 The following parameters are used in the JSON-encoded Credential Response body:
 
-* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. This specification defines the following parameters to be used inside this object:
+* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. The number of elements in the `credentials` array matches the number of keys that the Wallet has provided via the `proofs` parameter of the Credential Request, unless the Issuer decides to issue fewer Credentials. Each key provided by the Wallet is used to bind, at most, one Credential. This specification defines the following parameters to be used inside this object:
```
```suggestion
* `credentials`: OPTIONAL. Contains an array of one or more issued Credentials. It MUST NOT be used if the `transaction_id` parameter is present. The elements of the array MUST be objects. The number of elements in the `credentials` array matches the number of keys that the Wallet has provided via the `proofs` parameter of the Credential Request, unless the Issuer decides to issue fewer Credentials. Each key provided by the Wallet is used to bind to, at most, one Credential. This specification defines the following parameters to be used inside this object:
```

tplooker (approved):
LGTM with one minor suggestion

sloops77 (approved):

PR #534, opened 2025-05-31
Attempt to make OAuth2 security recommendations more actionable

### LINKED ISSUE #291: Improve content of security recommendations
The draft in Sections 5 and 6 refers to the security BCP:

> The Authorization Endpoint is used in the same manner as defined in [[RFC6749](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#RFC6749)], taking into account the recommendations given in [[I-D.ietf-oauth-security-topics](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#I-D.ietf-oauth-security-topics)].

> The Token Endpoint issues an Access Token and, optionally, a Refresh Token in exchange for the Authorization Code that Client obtained in a successful Authorization Response. It is used in the same manner as defined in [[RFC6749](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#RFC6749)] and follows the recommendations given in [[I-D.ietf-oauth-security-topics](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#I-D.ietf-oauth-security-topics)].

This wording is misleading, as not all protections mentioned in the BCP are mentioned here (e.g., PKCE is only implied in Section 6; the PKCE downgrade attack mitigation is not mentioned).

This may lead to implementers not implementing necessary security mechanisms.

jogu:
Daniel Fett pointed out that the current wording around the Best Current Practice for OAuth 2.0 Security document is misleading as VCI does not mention all the best current practices.

Reword to try and address this.

Add a recommendation to use FAPI2 as a way to comply with the BCP, and add some notes around where FAPI2 might not be applicable.

Also replace the references to RFC9700 with BCP240 as it is likely to be an evolving document.

closes #291

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -451,7 +451,7 @@ The Wallet does not create a response. UX control stays with the Wallet after co
 
 # Authorization Endpoint {#authorization-endpoint}
 
-The Authorization Endpoint is used in the same manner as defined in [@!RFC6749], taking into account the recommendations given in [@!RFC9700].
+The Authorization Endpoint is used in the same manner as defined in [@!RFC6749]. Implementers SHOULD follow the best current practices for OAuth 2.0 Security given in [@!BCP240].
 
 When the grant type `authorization_code` is used, it is RECOMMENDED to use PKCE [@!RFC7636] and Pushed Authorization Requests [@RFC9126]. PKCE prevents authorization code interception attacks. Pushed Authorization Requests ensure the integrity and authenticity of the authorization request.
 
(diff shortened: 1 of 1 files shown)


tplooker (approved):
LGTM, good improvement to just reference the BCP here more generally.

Sakurann (approved):
lgtm. 

just curious - it is best practice to reference as bcp240 and not as rfc9700 even if these are technically the same thing?

jogu:
> lgtm.
> 
> just curious - it is best practice to reference as bcp240 and not as rfc9700 even if these are technically the same thing?

Referencing rfc9700 kinda sets the expectation as "implement exactly what's in this RFC" whereas referencing BCP240 is a more explicit "follow OAuth security BCPs even if they're been updated since rfc9700".

PR #536, opened 2025-05-31
Make text about using credential_identifiers with scopes self-consistent

### LINKED ISSUE #514: Using credential_identifier with scopes seems ambiguous
https://github.com/openid/OpenID4VCI/pull/392 added an option for the AS to return credential_identifier when credentials are requested using scopes.

It resulted in text like this:

Image

that seems to say the there's a `credential_identifiers` in the top level of the token response in this case. And we define one in the IANA considerations.

But we have this text:

> REQUIRED when an Authorization Details of type openid_credential was returned from the Token Response. It MUST NOT be used otherwise.

Image

That only talks about `authorization_details`.

I think there's two possible interpretations of the current text, if scopes are used and the authorization server wants to return a `credential_identifiers`:

1. it needs to return an`authorization_details`
2. it returns a new `credential_identifiers` at the top level of the token response

I think "2" was what was intended. If it was, we should fix the text above to allow `credential_identifier` to be allowed in the credential request, and we should probably add an example token response for this case to make it clear, and change the 'token response' section to clearly define this new top level token response parameter.

If "1" was what was intended, we need to be more explicit that the AS needs to return an `authorization_details` and remove the IANA registration for a `credential_identifiers` parameter at the top level of the token endpoint response.

jogu:
The text sometimes refers to a top level credential_identifiers in the token response, and sometimes refers to it being within authorization_details.

As Stefan Charsley pointed out, there was a clear intent for it to be within authorization_details stated here:

https://github.com/openid/OpenID4VCI/pull/392#issuecomment-2478638816

so fix the text that implies it is at the top level of the token response.

closes #514

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -195,13 +195,13 @@ Below is the summary of how Credential(s) that are being issued are identified t
 - When the Wallet uses Authorization Details in the Authorization Request, the Wallet uses
   either `credential_configuration_id` parameters or `format` and other Credential Format
   specific parameters to identify the requested Credential Configurations. In this case,
-  the Authorization Server MUST return `credential_identifiers` parameter in the Token Response,
+  the Authorization Server MUST return an `authorization_details` parameter containing the `credential_identifiers` parameter in the Token Response,
   and the Wallet uses those `credential_identifier` values in the Credential Request.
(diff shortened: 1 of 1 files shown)


paulbastian (approved):
Approved. I also think this introduction section should not contain VERBS in the first place

charsleysa (approved):

c2bo (approved):

Issue #507, opened 2025-05-16
Integrity of the JWK used for Credential Response encryption


GarethCOliver:
Currently, there is nothing that ensures the integrity of the JWK used for Credential Response encryption. When not using credential request encryption, a party in the middle that can mutate the request can easily replace the jwk without the Client or the Issuer being aware. Credential Request encryption helps, but in principle encryption should not be used for integrity. 

Conceptually, there are two solutions here: 

1) provide a proof over the key (a key_attestation or other proof mechanism as used for other credential keys could work.)

2) provide an optional signature over the request using an already known key (dpop key would be an example)

tplooker:
I think option 2 is a bit more generalised here given the JWK used for response encryption isn't the only aspect of the request that benefits from integrity guarantees. In effect it would mean adding support for signed credential requests, which depending on the shape that #505 takes it could be relatively easy to achieve as we could leverage JWT's to provided either (or both) signed and encrypted requests. .

tplooker:
Beyond the credential response encryption key the public key supplied to issue the credential could also be substituted if one were using the JWT proof type and hence would benefit from being able to be integrity protected in the credential request in certain situations.

Sakurann:
WG discussion:
- in most cases, TLS is what guarantees integrity of the JWK used for Credential Response encryption
- option 1: address this in 1.1. but if we do that we want to remove all credential request & response encryptions in 1.0
- option 2: fix it in 1.0. (seems to be preferred also because this is an ISO requirement)
  - in VCI, define how to sign the credential request and in HAIP mandate it to be DPoP key (does DPoP key being HW bound matter? key integrity and strength of encryption are related)
  - bind DPoP or wallet attestation to the encryption key attestation here
- another option might be mandating encrypting credential request when credential response encryption.

jogu:
I think my worry about the "wallet attestation" and "dpop" type ideas is that we're just moving the trust around, without being clear what we do and don't trust in the first place.

So for example if we say "a wallet attestation is sufficient for the encryption key", I believe that we're saying (based on wallet attestation as per https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#appendix-E ) is that we trust the wallet backend to provide a correct attestation for the key, but we don't trust it to not alter the request from the wallet.

[...]

PR #540, opened 2025-06-05
Integrity of the JWK used for Credential Response encryption - Option 2

c2bo:
Closes https://github.com/openid/OpenID4VCI/issues/507

This is the option creating a JWT for credential request with either the dpop or wallet attestation keys.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -806,15 +806,15 @@ For Cryptographic Key Binding, the Client has the following options defined in (
 
 ## Credential Request {#credential-request}
 
-A Client makes a Credential Request to the Credential Endpoint by sending the following parameters in the entity-body of an HTTP POST request using the `application/json` media type.
+A Client makes a Credential Request to the Credential Endpoint by sending an HTTP POST request using the `application/json` or `application/credential-request+jwt` media types. See (#credential-request-signed) for details about the payload using `application/credential-request+jwt`. In the case of `application/json`, the request contains the following parameters in the entity-body:
 
 * `credential_identifier`: REQUIRED when an Authorization Details of type `openid_credential` was returned from the Token Response. It MUST NOT be used otherwise. A string that identifies a Credential Dataset that is requested for issuance. When this parameter is used, the `credential_configuration_id` MUST NOT be present.
 * `credential_configuration_id`: REQUIRED if a `credential_identifiers` parameter was not returned from the Token Response as part of the `authorization_details` parameter. It MUST NOT be used otherwise. String that uniquely identifies one of the keys in the name/value pairs stored in the `credential_configurations_supported` Credential Issuer metadata. The corresponding object in the `credential_configurations_supported` map MUST contain one of the value(s) used in the `scope` parameter in the Authorization Request. When this parameter is used, the `credential_identifier` MUST NOT be present.
 * `proofs`: OPTIONAL. Object providing one or more proof of possessions of the cryptographic key material to which the issued Credential instances will be bound to. The `proofs` parameter contains exactly one parameter named as the proof type in (#proof-types), the value set for this parameter is a non-empty array containing parameters as defined by the corresponding proof type.
 * `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If this request element is not present, the corresponding credential response returned is not encrypted.
-    * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
-    * `alg`: REQUIRED. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses.
-    * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
+  * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
+  * `alg`: REQUIRED. JWE [@!RFC7516] `alg` algorithm [@!RFC7518] for encrypting Credential Responses.
+  * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
 
 See (#identifying_credential) for the summary of the options how requested Credential(s) are identified throughout the Issuance flow.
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -909,6 +909,45 @@ The Client MAY request encrypted responses by providing its encryption parameter
...
+
+The JWT is signed by the Wallet and contains the following elements:
+
+* in the JOSE header,
+  * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry [@IANA.JOSE]. It MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
+  * `typ`: REQUIRED. MUST be `credential-request+jwt`, which explicitly types the Credential Request JWT as recommended in Section 3.11 of [@!RFC8725].
+  * `kid`: OPTIONAL. MUST be `dpop`, or `wallet-attestation`. Identifies the key the Wallet used to sign the Credential Request. If the `kid` is `dpop`, the Credential Request JWT is signed with the DPoP key of the refresh or access token. If the `kid` is `wallet-attestation`, the key that is used to sign the proof of possession for the Wallet Attestation as described in (#walletattestation) is used to sign the Credential Request JWT. If not present, the key used to sign the Credential Request JWT defaults to the DPoP key.
```
GarethCOliver: Is there any particular reason to specify the value vs just say it must be the kid of the key used for dpop or wallet-attestation? (seems slightly odd to force the value, the client and the verifier should be aware of that key and its kid, right?).

GarethCOliver: any reason not to just make the kid mandatory?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -909,6 +909,45 @@ The Client MAY request encrypted responses by providing its encryption parameter
...
+
+* in the JWT body,
+  * `iat`: REQUIRED (number). Integer for the time at which the key attestation was issued using the syntax defined in [@!RFC7519].
+  * `credential_request`: An Object containing the Credential Request parameters:
```
GarethCOliver: it'd be nice if we just defined this once and used it everywhere.

GarethCOliver:
seems reasonable to me

openid-4-verifiable-credential-issuance-1_0.md
c2bo: Good catch, it should probably be REQUIRED

c2bo: Right now `kid` is not required for key attestation and does not exist for dpop since we only have the `cnf` claim to convey the key. My idea with these static identifiers was to allow an easy way to convey which key was used, but I agree it doesn't feel very clean to me. Any better ideas?

openid-4-verifiable-credential-issuance-1_0.md
c2bo: I wasn't sure if just linking to 8.2 as the content of this object would be sufficient Happy to change that - it would also make everything a lot easier to manage & read.

openid-4-verifiable-credential-issuance-1_0.md
nklomp: What is the reason to be using the `kid` to make the distinction?
A `kid` value is already present on a JWK in a lot of cases. kid's then are being used for instance when DIDs are involved, pointing to an actual identifier of the key (VM relationship). This means that now any Key Management System needs to do identifier juggling.

Is there a specific reason why another header cannot be used?

c2bo: We could use something else. We could also decide to tie this mechanism to DPoP since the RS should know the DPoP key anyway and remove any identifier - that might be the cleanest way if people are fine with limiting this to the combination with DPoP

c2bo:
Updated to only allow DPoP key and remove `kid` + reference the parameter section instead of re-defining. This makes the PR a lot smaller and cleaner imho

Sakurann:
WG discussion:
direction after the discussion:
- direction seems to be that integrity protection of the credential response encryption key cannot be achieved in a meaningful way as long as trust relationship between issuer and the wallet frontend is established through the wallet backend. at least without relying on OS specific mechanisms.
- to protect the credential response encryption key from non-backend attackers, credential request encryption is sufficient, if credential request encryption is mandatory with credential response encryption?

rest of the discussion:
- why mobile OS provided key can't authenticate the frontend to the issuer? it's not the key attestation that proves frontend's integrity
- what requirements we have regarding the key that response is encrypted to?
    - that key needs to be tied to the wallet that controls the access token. can't be an arbitrary key. all key/wallet attestations sent in the credential request are generated by the wallet backend.
- we are facilitating architecture that involves wallet backend - how do existing implementations cope with the wallet backend being in the middle?
    - these implementations use android OS key attestation directly. works because wallet backend is a different entity from the android OS PKI? 
- what authenticates the app is wallet attestation, not key attestation
- Lee had a suggestion to sign over the encryption key in the credential response. but that, malicious backend can spoof.

tplooker:
Thanks for the notes @Sakurann, few comments inline.

> direction seems to be that integrity protection of the credential response encryption key cannot be achieved in a meaningful way as long as trust relationship between issuer and the wallet frontend is established through the wallet backend. at least without relying on OS specific mechanisms.

If we supported signed credential requests, where the signing key was the DPoP key I think this would allow for an architecture that doesn't require a wallet backend, if that is what this is intended to mean?

I do however think supporting models that allow for using OS specific mechanisms like key attestations is important too.

> to protect the credential response encryption key from non-backend attackers, credential request encryption is sufficient, if credential request encryption is mandatory with credential response encryption?

Not sure I follow this, encryption approaches that involve a static key pair (the credential issuers) and an ephemeral key pair (the wallets) are not really immune from MITM attacks where an attacker can just generate a new ephemeral encryption key and encrypt a new request to the credential issuer. The issue essentially is, how to establish trust in who encrypted the request.

jogu:
> Thanks for the notes @Sakurann, few comments inline.
> 
> > direction seems to be that integrity protection of the credential response encryption key cannot be achieved in a meaningful way as long as trust relationship between issuer and the wallet frontend is established through the wallet backend. at least without relying on OS specific mechanisms.
> 
> If we supported signed credential requests, where the signing key was the DPoP key I think this would allow for an architecture that doesn't require a wallet backend, if that is what this is intended to mean? Which is what this PR does?

The goal of this work was, as I understand it, to prevent a malicious wallet backend from being able to access credentials when encrypted credential requests & responses are being passed from the wallet app to the wallet backend and then to the issuer. The discussion in the WG reached the conclusion that there is no realistic way to provide this guarantee. It is pretty much inherent in the current model that the wallet backend (which is pretty much required for any architecture that uses OAuth 2.0 Attestation-Based Client Authentication) has to be trusted.

The discussion further evolved onto stating that the purpose of request/response encryption was to avoid the wallet backend unnecessarily seeing PII.

> I do however think supporting models that allow for using OS specific mechanisms like key attestations is important too.

Yes, but these don't help solve the problem (an attestation that a key is in a hardware backed store doesn't tell you whether it's on the intended device or a device the attacker is using).

> 
> > to protect the credential response encryption key from non-backend attackers, credential request encryption is sufficient, if credential request encryption is mandatory with credential response encryption?
> 
> Not sure I follow this, encryption approaches that involve a static key pair (the credential issuers) and an ephemeral key pair (the wallets) are not really immune from MITM attacks where an attacker can just generate a new ephemeral encryption key and encrypt a new request to the credential issuer. The issue essentially is, how to establish trust in who encrypted the request.

I think that is exactly the issue - we couldn't find a way to establish that trust.

jogu:
As per Kristina's comment above, and further discussion on today's WG call - closing for now. We can revisit in the future. (And thanks to Christian for putting together these PRs so we had some clear options to discuss the pros & cons of - I think that really helped us reach a conclusion).

PR #550, opened 2025-06-18
re-adding missing file examples/authorization_details_sd_jwt_vc.json

paulbastian:


examples/authorization_details_sd_jwt_vc.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "type": "openid_credential",
+    "format": "dc+sd-jwt",
+    "vct": "SD_JWT_VC_example_in_OpenID4VCI"
```
jogu: authorization_details can only use the credential_configuration_id form now I believe, so I think this would make sense and is consistent with the examples for other formats:

```suggestion
    "credential_configuration_id": "PID_SD_JWT_VC",
    "claims": [
      {"path": ["given_name"]},
      {"path": ["family_name"]}
    ]
```

paulbastian: that explains why I originally removed the file :D

tplooker (approved):

PR #552, opened 2025-06-19
Clarify meaning of absence of some credential metadata

### LINKED ISSUE #530: Semantics of missing optional metadata fields is unclear
It is not clear how should missing optional fields in `credential_configurations_supported` map should be interpreted by the wallet. E.g. if `cryptographic_binding_methods_supported` is missing, does it indicate a non-key-bound credential or that the issuing authority did not bother to indicate this? From the wallet perspective, we do not want to take our users through a (potentially lengthy) issuance process just to get a credential that cannot be used at the end. This information is needed upfront, the alternative is poor usability.

Proposal is to either define default values for the missing fields or (preferably) require them. This does not seem like a big burden on the issuance server.

jogu:
Make some items required when key bound credentials are in use, so the wallets knows if it has to supply proofs or not, and if so in what format.

closes #530


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1447,9 +1447,9 @@ This specification defines the following Credential Issuer Metadata parameters:
...
+  * `cryptographic_binding_methods_supported`: OPTIONAL. A non-empty array of case sensitive strings that identify the representation of the cryptographic key material that the issued Credential is bound to, as defined in (#credential-binding). Support for keys in JWK format [@!RFC7517] is indicated by the value `jwk`. Support for keys expressed as a COSE Key object [@!RFC8152] (for example, used in [@!ISO.18013-5]) is indicated by the value `cose_key`. If key binding is required for this credential, both `cryptographic_binding_methods_supported` and `cryptographic_binding_methods_supported` MUST be present. If `cryptographic_binding_methods_supported` is absent, key binding is not required for this credential. When the Cryptographic Key Binding method is a DID, valid values are a `did:` prefix followed by a method-name using a syntax as defined in Section 3.1 of [@!DID-Core], but without a `:`and method-specific-id. For example, support for the DID method with a method-name "example" would be represented by `did:example`.
```
charsleysa:
```suggestion
  * `cryptographic_binding_methods_supported`: OPTIONAL. A non-empty array of case sensitive strings that identify the representation of the cryptographic key material that the issued Credential is bound to, as defined in (#credential-binding). Support for keys in JWK format [@!RFC7517] is indicated by the value `jwk`. Support for keys expressed as a COSE Key object [@!RFC8152] (for example, used in [@!ISO.18013-5]) is indicated by the value `cose_key`. If key binding is required for this credential, both `cryptographic_binding_methods_supported` and `proof_types_supported ` MUST be present. If `cryptographic_binding_methods_supported` is absent, key binding is not required for this credential. When the Cryptographic Key Binding method is a DID, valid values are a `did:` prefix followed by a method-name using a syntax as defined in Section 3.1 of [@!DID-Core], but without a `:`and method-specific-id. For example, support for the DID method with a method-name "example" would be represented by `did:example`.
```

jogu: Ooops. Fixed, thanks!

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1453,9 +1453,9 @@ This specification defines the following Credential Issuer Metadata parameters:
...
+  * `proof_types_supported`: OPTIONAL. Object that describes specifics of the key proof(s) that the Credential Issuer supports. `proof_types_supported` MUST be present if `cryptographic_binding_methods_supported` is present, and MUST be omitted otherwise. If absent, the Wallet is not required to supply proofs when requesting this credential. This object contains a list of name/value pairs, where each name is a unique identifier of the supported proof type(s). Valid values are defined in (#proof-types), other values MAY be used. The Wallet also uses this identifier in the Credential Request as defined in (#credential-request). The value in the name/value pair is an object that contains metadata about the key proof and contains the following parameters defined by this specification:
```
```suggestion
  * `proof_types_supported`: OPTIONAL. Object that describes specifics of the key proof(s) that the Credential Issuer supports. It MUST be present if `cryptographic_binding_methods_supported` is present, and omitted otherwise. If absent, the Wallet is not required to supply proofs when requesting this credential. This object contains a list of name/value pairs, where each name is a unique identifier of the supported proof type(s). Valid values are defined in (#proof-types), other values MAY be used. The Wallet also uses this identifier in the Credential Request as defined in (#credential-request). The value in the name/value pair is an object that contains metadata about the key proof and contains the following parameters defined by this specification:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1453,9 +1453,9 @@ This specification defines the following Credential Issuer Metadata parameters:
...
+  * `cryptographic_binding_methods_supported`: OPTIONAL. A non-empty array of case sensitive strings that identify the representation of the cryptographic key material that the issued Credential is bound to, as defined in (#credential-binding). Support for keys in JWK format [@!RFC7517] is indicated by the value `jwk`. Support for keys expressed as a COSE Key object [@!RFC8152] (for example, used in [@!ISO.18013-5]) is indicated by the value `cose_key`. If key binding is required for this credential, both `cryptographic_binding_methods_supported` and `proof_types_supported` MUST be present. If `cryptographic_binding_methods_supported` is absent, key binding is not required for this credential. When the Cryptographic Key Binding method is a DID, valid values are a `did:` prefix followed by a method-name using a syntax as defined in Section 3.1 of [@!DID-Core], but without a `:`and method-specific-id. For example, support for the DID method with a method-name "example" would be represented by `did:example`.
```
```suggestion
  * `cryptographic_binding_methods_supported`: OPTIONAL. A non-empty array of case sensitive strings that identify the representation of the cryptographic key material that the issued Credential is bound to, as defined in (#credential-binding). It MUST be present when key binding is required for a credential, and omitted otherwise. If absent, key binding is not required for this credential. Support for keys in JWK format [@!RFC7517] is indicated by the value `jwk`. Support for keys expressed as a COSE Key object [@!RFC8152] (for example, used in [@!ISO.18013-5]) is indicated by the value `cose_key`. When the Cryptographic Key Binding method is a DID, valid values are a `did:` prefix followed by a method-name using a syntax as defined in Section 3.1 of [@!DID-Core], but without a `:`and method-specific-id. For example, support for the DID method with a method-name "example" would be represented by `did:example`.
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1453,9 +1453,9 @@ This specification defines the following Credential Issuer Metadata parameters:
...
   * `credential_signing_alg_values_supported`: OPTIONAL. A non-empty array of algorithm identifiers that identify the algorithms that the Issuer uses to sign the issued Credential. Algorithm identifier types and values used are determined by the Credential Format and are defined in (#format-profiles).
```
```suggestion
  * `credential_signing_alg_values_supported`: OPTIONAL. A non-empty array of algorithm identifiers that identify the algorithms that the Issuer uses to sign the issued Credential. Algorithm identifier types and values used are determined by the Credential Format and are defined in (#format-profiles).
  * `cryptographic_binding_methods_supported`: OPTIONAL. A non-empty array of case sensitive strings that identify the representation of the cryptographic key material that the issued Credential is bound to, as defined in (#credential-binding). It MUST be present when key binding is required for a credential, and omitted otherwise. If absent, key binding is not required for this credential. Support for keys in JWK format [@!RFC7517] is indicated by the value `jwk`. Support for keys expressed as a COSE Key object [@!RFC8152] (for example, used in [@!ISO.18013-5]) is indicated by the value `cose_key`. When the Cryptographic Key Binding method is a DID, valid values are a `did:` prefix followed by a method-name using a syntax as defined in Section 3.1 of [@!DID-Core], but without a `:`and method-specific-id. For example, support for the DID method with a method-name "example" would be represented by `did:example`.
```

Sakurann (approved):

PR #556, opened 2025-06-26
Editorial: authorization details in token requests/responses

sloops77:
Suggested edits to clarify when token responses should include `authorization_details`

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -659,7 +659,7 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
     credential_configuration_id%22%3A%20%22UniversityDegreeCredential%22%7D%5D
 ```
 
-The Wallet SHOULD NOT send the `authorization_details` parameter in the Token Request if the parameter has been previously sent in the Authorization Request(#credential-authz-request)
+The Wallet may send the `authorization_details` parameter in the Token Request even when the parameter has been previously sent in the [Authorization Request](#credential-authz-request) as described in Section 6.1 of [@!RFC9396]. It allows the AS to make a decision based on whether the Wallet is asking for more or less access than the previous request. With respect to `authorization_details` items using the `credential_configuration_id` introduced in this specification, it is RECOMMENDED that a AS would accept a request from the Wallet containiing a subset of `credential_configuration_id` parameters received in the original [Authorization Request](#credential-authz-request) and issue a token for the reduced set.
 
 ## Successful Token Response {#token-response}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -659,7 +659,7 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
     credential_configuration_id%22%3A%20%22UniversityDegreeCredential%22%7D%5D
 ```
 
-The Wallet SHOULD not send the `authorization_details` parameter in the Token Request if the parameter has been previously sent in the Authorization Request(#credential-authz-request)
+The Wallet SHOULD NOT send the `authorization_details` parameter in the Token Request if the parameter has been previously sent in the Authorization Request(#credential-authz-request)
 
 ## Successful Token Response {#token-response}
 
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -659,7 +659,7 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
     credential_configuration_id%22%3A%20%22UniversityDegreeCredential%22%7D%5D
 ```
 
-The Wallet may send the `authorization_details` parameter in the Token Request even when the parameter has been previously sent in the [Authorization Request](#credential-authz-request) as described in Section 6.1 of [@!RFC9396]. It allows the AS to make a decision based on whether the Wallet is asking for more or less access than the previous request. With respect to `authorization_details` items using the `credential_configuration_id` introduced in this specification, it is RECOMMENDED that a AS would accept a request from the Wallet containiing a subset of `credential_configuration_id` parameters received in the original [Authorization Request](#credential-authz-request) and issue a token for the reduced set.
+The Wallet may send the `authorization_details` parameter in the Token Request even when the parameter has been previously sent in the [Authorization Request](#credential-authz-request) as described in Section 6.1 of [@!RFC9396]. It allows the AS to make a decision based on whether the Wallet is asking for more or less access than the previous request. With respect to `authorization_details` items using the `credential_configuration_id` introduced in this specification, it is RECOMMENDED that the AS would accept a request from the Wallet containing a subset of `credential_configuration_id` parameters received in the original [Authorization Request](#credential-authz-request) and issue a token for the reduced set.
 
 ## Successful Token Response {#token-response}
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -659,6 +659,8 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
     credential_configuration_id%22%3A%20%22UniversityDegreeCredential%22%7D%5D
 ```
 
+The Wallet SHOULD not send in the Token Request `authorization_details` if it has been previosuly sent in the Authorization Request(#credential-authz-request)
```
```suggestion
The Wallet SHOULD not send the `authorization_details` parameter in the Token Request if the parameter has been previously sent in the Authorization Request(#credential-authz-request)
```
```suggestion
The Wallet SHOULD not send in the Token Request `authorization_details` if it has been previosuly sent in the Authorization Request(#credential-authz-request)
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -667,8 +669,8 @@ The Authorization Server might decide to authorize issuance of multiple instance
...
+* `authorization_details`: REQUIRED when the `authorization_details` parameter is used in either the [Authorization Request](#credential-authz-request) or [Token Request](#token-request) as defined in (#authorization-details). OPTIONAL when `scope` parameter was used to request issuance of a Credential of a certain Credential Configuration. It is a non-empty array of objects, as defined in Section 7 of [@!RFC9396]. In addition to the parameters defined in (#authorization-details), this specification defines the following parameter to be used with the authorization details type `openid_credential` in the Token Response:
```
```suggestion
* `authorization_details`: REQUIRED when the `authorization_details` parameter, as defined in (#authorization-details), is used in either the [Authorization Request](#credential-authz-request) or [Token Request](#token-request). OPTIONAL when `scope` parameter was used to request issuance of a Credential of a certain Credential Configuration. It is a non-empty array of objects, as defined in Section 7 of [@!RFC9396]. In addition to the parameters defined in (#authorization-details), this specification defines the following parameter to be used with the authorization details type `openid_credential` in the Token Response:
```
```suggestion
* `authorization_details`: REQUIRED when the `authorization_details` parameter is used in either the [Authorization Request](#credential-authz-request) or [Token Request](#token-request) as defined in (#authorization-details). OPTIONAL when `scope` parameter was used to request issuance of a Credential of a certain Credential Configuration. It is a non-empty array of objects, as defined in Section 7 of [@!RFC9396]. In addition to the parameters defined in (#authorization-details), this specification defines the following parameter to be used with the authorization details type `openid_credential` in the Token Response:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -659,6 +659,8 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
     credential_configuration_id%22%3A%20%22UniversityDegreeCredential%22%7D%5D
 ```
 
+The Wallet SHOULD not send the `authorization_details` parameter in the Token Request if the parameter has been previously sent in the Authorization Request(#credential-authz-request)
```
dastoikov: Suggestion: SHOULD not -> SHOULD NOT

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -667,8 +669,8 @@ The Authorization Server might decide to authorize issuance of multiple instance
...
+* `authorization_details`: REQUIRED when the `authorization_details` parameter, as defined in (#authorization-details), is used in either the [Authorization Request](#credential-authz-request) or [Token Request](#token-request). OPTIONAL when `scope` parameter was used to request issuance of a Credential of a certain Credential Configuration. It is a non-empty array of objects, as defined in Section 7 of [@!RFC9396]. In addition to the parameters defined in (#authorization-details), this specification defines the following parameter to be used with the authorization details type `openid_credential` in the Token Response:
+  * `credential_identifiers`: REQUIRED. A non-empty array of strings, each uniquely identifying a Credential Dataset that can be issued using the Access Token returned in this response. Each of these Credential Datasets corresponds to the Credential Configuration referenced in the `credential_configuration_id` parameter. The Wallet MUST use these identifiers together with an Access Token in subsequent Credential Requests. See (#identifying_credential) for the summary of the options how requested Credential(s) are identified throughout the Issuance flow.
```
jogu, line 673: I'm not sure either the existing sentence or the new sentence make sense in the case where there was only a scope in the request and the AS has chosen to return `authorization_details`.

sloops77: @jogu My reading of the existing spec is that this updated sentence does make sense:
1. If `authorization_details` was sent in a request, it contained a `credential_configuration_id`
2. If `scope` was sent then it's optional for the token response to contain `authorization_details`

In either case that `authorization_details` is in the token response, the `credential_configuration_id` is a REQUIRED parameter. See https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-using-authorization-details

jogu: Thanks @sloops77 - I've read it again and actually I think your change it good. Thanks for the explanation!

openid-4-verifiable-credential-issuance-1_0.md
jogu: This change was applied so marking this as resolved - thanks!

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -659,6 +659,8 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
     credential_configuration_id%22%3A%20%22UniversityDegreeCredential%22%7D%5D
 ```
 
+The Wallet SHOULD NOT send the `authorization_details` parameter in the Token Request if the parameter has been previously sent in the Authorization Request(#credential-authz-request)
```
jogu: I'm not sure if this is too strong. There are potential cases where you might want to do this to obtain an access token that has access to fewer credentials than were authorised. Not sure exactly how is best to express that though.

sloops77: @jogu I think the discussion in [RFC9396](https://www.rfc-editor.org/rfc/rfc9396.html#name-comparing-authorization-det) is relevant here. I think I should rewrite this to defer to that section and make a statement about how the token request can ask for fewer permissions than the previous request. 

Do you think an example is needed - I would say not since unlike in Rich Authz, it's an unlikely scenario

ok?

jogu: sounds good to me

charsleysa (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -659,6 +659,8 @@ grant_type=urn:ietf:params:oauth:grant-type:pre-authorized_code
     credential_configuration_id%22%3A%20%22UniversityDegreeCredential%22%7D%5D
 ```
 
+The Wallet may send the `authorization_details` parameter in the Token Request even when the parameter has been previously sent in the [Authorization Request](#credential-authz-request) as described in Section 6.1 of [@!RFC9396]. It allows the AS to make a decision based on whether the Wallet is asking for more or less access than the previous request. With respect to `authorization_details` items using the `credential_configuration_id` introduced in this specification, it is RECOMMENDED that a AS would accept a request from the Wallet containiing a subset of `credential_configuration_id` parameters received in the original [Authorization Request](#credential-authz-request) and issue a token for the reduced set.
```
dastoikov:
```suggestion
The Wallet may send the `authorization_details` parameter in the Token Request even when the parameter has been previously sent in the [Authorization Request](#credential-authz-request) as described in Section 6.1 of [@!RFC9396]. It allows the AS to make a decision based on whether the Wallet is asking for more or less access than the previous request. With respect to `authorization_details` items using the `credential_configuration_id` introduced in this specification, it is RECOMMENDED that the AS would accept a request from the Wallet containing a subset of `credential_configuration_id` parameters received in the original [Authorization Request](#credential-authz-request) and issue a token for the reduced set.
```

dastoikov (approved):

PR #565, opened 2025-07-07
clarification on signed metadata that parameters are included as top-…

paulbastian:
…level claims in JWS payload


sz3nt:
This makes the related section clear enough for me now, thank you! :)

jogu (approved):

bc-pi (approved):

PR #583, opened 2025-07-22
Merge Deferred Credential Endpoint text into Credential Endpoint

### LINKED ISSUE #523: [Editorial] Refactor Credential and Deferred Credential endpoint text to be less duplicative
Currently Credential Request/Responses are defined separately for Credential and Deferred Credential Endpoints, with the Deferred Credential Request/Response re-using a subset of the Credential Endpoint and adding another field. 

If we instead defined Credential Request and Response separately to the two endpoints, and simply provided the restrictions (i.e. Credential Request MUST contain proofs. Deferrered Credential Request MUST contain transaction_id), we'd remove duplication, prevent issues where parameters are added to one and not the other and make it easier if we wanted to move to having only Credential Endpoint handle both in the future. 

This wouldn't be a technical change, but is a semantic one (i.e. that the request to each endpoint is conceptually the same, and the response from each is conceptually the same). This is most clear on the response, but  less so on the request. 

I do think it'd make reading and understanding deferred issuance a lot easier so a good simplification to make.

GarethCOliver:
Resolves #523

The goal of this is to change the semantics of how requests to and from Credential Endpoint and Deferred Credential Endpoint work. 

Today: Credential Endpoint takes Credential Requests and returns Credential Responses, while Deferred Credential Endpoint takes Deferred Credential Requests (which references some parameters from Credential Request) and return Deferred Credential Responses (which references some parameters from Credential Responses). 

After this change: Credential Endpoint and Deferred Credential Endpoint take Credential Requests and return Credential Responses. 
  - Credential Requests to the Credential Endpoint can additionally include parameters related to proofs + identifying the Credential configuration
  - Credential Requests to the Deferred Credential Endpoint can additionally include `transaction_id`. 
  - Credential Responses are the same from both. 

This results in one normative change: 
- Credential Responses from Deferred Credential Endpoint that continue to defer MUST include `transaction_id`, unchanged from the request. This does not result in a behavioral change, and aligns what is returned from Credential Endpoint and Deferred Credential Endpoint.

The intended benefits of this semantic change are as follows: 
  - The endpoints become more consistent with one another, reducing edge cases when comparing deferred vs non-deferred flows. 
  - Wallets can process responses uniformly, regardless of whether they are returned initially or in subsequent requests
  - Future extensions more easily apply to both (for both responses and requests).

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -812,7 +812,7 @@ For Cryptographic Key Binding, the Client has different options to provide Crypt
 
 A Client makes a Credential Request to the Credential Endpoint by sending parameters in the entity-body of an HTTP POST request. The Credential Request MAY be encrypted (on top of TLS) using the `credential_request_encryption` parameter in (#credential-issuer-metadata) as specified in (#encrypted-messages).
 
-In an initial Credential Request the Client includes the following parameters: 
+In a Credential Request made to the Credential Endpoint the Client includes the following parameters: 
 
 * `credential_identifier`: REQUIRED when an Authorization Details of type `openid_credential` was returned from the Token Response. It MUST NOT be used otherwise. A string that identifies a Credential Dataset that is requested for issuance. When this parameter is used, the `credential_configuration_id` MUST NOT be present.
 * `credential_configuration_id`: REQUIRED if a `credential_identifiers` parameter was not returned from the Token Response as part of the `authorization_details` parameter. It MUST NOT be used otherwise. String that uniquely identifies one of the keys in the name/value pairs stored in the `credential_configurations_supported` Credential Issuer metadata. The corresponding object in the `credential_configurations_supported` map MUST contain one of the value(s) used in the `scope` parameter in the Authorization Request. When this parameter is used, the `credential_identifier` MUST NOT be present.
@@ -827,41 +827,6 @@ The proof(s) in the `proofs` parameter MUST incorporate the Credential Issuer Id
 The `proofs` parameter MUST be present if the `proof_types_supported` parameter is present in the `credential_configurations_supported` parameter of the Issuer metadata for the requested Credential.
 
 The `c_nonce` value is retrieved from the Nonce Endpoint as defined in (#nonce-endpoint).
-
-In subsequent (i.e Deferred) Credential Requests the Client instead includes the following parameter:
-
-* `transaction_id`: REQUIRED. String identifying a Deferred Issuance transaction.
-
-The Credential Issuer MUST invalidate the `transaction_id` after the Credential for which it was meant has been obtained by the Wallet.
-
-The following is a non-normative example of a Deferred Credential Request:
-
-```
-POST /deferred_credential HTTP/1.1 
-Host: server.example.com
-Content-Type: application/json
-Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
-
-{
-  "transaction_id": "8xLOxBtZp8"
-}
-```
-
-All Credential Request(s) include the following parameters:
-
-* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If this request element is not present, the corresponding credential response returned is not encrypted.
-    * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
-    * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
-    * `zip`: OPTIONAL. JWE [@!RFC7516] `zip` algorithm [@!RFC7518] for compressing Credential Responses prior to encryption. If absent then compression MUST not be used.
-
-Additional Credential Request parameters MAY be defined and used.
-The Credential Issuer MUST ignore any unrecognized parameters.
-
-The Credential Issuer indicates support for encrypted requests by including the `credential_request_encryption` parameter in the Credential Issuer Metadata. The Client MAY encrypt the request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`. 
-
-When performing Credential Request encryption, the Client MUST encode the information in the Credential Request in a JWT as specified by (#encrypted-messages), using the parameters from the `credential_request_encryption` object in the Credential Issuer Metadata.
-
-If the Credential Request is not encrypted, the media type of the request MUST be set to `application/json`.
 
 Below is a non-normative example of a Credential Request for a Credential in [@ISO.18013-5] format using the Credential configuration identifier and a key proof type `jwt` (with line breaks within values for display purposes only):
 
@@ -946,6 +911,41 @@ Authorization: BEARER czZCaGRSa3F0MzpnWDFmQmF0M2JW
 }
 ```
 
+If the Credential Issuance was deferred, the Wallet continues by making requests to the Deferred Credential Endpoint. These Credential Requests include the following parameter:
+
+* `transaction_id`: REQUIRED. String identifying a Deferred Issuance transaction.
+
+The Credential Issuer MUST invalidate the `transaction_id` after the Credential for which it was meant has been obtained by the Wallet.
+
+The following is a non-normative example of a Deferred Credential Request:
+
+```
+POST /deferred_credential HTTP/1.1 
+Host: server.example.com
+Content-Type: application/json
+Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
+
+{
+  "transaction_id": "8xLOxBtZp8"
+}
+```
+
+All Credential Request(s) include the following parameters:
+
+* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If this request element is not present, the corresponding credential response returned is not encrypted.
+    * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
+    * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
+    * `zip`: OPTIONAL. JWE [@!RFC7516] `zip` algorithm [@!RFC7518] for compressing Credential Responses prior to encryption. If absent then compression MUST not be used.
+
+Additional Credential Request parameters MAY be defined and used.
+The Credential Issuer MUST ignore any unrecognized parameters.
+
+The Credential Issuer indicates support for encrypted requests by including the `credential_request_encryption` parameter in the Credential Issuer Metadata. The Client MAY encrypt the request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`. 
+
+When performing Credential Request encryption, the Client MUST encode the information in the Credential Request in a JWT as specified by (#encrypted-messages), using the parameters from the `credential_request_encryption` object in the Credential Issuer Metadata.
+
+If the Credential Request is not encrypted, the media type of the request MUST be set to `application/json`.
+
 The Credential Issuer indicates support for encrypted responses by including the `credential_response_encryption` parameter in the Credential Issuer Metadata. The Client MAY request encrypted responses by providing its encryption parameters in the Credential Request when `encryption_required` is `false` and MUST do so when `encryption_required` is `true`. Credential Request encryption MUST be used if the `credential_response_encryption` parameter is included, to prevent it being substituted by an attacker.
 
 ## Credential Response {#credential-response}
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -830,6 +828,32 @@ The `proofs` parameter MUST be present if the `proof_types_supported` parameter
 
 The `c_nonce` value is retrieved from the Nonce Endpoint as defined in (#nonce-endpoint).
 
+In subsequent (i.e Deferred) Credential Requests the Client instead includes the following parameter:
+
+* `transaction_id`: REQUIRED. String identifying a Deferred Issuance transaction.
```
jogu: I don't think we had working group consensus to allow `transaction_id` to be sent to the normal credential endpoint, which seems to be what is now allowed?

I don't really mind whether we do this or not, but if we do allow `transaction_id` to be used at the credential endpoint then I think we have to remove the deferred credential endpoint entirely, as it wouldn't be adding anything useful that I can see and just introduces extra potential protocol flows that have to be implemented/tested.

jogu (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: As it stands, the text does not allow returning it to the Credential Endpoint (due to the text when `transaction_id` is returned in the Credential Response requiring subsequent Credential Requests be made to the Deferred Credential Endpoint. So as it stands it is only editorial, as it just makes clear that Credential Requests/Responses are being made to both endpoints. 

Obviously this also makes it trivial to remove Deferred Credential Endpoint if we so desire if there is consensus by updating that text to allow doing subsequent requests to the Credential Endpoint.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -830,6 +828,32 @@ The `proofs` parameter MUST be present if the `proof_types_supported` parameter
...
+Host: server.example.com
+Content-Type: application/json
+Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
+
+{
+  "transaction_id": "8xLOxBtZp8"
+}
+```
+
+The following parameters may be any Credential Request:
```
```suggestion
All Credential Request(s) include the following parameters:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -939,7 +963,7 @@ The following parameters are used in the JSON-encoded Credential Response body:
...
-* `transaction_id`: OPTIONAL. String identifying a Deferred Issuance transaction. This parameter is contained in the response if the Credential Issuer cannot immediately issue the Credential. The value is subsequently used to obtain the respective Credential with the Deferred Credential Endpoint (see (#deferred-credential-issuance)). It MUST not be used if the `credentials` parameter is present. It MUST be invalidated after the Credential for which it was meant has been obtained by the Wallet.
+* `transaction_id`: OPTIONAL. String identifying a Deferred Issuance transaction. This parameter is contained in the response if the Credential Issuer cannot immediately issue the Credential. The value is subsequently used to obtain the respective Credential with the Deferred Credential Endpoint (see (#deferred-credential-issuance)). It MUST not be used if the `credentials` parameter is present. It MUST be invalidated after the Credential for which it was meant has been obtained by the Wallet. This MAY be omitted if the request contained `transaction_id`, in which case the Client should continue to use the previously obtained `transaction_id`.
```
jogu: I'm not sure I'm keen on introducing behaviour where the Credential Issuer can return a new `transaction_id` midway through a deferred issuance.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1035,114 +1060,7 @@ Cache-Control: no-store
...
-Content-Type: application/json
-Cache-Control: no-store
-
-{
-  "error": "invalid_transaction_id"
-}
-```
-
-In the event the Credential Issuer can no longer issue the credential(s), the `credential_request_denied` error code as defined in (#credential-request-errors) should be used in response to a request. A wallet upon receiving this error SHOULD stop making requests to the deferred credential endpoint for the given `transaction_id`. 
+This endpoint is a specialized Credential Endpoint for handling Deferred Credential Requests. The Deferred Credential Endpoint follows all the requirements in {#credential-endpoint}. All Credential Requests MUST contain `transaction_id`. Support for this endpoint is OPTIONAL.
```
jogu: I think the "specialised Credential Endpoint" is part of what confused me - "specialized" I think can infer that normal credential endpoint can also be used for this case. Not sure what a better word is though.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -830,6 +828,32 @@ The `proofs` parameter MUST be present if the `proof_types_supported` parameter
 
 The `c_nonce` value is retrieved from the Nonce Endpoint as defined in (#nonce-endpoint).
 
+In subsequent (i.e Deferred) Credential Requests the Client instead includes the following parameter:
```
```suggestion
If Credential Issuance was deferred, the Wallet continues by making a request to the Deferred Credential Endpoint containing the following parameter:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -939,7 +963,7 @@ The following parameters are used in the JSON-encoded Credential Response body:
...
+* `transaction_id`: OPTIONAL. String identifying a Deferred Issuance transaction. This parameter is contained in the response if the Credential Issuer cannot immediately issue the Credential. The value is subsequently used to obtain the respective Credential with the Deferred Credential Endpoint (see (#deferred-credential-issuance)). It MUST not be used if the `credentials` parameter is present. It MUST be invalidated after the Credential for which it was meant has been obtained by the Wallet. This MAY be omitted if the request contained `transaction_id`, in which case the Client should continue to use the previously obtained `transaction_id`.
 * `interval`: REQUIRED if `transaction_id` is present. Contains a positive number that represents the minimum amount of time in seconds that the Wallet SHOULD wait after receiving the response before sending a new request to the Deferred Credential Endpoint. It MUST NOT be used if the `credentials` parameter is present.
```
jogu, line 967: We've (I think accidentally) changed `interval` in the deferred credential response to being optional.

openid-4-verifiable-credential-issuance-1_0.md
jogu: Ah okay... I've suggested a couple of wording improvements based on that.

Overall I'm not convinced the current state is an improvement, currently it feels to me like the existing text is clearer, and it seems like we've changed normative behaviour in a few places (allowing interval to be omitted and allowing transaction_id to be changed during a deferred issuance), and if we're going this way we should probably retain an example of a deferred credential response.

GarethCOliver: moved away from this, now there is never a difference in deferred Credential response

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: fair enough, resolved by mandating it be unchanged.

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: tried something better.

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: updated to something equivalent (and the lead in to the previous section to specify Credential Endpoint rather than being more open.

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: resolved by requiring transaction_id

GarethCOliver:
updated description and text to more accurately reflect the goals of this merge and why I believe it to be an improvement semantically over the current text.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -810,15 +810,13 @@ For Cryptographic Key Binding, the Client has different options to provide Crypt
 
 ## Credential Request {#credential-request}
 
-A Client makes a Credential Request to the Credential Endpoint by sending the following parameters in the entity-body of an HTTP POST request. The Credential Request MAY be encrypted (on top of TLS) using the `credential_request_encryption` parameter in (#credential-issuer-metadata) as specified in (#encrypted-messages).
+A Client makes a Credential Request to the Credential Endpoint by sending parameters in the entity-body of an HTTP POST request. The Credential Request MAY be encrypted (on top of TLS) using the `credential_request_encryption` parameter in (#credential-issuer-metadata) as specified in (#encrypted-messages).
```
tlodderstedt:
```suggestion
A Client makes a Credential Request to the Credential Endpoint by sending parameters in the entity-body of an HTTP POST request. The Credential Request MAY be encrypted (on top of TLS) using the encryption settings as defined in the `credential_request_encryption` metadata parameter in (#credential-issuer-metadata) as specified in (#encrypted-messages).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -922,6 +911,41 @@ Authorization: BEARER czZCaGRSa3F0MzpnWDFmQmF0M2JW
...
+Host: server.example.com
+Content-Type: application/json
+Authorization: Bearer czZCaGRSa3F0MzpnWDFmQmF0M2JW
+
+{
+  "transaction_id": "8xLOxBtZp8"
+}
+```
+
+All Credential Request(s) include the following parameters:
```
tlodderstedt:
```suggestion
All Credential Request(s) MAY include the following parameters:
```

tlodderstedt (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -810,15 +810,13 @@ For Cryptographic Key Binding, the Client has different options to provide Crypt
...
 * `proofs`: OPTIONAL. Object providing one or more proof of possessions of the cryptographic key material to which the issued Credential instances will be bound to. The `proofs` parameter contains exactly one parameter named as the proof type in (#proof-types), the value set for this parameter is a non-empty array containing parameters as defined by the corresponding proof type.
-* `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If this request element is not present, the corresponding credential response returned is not encrypted.
-    * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
-    * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
-    * `zip`: OPTIONAL. JWE [@!RFC7516] `zip` algorithm [@!RFC7518] for compressing Credential Responses prior to encryption. If absent then compression MUST not be used.
```
Sakurann, line 821: why is this removed from here? this applies to non-deferred cases, too, right?

GarethCOliver: It's in the section of common parameters. See line 935-938. 

The structure of this section is: 

parameters only for Credential Requests to the Credential Endpoint. 
parameters only for Credential Requests to Deferred Credential Endpoint
parameters common to All Credential Requests.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1035,114 +1060,7 @@ Cache-Control: no-store
...
-Content-Type: application/json
-Cache-Control: no-store
-
-{
-  "error": "invalid_transaction_id"
-}
-```
-
-In the event the Credential Issuer can no longer issue the credential(s), the `credential_request_denied` error code as defined in (#credential-request-errors) should be used in response to a request. A wallet upon receiving this error SHOULD stop making requests to the deferred credential endpoint for the given `transaction_id`. 
+The Deferred Credential Endpoint is used to handle deferred Credential Issuance. The Deferred Credential Endpoint follows the requirements in {#credential-endpoint}. Support for this endpoint is OPTIONAL.
```
charsleysa, line 1063: Is support for the deferred credential endpoint optional for both the wallet and the issuer or only the issuer?

sloops77:
I am not a fan of structure that results from these changes. I think it makes it harder to implement and is quite confusing

The definition of the request parameter used in `Deferred Credential Endpoint`, which is ONLY used within that endpoint, is now defined and the examples shown under the `Credential Endpoint` section. There isnt a markdown link to the relevant section - and it's up to the reader to scroll around and find it towards the bottom of that section.

I also agree with @charsleysa, that it would be a lot cleaner to combine the endpoints completely

openid-4-verifiable-credential-issuance-1_0.md
GarethCOliver: I guess if Wallets don't support it, they can not successfully retrieve credentials that are deferred. I don't know what requirements we make around wallets in VCI, but I'd mandate it for wallets in HAIP.

GarethCOliver:
@charsleysa I personally feel the same way, but making such a change at this time seemed like more than people wanted to take on. If there is support from the WG to make that change it is trivial to do so. What you noted here is basically the motivation for starting this PR. I generally find that in practice having two things that are meant to have the same functionality but are subtly different results in practical inter-op issues. 

I could update the structure even as-is by adding some sections for clarity, if it helps readability: 

## (Common? Base? Shared) Credential Request (Parameters?) 
  - details common parameters and text, notably e2ee

###  Credential Request (Initial Credential Request?)
    - details the proofs etc for the initial credential request
 
### Deferred Credential Request
  - details transaction id

## Credential Response
 - applies to both

Oh, Encrypted Messages should have one less #, it's not meant to be in the Deferred Credential Endpoint but a section of its own (as it applies to both, in the same manner).

awoie (changes requested):
I appreciate the effort behind this PR and the motivation to improve the spec. If it would result in less code and a more maintainable structure, that would be a great outcome. That said, I'm not entirely sure this is necessarily the case here.

First of all, it's not entirely clear whether, with the current proposed changes, there's an unintended side effect where the deferred endpoint could be used to send a regular credential request without a transaction_id, since the text simply states that the endpoint follows the requirements of the regular credential endpoint.

From an editorial perspective, I find some of the structural changes make the document harder to follow:
- I would expect all parameters for the credential request sent to the credential endpoint to be listed together at the start of the credential request section.
- The "common parameters" section feels quite hidden, as it appears only after all the examples.
- It's harder to follow that transaction_id can be provided in the credential request without first introducing the fact that the credential response (in deferred issuance) can contain a transaction_id.
- ...

Personally, I'd be more inclined to merge the credential and deferred credential endpoints, as others have suggested, to keep the flow cohesive and reduce duplication.

openid-4-verifiable-credential-issuance-1_0.md
paulbastian: interval is required is transaction_id is present, what's wrong here?

paulbastian (changes requested):
I have to admit that this reshuffeling of sections seems confusing to me. If we want to simplify this, then we should merge both endpoints together. Also, it is even possible with the current draft to run both endpoints over the same URL.

jogu:
Discussed on today's WG call - given the recent comments it seems we have a consensus not to move forward with this one and to close.

Gareth to open a new PR that unifies the deferred/non-deferred responses.

PR #586, opened 2025-07-24
Fix url for VC_DATA to reference the 1.1 version

lebedenko-ubique:
Currently `VC_DATA` references `https://www.w3.org/TR/vc-data-model` which is now on version `2.0`, as the intention is to reference the `1.1` version, the url should be updated to `https://www.w3.org/TR/2022/REC-vc-data-model-20220303`.


tlodderstedt (approved):

jogu:
@c2bo 

> Should we pin the other w3c spec links also to stable urls (e.g., data integrity)?

Good point, I opened an issue for that and marked it ready for PR.

PR #589, opened 2025-07-31
Adds the ability to continue after redirect_to_web interaction on IAE

tplooker:
Spun out of #509, this PR adds the ability to continue an authorisation process at the IAE after a redirect_to_web interaction.

Following feedback on the WG call i've added the requirement for PKCE in the event it was used in the initial request.

### DIFF
--- a/1.1/openid-4-verifiable-credential-issuance-1_1.md
+++ b/1.1/openid-4-verifiable-credential-issuance-1_1.md
@@ -770,7 +770,7 @@ The following non-normative example shows a payload of a signed request object:
 
 ### Follow-up Request {#follow-up-request}
 
-Follow-up requests to the Interactive Authorization Endpoint only MUST include the `auth_session` value received most recently from the Authorization Server (see (#iar-interaction-required-response)).
+Follow-up requests to the Interactive Authorization Endpoint MUST include the `auth_session` value received most recently from the Authorization Server (see (#iar-interaction-required-response)).
 
 Besides `auth_session`, follow-up requests only include the parameters that are in response to the interaction type the Authorization Server requested in the most recent response. The specific parameters are defined by each interaction type.
 
@@ -802,11 +802,12 @@ Except in error cases, the following key is required in the JSON document of the
 Depending on this assessment, the response from the Interactive Authorization Endpoint can take one of the following forms:
 
 ### Interaction Required Response {#iar-interaction-required-response}
+
 By setting `status` to `require_interaction` in the response, the Authorization Server requests an additional user interaction.
 In this case, the following keys MUST be present in the response as well:
 
 * `type`: REQUIRED. String indicating which type of interaction is required, as defined below. The Authorization Server MUST NOT set this to a value that was not included in the `interaction_types_supported` parameter sent by the Wallet.
-* `auth_session`: REQUIRED. String containing a value that allows the Authorization Server to associate subsequent requests by this Wallet with the ongoing authorization request sequence. Wallets SHOULD treat this value as an opaque value.
+* `auth_session`: REQUIRED. String containing a value that allows the Authorization Server to associate subsequent requests by this Wallet with the ongoing authorization request sequence. Wallets SHOULD treat this value as an opaque value. The value returned MUST be distinct for each interactive authorization response.
 
 The Wallet MUST include the most recently received `auth_session` in follow-up requests to the Interactive Authorization Endpoint.
 
@@ -950,7 +951,19 @@ Cache-Control: no-store
 }
 ```
 
-Once this phase of the Authorization process is completed, the Authorization Server MUST redirect back to the Wallet as per [@RFC6749]. If the Authorization process is complete when this redirect occurs, the Authorization Server returns a response with the `code` parameter as per Section 1.3.1 of [@RFC6749].
+Once this phase of the Authorization process is completed, the Authorization Server MUST redirect back to the Wallet as per [@RFC6749]. If the Authorization process is complete when this redirect occurs, the Authorization Server returns a response with the `code` parameter as per Section 1.3.1 of [@RFC6749]. If the Authorization process is not complete when this redirect occurs, the Authorization Server returns a response with the `auth_session` parameter. In the event a Wallet receives a response from the Authorization Server which features the `auth_session` parameter, the Wallet MUST make a follow-up request as per (#follow-up-request) to continue the Authorization process. In the event that PKCE as defined in [@RFC7636] was used in the initial authorization request to the interactive authorization endpoint, the Authorization Server MUST enforce the correct usage of the `code_verifier` in the follow-up request that follows the completion of the `redirect_to_web` interaction.
+
+To ensure the security of the `redirect_to_web` flow, the redirect URI MUST be an `https` URL as per Section 7.2 of [@!RFC8252]. The Wallet MUST NOT use an embedded user-agent to perform the `redirect_to_web` flow. The considerations in Section 8.12 of [@!RFC8252] apply. Platform-specific implementation details are provided in Appendix B of the same document.
+
+A non-normative example of a follow-up request featuring PKCE:
+
+```
+POST /iar HTTP/1.1
+Host: server.example.com
+Content-Type: application/x-www-form-urlencoded
+
+auth_session=wxroVrBY2MCq4dDNGXACS&code_verifier=avjebhrnqwketh
+```
 
 #### Custom Interaction Extensions {#iar-custom-extensions}
 


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -902,7 +902,17 @@ Cache-Control: no-store
...
+
+A Non-normative Example of a follow-up request featuring PKCE:
```
danielfett:
```suggestion
A non-normative example of a follow-up request featuring PKCE:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -902,7 +902,17 @@ Cache-Control: no-store
...
...f the Authorization process is completed, the Authorization Server MUST redirect back to the Wallet as per [@RFC6749]. If the Authorization process is complete when this redirect occurs, the Authorization Server returns a response with the `authorization_code` parameter as per Section 1.3.1 of [@RFC6749]. If the Authorization process is not complete when this redirect occurs, the Authorization Server returns a response with the `auth_session` parameter. In the event a Wallet receives a response from the Authorization Server which features the `auth_session` parameter, the Wallet MUST make a follow-up request as per (#follow-up-request) to continue the Authorization process. In the event that PKCE as defined in [@RFC7636] was used in the initial authorization request to the interactive authorization endpoint, the Wallet MUST include the `code_verifier` in the follow-up request that follows the completion of the `redirect_to_web` interaction. If the `code_verifier` parameter is present in a follow-up request as per (#follow-up-request), the Authorization Server MUST use the `code_verifier` parameter value to verify against the original `code_challenge` present in the initial request.
```
danielfett: The `authorization_code` is fresh information sent from the AS to the Wallet; the `auth_session` is not. There should in any case be a fresh value sent in the authorization response. This could either be a fresh `auth_session` or a separate value.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -902,7 +902,17 @@ Cache-Control: no-store
...
+
```
danielfett:
```suggestion

To ensure the security of the `redirect_to_web` flow, the redirect URI MUST be an `https` URL as per Section 7.2 of [@!RFC8252]. The Wallet MUST NOT use an embedded user-agent to perform the `redirect_to_web` flow. The considerations in Section 8.12 of [@!RFC8252] apply. Platform-specific implementation details are provided in Appendix B of the same document.

```

openid-4-verifiable-credential-issuance-1_0.md
danielfett: Also, PKCE should be mandatory for this flow.

Sakurann:
wg discussion:
- no other implementers out of those present on the call
- doing this in 1.1 is an option. potentially can be cleanly extended in 1.1, using existing extension points, using a new type in IAE.

openid-4-verifiable-credential-issuance-1_0.md
jogu: I agree we need to return a new value.

 I'm not sure I see a clear reason why this flow needs PKCE any more than the standard authorization code flow does (where VCI currently has it as 'RECOMMENDED'). If we do anything it should probably be to link to the `#securitybcp` section where we recommend FAPI in general.
 
 (I don't necessarily think it would be bad to make the language around this stronger in all cases, I just don't currently see the reason for making it stronger in just this particular case of `redirect_to_web`.)

awoie: > I agree we need to return a new value.
> 
> I'm not sure I see a clear reason why this flow needs PKCE any more than the standard authorization code flow does (where VCI currently has it as 'RECOMMENDED'). If we do anything it should probably be to link to the `#securitybcp` section where we recommend FAPI in general.
> 
> (I don't necessarily think it would be bad to make the language around this stronger in all cases, I just don't currently see the reason for making it stronger in just this particular case of `redirect_to_web`.)

I was about to say the same thing, so +1.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -902,7 +902,19 @@ Cache-Control: no-store
...
...f the Authorization process is completed, the Authorization Server MUST redirect back to the Wallet as per [@RFC6749]. If the Authorization process is complete when this redirect occurs, the Authorization Server returns a response with the `authorization_code` parameter as per Section 1.3.1 of [@RFC6749]. If the Authorization process is not complete when this redirect occurs, the Authorization Server returns a response with the `auth_session` parameter. In the event a Wallet receives a response from the Authorization Server which features the `auth_session` parameter, the Wallet MUST make a follow-up request as per (#follow-up-request) to continue the Authorization process. In the event that PKCE as defined in [@RFC7636] was used in the initial authorization request to the interactive authorization endpoint, the Wallet MUST include the `code_verifier` in the follow-up request that follows the completion of the `redirect_to_web` interaction. If the `code_verifier` parameter is present in a follow-up request as per (#follow-up-request), the Authorization Server MUST use the `code_verifier` parameter value to verify against the original `code_challenge` present in the initial request.
```
charsleysa: `authorization_code` -> `code`
```suggestion
Once this phase of the Authorization process is completed, the Authorization Server MUST redirect back to the Wallet as per [@RFC6749]. If the Authorization process is complete when this redirect occurs, the Authorization Server returns a response with the `code` parameter as per Section 1.3.1 of [@RFC6749]. If the Authorization process is not complete when this redirect occurs, the Authorization Server returns a response with the `auth_session` parameter. In the event a Wallet receives a response from the Authorization Server which features the `auth_session` parameter, the Wallet MUST make a follow-up request as per (#follow-up-request) to continue the Authorization process. In the event that PKCE as defined in [@RFC7636] was used in the initial authorization request to the interactive authorization endpoint, the Wallet MUST include the `code_verifier` in the follow-up request that follows the completion of the `redirect_to_web` interaction. If the `code_verifier` parameter is present in a follow-up request as per (#follow-up-request), the Authorization Server MUST use the `code_verifier` parameter value to verify against the original `code_challenge` present in the initial request.  
```

charsleysa:
I think this PR should be 1.0. The additional complexity is marginal if the wallet is already planning to support redirect_to_web.

The only thing I think maybe should be changed is that the `status` should also be included in the response (so `status=ok` for authorization code response and `status=require_interaction` for continuation).

From the security perspective, I don't see this being any more vulnerable than a typical authorization endpoint response and existing best practice mitigations should be sufficient.

openid-4-verifiable-credential-issuance-1_0.md
javereec: Why was the original text changed from the AS returning `code` to now returning `authorization_code` at the end of the interaction?

tplooker:
> I think the thought was that a new type could be defined that supported this; for example "redirect_to_web_version2" (though I'm sure we can come up with a better name).

I don't think this is a viable solution, we just end up largely duplicating this mechanism here with 2% difference, thats likely to lead to its own security challenges.

> The slight downside of how it is currently in this PR is that the wallet isn't permitted to opt in to the additional complexity. (Though maybe the additional complexity is small.)

To the contrary I don't think the wallet should be able to opt out here. Thats needless fragmentation which will just hurt interoperability. The complexity is also very low, a wallet supporting IAE will already know how to send a follow up request, all this PR does is detail how the wallet does this when receiving a new auth_session parameter from the Authorisation response.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -902,7 +900,19 @@ Cache-Control: no-store
...
+
+To ensure the security of the `redirect_to_web` flow, the redirect URI MUST be an `https` URL as per Section 7.2 of [@!RFC8252]. The Wallet MUST NOT use an embedded user-agent to perform the `redirect_to_web` flow. The considerations in Section 8.12 of [@!RFC8252] apply. Platform-specific implementation details are provided in Appendix B of the same document.
+
+A non-normative example of a follow-up request featuring PKCE:
```
javereec:
```suggestion
A non-normative example of a follow-up request, that follows the completion of the `redirect_to_web` interaction, featuring PKCE:
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -902,7 +900,19 @@ Cache-Control: no-store
...
+
+To ensure the security of the `redirect_to_web` flow, the redirect URI MUST be an `https` URL as per Section 7.2 of [@!RFC8252]. The Wallet MUST NOT use an embedded user-agent to perform the `redirect_to_web` flow. The considerations in Section 8.12 of [@!RFC8252] apply. Platform-specific implementation details are provided in Appendix B of the same document.
+
```
deshmukhrajvardhan: Couple of curious comments:
1. Will we rely on a pre-defined text for error handling? or specify it in here.
2. https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html in here details about `code_verifier` or  `code_challenge` aren't specified. Do we need to mention these terms there. e.g. 
Maybe update https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-5-2 to
"When the grant type `authorization_code` is used, it is RECOMMENDED to use PKCE [[RFC7636](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#RFC7636)] and Pushed Authorization Requests [[RFC9126](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#RFC9126)]. In that case `code_challenge` is present in the initial request and in later interactions, the Wallet MUST include the `code_verifier` which will be verified by the Authorization Server against the original `code_challenge` present in the initial request."

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -725,7 +725,7 @@ response_type=code
 
 ### Follow-up Request {#follow-up-request}
 
-Follow-up requests to the Interactive Authorization Endpoint only MUST include the `auth_session` value received most recently from the Authorization Server (see (#iar-interaction-required-response)).
+Follow-up requests to the Interactive Authorization Endpoint MUST include the `auth_session` value received most recently from the Authorization Server (see (#iar-interaction-required-response)).
```
jogu: I think the "only" was correct (if clumsy) - the first request must not include `auth_session`.

```suggestion
Follow-up requests to the Interactive Authorization Endpoint only MUST include the `auth_session` value received most recently from the Authorization Server (see (#iar-interaction-required-response)).
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -755,14 +755,12 @@ Except in error cases, the following key is required in the JSON document of the
...
 In this case, the following keys MUST be present in the response as well:
 
 * `type`: REQUIRED. String indicating which type of interaction is required, as defined below. The Authorization Server MUST NOT set this to a value that was not included in the `interaction_types_supported` parameter sent by the Wallet.
-* `auth_session`: REQUIRED. String containing a value that allows the Authorization Server to associate subsequent requests by this Wallet with the ongoing authorization request sequence. Wallets SHOULD treat this value as an opaque value.
-
-The Wallet MUST include the `auth_session` in all follow-up requests to the Interactive Authorization Endpoint.
-If, as a response to such a follow-up request, the Wallet receives an `auth_session` value that differs from the one sent in the request, it MUST abort the issuance process.
+* `auth_session`: REQUIRED. String containing a value that allows the Authorization Server to associate subsequent requests by this Wallet with the ongoing authorization request sequence. Wallets SHOULD treat this value as an opaque value. The value returned MUST be distinct for each interactive authorization response.
```
jogu: Think it needs to be unpredictable as well probably to get security benefits?

```suggestion
* `auth_session`: REQUIRED. String containing a value that allows the Authorization Server to associate subsequent requests by this Wallet with the ongoing authorization request sequence. Wallets SHOULD treat this value as an opaque value. The value returned MUST be distinct and unpredictable for each interactive authorization response.
```

openid-4-verifiable-credential-issuance-1_0.md
jogu: For 2, we don't need to include any extra normative text - the behaviour of PKCE with PAR etc is already well defined.

We do need this extra text (that Tobias added above) for `redirect_to_web` though as that behaviour isn't already defined.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -902,7 +900,19 @@ Cache-Control: no-store
...
...e this phase of the Authorization process is completed, the Authorization Server MUST redirect back to the Wallet as per [@RFC6749]. If the Authorization process is complete when this redirect occurs, the Authorization Server returns a response with the `code` parameter as per Section 1.3.1 of [@RFC6749]. If the Authorization process is not complete when this redirect occurs, the Authorization Server returns a response with the `auth_session` parameter. In the event a Wallet receives a response from the Authorization Server which features the `auth_session` parameter, the Wallet MUST make a follow-up request as per (#follow-up-request) to continue the Authorization process. In the event that PKCE as defined in [@RFC7636] was used in the initial authorization request to the interactive authorization endpoint, the Wallet MUST include the `code_verifier` in the follow-up request that follows the completion of the `redirect_to_web` interaction. If the `code_verifier` parameter is present in a follow-up request as per (#follow-up-request), the Authorization Server MUST use the `code_verifier` parameter value to verify against the original `code_challenge` present in the initial request.
```
jogu: I think this is the wrong way around - if a code_challenge was present, the followup request to the IAE MUST be rejected if code_verifier is absent?

I think I'd prefer language more like "Following a `redirect_to_web` that returns an `auth_session`, the Wallet and Issuer must do the PKCE checks in the next IAE call in the same way as they would be done at the token endpoint for the authorization code flow".

If we're not going to say something like that we need to be really careful with the wording. And/or need to refer to https://www.rfc-editor.org/rfc/rfc9700.html#section-4.8.2

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -902,7 +900,19 @@ Cache-Control: no-store
...
+
+To ensure the security of the `redirect_to_web` flow, the redirect URI MUST be an `https` URL as per Section 7.2 of [@!RFC8252]. The Wallet MUST NOT use an embedded user-agent to perform the `redirect_to_web` flow. The considerations in Section 8.12 of [@!RFC8252] apply. Platform-specific implementation details are provided in Appendix B of the same document.
+
+A non-normative example of a follow-up request, that follows the completion of the `redirect_to_web` interaction, featuring PKCE:
```
jogu: Maybe this is slightly better? No strong feeling though.

```suggestion
A non-normative example of a follow-up request, that could be sent following the completion of the `redirect_to_web` interaction, featuring PKCE:
```

openid-4-verifiable-credential-issuance-1_0.md
awoie: And short-lived similar to `code`?

openid-4-verifiable-credential-issuance-1_0.md
tplooker: `only` in this part of the sentence doesn't seem to fit for me? Do you mean only this parameter?

openid-4-verifiable-credential-issuance-1_0.md
tplooker: I was using `follows` here because its a follow up request, `could` seems to be less precise, because if you get auth session back in the auth response you have to do a follow up request, there is no choice.

openid-4-verifiable-credential-issuance-1_0.md
jogu: "only followup requests"

openid-4-verifiable-credential-issuance-1_0.md
jogu: I withdraw my suggestion then - thanks :)

openid-4-verifiable-credential-issuance-1_0.md
jogu: or to be extra clear:

Follow up requests must include `auth_session`.
The initial request in each session must not include `auth_session`.

openid-4-verifiable-credential-issuance-1_0.md
tplooker: For 1 yes we will rely on the standard error handling for OAuth 2 authorisation responses.

openid-4-verifiable-credential-issuance-1_0.md
tplooker: I think the original text is already clear is it not, specifically this piece

Follow-up requests to the Interactive Authorization Endpoint MUST include the `auth_session` value

Its both precise about the request this applies to and what MUST be included which is `auth_session`. Perhaps im missing your point?

jogu: It was the "initial request in each session must not include auth_session" bit that "only" adds a bit too.
I've read again, I think it's okay with your change as the "Initial request" section makes it pretty clear `auth_session` shouldn't be there in that case.

jogu:
> Discussed on today's WG call - consensus to merge once above issues resolved. Joseph to check to see if Stuttgart team could have a very quick check.

Update on security analysis of IAE (including this PR) - OIDF board has approved some further analysis of OID4VCI but we need to get the spec into a state where the analysis can be started...

PR #594, opened 2025-08-01
clarify that time-related information needs to be randomized

### LINKED ISSUE #592: timestamp fo the batch issued credentials
there was an issue raised regarding traceability concerns in batch issuance that all tokens are generated with the same timestamp, making them traceable. If the issuer generates a new timestamp for each attestation, during batch issuance, the issuance of 10/30/50 attestations often falls within the same second, or at most, spans two seconds.

Suggestion would be to add a privacy considerations section like the following (it is also in the sd-jwt rfc): 

"claims carrying time information, like iat, exp, and nbf, MUST either be randomized within a time period considered appropriate (e.g., randomize iat within the last 24 hours and calculate exp accordingly) or rounded (e.g., rounded down to the beginning of the day)."

Sakurann:
resolves #592

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1948,7 +1948,7 @@ methods, including but not limited to the following ones:
 * Issue a batch of Credentials with the same Credential Dataset to facilitate the use of a unique Credential per presentation or per Verifier. This approach solely aids in achieving Verifier-to-Verifier unlinkability.
 * Use cryptographic schemes that can provide non-correlation.
 
-In batch-issued Credentials, claims containing time-related information, such as issuance or expiration dates, SHOULD be either randomized within an appropriate time window (e.g., within the last 24 hours) or rounded (e.g., to the start of the day), to avoid unnecessary correlation.
+Claims containing time-related information, such as issuance or expiration dates, SHOULD be either individually randomized within an appropriate time window (e.g., within the last 24 hours), or rounded (e.g., to the start of the day), to avoid unintended correlation factors.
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1948,6 +1948,8 @@ methods, including but not limited to the following ones:
 * Issue a batch of Credentials with the same Credential Dataset to facilitate the use of a unique Credential per presentation or per Verifier. This approach solely aids in achieving Verifier-to-Verifier unlinkability.
 * Use cryptographic schemes that can provide non-correlation.
 
+In batch-issued Credentials, claims containing time-related information, such as issuance or expiration dates, SHOULD be either randomized within an appropriate time window (e.g., within the last 24 hours) or rounded (e.g., to the start of the day), to avoid unnecessary correlation.
```
c2bo:
```suggestion
In batch-issued Credentials, claims containing time-related information, such as issuance or expiration dates, SHOULD be either individually randomized within an appropriate time window (e.g., within the last 24 hours), or rounded (e.g., to the start of the day), to avoid unintended correlation factors.
```

jogu: I don't think the problem is limited to batch-issued credentials, building upon Christian's suggestion and generalising it a little:

```suggestion
Claims containing time-related information, such as issuance or expiration dates, SHOULD be either individually randomized within an appropriate time window (e.g., within the last 24 hours), or rounded (e.g., to the start of the day), to avoid unintended correlation factors.
```

jogu (approved):

sakimura (approved):

PR #596, opened 2025-08-02
Add statement that clarifies the use of scope and authorization_details  

### LINKED ISSUE #572: Clarify what absence of scope from credential_configurations_supported means
Some people are apparently pushing the interpretation that the optional "scope" element being missing from https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#section-11.2.4 means that the credential configuration id should be used as the scope instead.

I believe the correct interpretation is that RAR must be used. We should add a sentence like:

> If not present, the credential can only be requested using `authorization_details`

lchacha:
Fixes #572


selfissued (approved):
Thanks for doing this, @lchacha

charsleysa (approved):

jogu (approved):

lchacha:
I agree, I have moved the sentence to section 5.1.2 where we have similar clarification on the use of `scope` and `authorization_details`. @Sakurann

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1709,7 +1710,7 @@ This specification defines the following Credential Issuer Metadata parameters:
...
-  * `scope`: OPTIONAL. A JSON string identifying the scope value that this Credential Issuer supports for this particular Credential. The value can be the same across multiple `credential_configurations_supported` objects. The Authorization Server MUST be able to uniquely identify the Credential Issuer based on the scope value. The Wallet can use this value in the Authorization Request as defined in (#credential-request-using-type-specific-scope). Scope values in this Credential Issuer metadata MAY duplicate those in the `scopes_supported` parameter of the Authorization Server.
+  * `scope`: OPTIONAL. A JSON string identifying the scope value that this Credential Issuer supports for this particular Credential. The value can be the same across multiple `credential_configurations_supported` objects. The Authorization Server MUST be able to uniquely identify the Credential Issuer based on the scope value. The Wallet can use this value in the Authorization Request as defined in (#credential-request-using-type-specific-scope). Scope values in this Credential Issuer metadata MAY duplicate those in the `scopes_supported` parameter of the Authorization Server.
```
```suggestion
  * `scope`: OPTIONAL. A JSON string identifying the scope value that this Credential Issuer supports for this particular Credential. The value can be the same across multiple `credential_configurations_supported` objects. The Authorization Server MUST be able to uniquely identify the Credential Issuer based on the scope value. The Wallet can use this value in the Authorization Request as defined in (#credential-request-using-type-specific-scope). Scope values in this Credential Issuer metadata MAY duplicate those in the `scopes_supported` parameter of the Authorization Server. If `scope` is absent, the only alternate way for requesting Credential is using `authorization_details` [@!RFC9396], conditional to `authorization_details_types_supported` AS metadata being present.
```
@jogu @awoie is this better?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -550,6 +550,7 @@ Host: server.example.com
 ```
 
 If a scope value related to Credential issuance and the `authorization_details` request parameter containing objects of type `openid_credential` are both present in a single request, the Credential Issuer MUST interpret these individually. However, if both request the same Credential type, then the Credential Issuer MUST follow the request as given by the authorization details object.
+If `scope` is absent, the credential can only be requested using `authorization_details` [@!RFC9396].
```
```suggestion
```

openid-4-verifiable-credential-issuance-1_0.md
jogu: If 'scope' is absent and none of the ASes have `authorization_details_types_supported` then the configuration is just invalid and the wallet can't proceed. I don't know if we need to get into that.

Sakurann: the fact that this PR is needed to begin with, seems to be a prove that we need to get into this. also this is the text we discussed during the WG call i think

Sakurann: people were using configuration_ids instead of treating it as invalid, so yes

Sakurann (approved):

PR #600, opened 2025-08-05
Add example of signed request to IAR endpoint

### LINKED ISSUE #579: add example for signed request to IAR endpoint
It's possibly odd that in an signed request this would mean:

`"interaction_types_supported": "openid4vp_presentation,redirect_to_web"`

when possibly people might expect:

`"interaction_types_supported": [ "openid4vp_presentation", "redirect_to_web"]`

(Not sure I'm suggesting changing anything, but an example of the contents of a signed request could be useful to avoid misinterpretation.)

_Originally posted by @jogu in https://github.com/openid/OpenID4VCI/pull/509#discussion_r2208212623_

jogu:
closes #579

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -723,6 +723,32 @@ response_type=code
 &interaction_types_supported=openid4vp_presentation,redirect_to_web
 ```
 
+The following non-normative example shows an initial request to the Interactive Authorization Endpoint with a signed request object:
+
+```http
+POST /iar HTTP/1.1
+Host: server.example.com
+Content-Type: application/x-www-form-urlencoded
+
+request=eyJrd...
+```
+
+The following non-normative example shows a payload of a signed request object:
+
+```json
+{
+  "iss": "CLIENT1234",
+  "aud": "https://server.example.com",
+  "response_type": "code",
+  "client_id": "CLIENT1234",
+  "code_challenge": "E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM",
+  "code_challenge_method": "S256",
+  "redirect_uri": "https%3A%2F%2Fclient.example.org%2Fcb",
+  "authorization_details: {...},
+  "interaction_types_supported": "openid4vp_presentation,redirect_to_web"
+}
+```
+
 ### Follow-up Request {#follow-up-request}
 
 Follow-up requests to the Interactive Authorization Endpoint only MUST include the `auth_session` value received most recently from the Authorization Server (see (#iar-interaction-required-response)).


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -723,6 +723,32 @@ response_type=code
...
 
+The following non-normative example shows an initial request to the Interactive Authorization Endpoint with a signed request object:
+
+```http
+POST /iar HTTP/1.1
+Host: server.example.com
+Content-Type: application/x-www-form-urlencoded
+
+request=eyJrd...
+```
```
Sakurann: don't we need client authentication? per PR #601 ?

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -723,6 +723,32 @@ response_type=code
 &interaction_types_supported=openid4vp_presentation,redirect_to_web
 ```
 
+The following non-normative example shows an initial request to the Interactive Authorization Endpoint with a signed request object:
+
+```http
+POST /iar HTTP/1.1
+Host: server.example.com
```
jogu, line 730:
```suggestion
Host: server.example.com
OAuth-Client-Attestation: eyJ...
OAuth-Client-Attestation-PoP: eyJ...
```

openid-4-verifiable-credential-issuance-1_0.md
jogu: seems reasonable - done, thanks!

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -723,6 +723,34 @@ response_type=code
...
+```json
+{
+  "iss": "CLIENT1234",
+  "aud": "https://server.example.com",
+  "response_type": "code",
+  "client_id": "CLIENT1234",
+  "code_challenge": "E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM",
+  "code_challenge_method": "S256",
+  "redirect_uri": "https%3A%2F%2Fclient.example.org%2Fcb",
+  "authorization_details: {...},
```
charsleysa: shouldn't this be an array of objects? also missing quote
```suggestion
  "authorization_details": [...],
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -723,6 +723,34 @@ response_type=code
...
+```json
+{
+  "iss": "CLIENT1234",
+  "aud": "https://server.example.com",
+  "response_type": "code",
+  "client_id": "CLIENT1234",
+  "code_challenge": "E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM",
+  "code_challenge_method": "S256",
+  "redirect_uri": "https%3A%2F%2Fclient.example.org%2Fcb",
+  "authorization_details": [...],
```
jogu: Actually as it's so short I think we might as well add an actual real `authorization_details` here - here's one we use as an example elsewhere in the spec:

```suggestion
  "authorization_details": [
    {
      "type": "openid_credential",
      "credential_configuration_id": "UniversityDegreeCredential"
    }
  ],
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -723,6 +723,39 @@ response_type=code
...
+  "code_challenge_method": "S256",
+  "redirect_uri": "https%3A%2F%2Fclient.example.org%2Fcb",
+  "authorization_details": [
+    {
+      "type": "openid_credential",
+      "credential_configuration_id": "UniversityDegreeCredential"
+    }
+  ],
+  "interaction_types_supported": "openid4vp_presentation,redirect_to_web"
+}
```
Sakurann, line 756:
```suggestion
```json
{
  "iss": "CLIENT1234",
  "aud": "https://server.example.com",
  "response_type": "code",
  "client_id": "CLIENT1234",
  "code_challenge": "E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM",
  "code_challenge_method": "S256",
  "redirect_uri": "https%3A%2F%2Fclient.example.org%2Fcb",
  "authorization_details": [
    {
      "type": "openid_credential",
      "credential_configuration_id": "UniversityDegreeCredential"
    }
  ],
  "interaction_types_supported": "openid4vp_presentation,redirect_to_web",
  "nonce": "af0ifjsldkj"
}
```
what about nonce?
why it needs PKCE?

jogu: Note that other than expanding the content of authorization details, this is the same request as being made in the unsigned request just above.

nonce isn't required in a VCI request, as far as I know it doesn't have any defined meaning, am I missing something?

PKCE is recommended in https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#section-5 so including PKCE is just following the recommendation. The more underlying "why" is "because the IAR endpoint might result in a web based redirect, and the initial IAE call is the only place that can PKCE can currently be supplied". The PKCE challenge is I think not adding any security value if the AS never does a web redirect, but it's also harmless in that case. We could alternatively solve that by not supplying PKCE until the web redirect is required, but I can't currently think of a clean way to do that without reducing security.

Sakurann: yeah but then we should PKCE related guidance in IAR section. adding PKCE without anything in section 6 feels abrupt

jogu: I re-read all the text about PKCE and I think there's a few things we could do to make things clearer, including the point you mentioned about the IAR endpoint - raised https://github.com/openid/OpenID4VCI/issues/613 for this, thanks!

tplooker (approved):

GarethCOliver (approved):

PR #601, opened 2025-08-05
Add OAuth client authentication to various examples

### LINKED ISSUE #551: Add client authentication to some of the examples
Not a new problem, but the lack of any form of client authentication in this request is slightly disturbing :)

_Originally posted by @jogu in https://github.com/openid/OpenID4VCI/pull/509#discussion_r2116503744_

jogu:
Add the headers used for Wallet Attestations / OAuth 2.0 Attestation-Based Client Authentication to applicable examples.

closes #551

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -572,6 +572,8 @@ Below is a non-normative example of a Pushed Authorization Request:
 ```
 POST /par HTTP/1.1
 Host: server.example.com
+OAuth-Client-Attestation: eyJ...
+OAuth-Client-Attestation-PoP: eyJ...
 Content-Type: application/x-www-form-urlencoded
 
 response_type=code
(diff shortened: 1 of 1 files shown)


Sakurann (approved):
@paulbastian @c2bo @tplooker any chance you guys can easily generate complete examples?

Sakurann:
does IAR mandate client authentication? feels like that is not clear enough from the current spec text in section 6??

c2bo (approved):

jogu:
> does IAR mandate client authentication? feels like that is not clear enough from the current spec text in section 6??

There's text:

> Note: In case a Wallet Attestation is required by the Authorization Server, it has to be included in this request.

But I agree, it could be clearer... I'll open an issue.

jogu:
> > @paulbastian @c2bo @tplooker any chance you guys can easily generate complete examples?
> 
> Shouldn't be that hard, but I think adding one such example would be sufficient. Do you want me to add a PR that adds such an example to the new appendix for additional examples (https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#appendix-I) ?

Sounds good  - we have an example header/payload of a wallet attestation, but not an actual encoded signed one.

PR #614, opened 2025-08-08
add missing interaction type error code to IAE

### LINKED ISSUE #598: `interaction_types_supported` should probably be a metadata param and a request param; currently it is the latter only
When reading the IAR section, it says the following:

> interaction_types_supported: REQUIRED. Comma-separated list of strings indicating the types of interactions that the Authorization Server supports. The order of the values is not significant. The following values are defined by this specification:

> openid4vp_presentation: Indicates that the Authorization Server supports an OpenID4VP Presentation interaction, as defined in [Section 6.2.1.1](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#iar-require-presentation).
> redirect_to_web: Indicates that the Authorization Server supports a redirect to a web-based interaction, as defined in [Section 6.2.1.2](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#iar-redirect-to-web).

This section relates to the IAR request parameter (the Wallet provides), not to the AS metadata, hence should be updated accordingly as follows:

> interaction_types_supported: REQUIRED. Comma-separated list of strings indicating the types of interactions that the Wallet supports. The order of the values is not significant. The following values are defined by this specification:

> openid4vp_presentation: Indicates that the Wallet supports an OpenID4VP Presentation interaction, as defined in [Section 6.2.1.1](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#iar-require-presentation).
> redirect_to_web: Indicates that the Wallet supports a redirect to a web-based interaction, as defined in [Section 6.2.1.2](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#iar-redirect-to-web).

Additionally, the original text above should be moved to a dedicated AS metadata section and registered with IANA.

awoie:
Fixes #598

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -3193,6 +3193,18 @@ established by [@!RFC7591].
 * Change Controller: OpenID Foundation Digital Credentials Protocols Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
 * Reference: (#client-metadata) of this specification
 
+## OAuth Extensions Error Registry
+
+This specification registers the following errors in the IANA "OAuth Extensions Error" registry [@IANA.OAuth.Parameters] established by [@!RFC6755].
+
+### missing_interaction_type
+
+* Error name: `missing_interaction_type`
+* Error usage location: Interactive Authorization Error Response
+* Related protocol extension: OpenID for Verifiable Credential Issuance
+* Change controller: OpenID Foundation Digital Credentials Protocols Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
+* Specification document: (#iar-error-response) of this specification
+
 ## Well-Known URI Registry
 
 This specification registers the following well-known URI


tplooker (approved):

Sakurann (approved):

awoie:
> Does this need IANA registration in https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#extensions-error ?

I don't think so, since IAE is only similar to initial PAR requests but not interchangeable. Also, the IAE endpoint is fully under control of OID4VCI, hence I don't think it is required.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -969,6 +969,26 @@ Cache-Control: no-store
...
+
+```
+HTTP/1.1 400 Bad Request
+Content-Type: application/json
+Cache-Control: no-cache, no-store
+
+{
+  "error": "missing_interaction_type",
+  "error_description":
+    "The interaction_types_supported is missing a required interaction type"
```
jogu: It's worth guiding people towards returning an actionable error I think:

```suggestion
    "interaction_types_supported in the request is missing the required interaction type 'openid4vp_presentation'"
```

awoie:
> missing_interaction_type

I added the IANA registration section.

PR #615, opened 2025-08-08
add require_interactive_authorization_requests to AS metadata

### LINKED ISSUE #604: Metadata needed to signal requirement for Interactive Authorization Endpoint
Similar to PAR, we should introduce a metadata parameter to indicate whether the Interactive Authorization Endpoint (IAE) is required. Without such a signal, issuers/authorization servers would need to implement both flows: the traditional authorization endpoint and the IAE flow.

It is also needed to avoid dead end flows.

For PAR, we already have the `require_pushed_authorization_requests` flag. In my opinion, supporting both flows, the authorization endpoint, and IAE can be unnecessarily complex in certain scenarios, or even impossible.

I propose introducing the following metadata parameter:
- `require_iae_requests` (default: `false`)

awoie:
Fixes #604 

Additionally, and as an editorial change, I added `interactive_authorization_endpoint` to the AS metadata section, as it was previously missing. I also moved the corresponding normative "SHOULD use the IAE" statement into the AS metadata section, for consistency with how other endpoints are handled.

### DIFF
--- a/1.1/openid-4-verifiable-credential-issuance-1_1.md
+++ b/1.1/openid-4-verifiable-credential-issuance-1_1.md
@@ -640,7 +640,7 @@ This is an extension of the traditional Authorization Endpoint defined in [@!RFC
 A primary use case is requiring the Presentation of a Credential as a prerequisite for issuing a new Credential.
 Support for the Interactive Authorization Endpoint is OPTIONAL.
 
-The Authorization Server indicates support for interactive authorization by publishing the `interactive_authorization_endpoint` parameter in its Authorization Server Metadata. In this case, the Wallet SHOULD use this endpoint to obtain authorization.
+The Authorization Server indicates support for interactive authorization by publishing the `interactive_authorization_endpoint` parameter in its Authorization Server Metadata as defined in (#as-metadata).
 
 The following figure illustrates a flow using the Interactive Authorization Endpoint, where the Authorization Server requests a Presentation (of another Credential) from the Wallet as part of the authorization process to issue a Credential to that Wallet. The exact deployment model of the OpenID4VP Verifier in relation to the Authorization Server is out of scope of this specification. It can be integrated into the Authorization Server or a separate component, in which case backchannel communication between the Verifier and Authorization Server would need to happen (not shown here).
 
@@ -1836,7 +1836,9 @@ See (#additional-issuer-metadata-examples) for additional examples of Credential
 
 This specification also defines a new OAuth 2.0 Authorization Server metadata [@!RFC8414] parameter to publish whether the Authorization Server that the Credential Issuer relies on for authorization supports anonymous Token Requests with the Pre-Authorized Grant Type. It is defined as follows:
 
-* `pre-authorized_grant_anonymous_access_supported`: OPTIONAL. A boolean indicating whether the Credential Issuer accepts a Token Request with a Pre-Authorized Code but without a `client_id`. The default is `false`. 
+* `pre-authorized_grant_anonymous_access_supported`: OPTIONAL. A boolean indicating whether the Credential Issuer accepts a Token Request with a Pre-Authorized Code but without a `client_id`. The default is `false`.
+* `interactive_authorization_endpoint`: OPTIONAL. URL of the Authorization Server's Interactive Authorization Endpoint. This URL MUST use the https scheme and MAY contain port, path, and query parameter components. If omitted, the Authorization Server does not support the Interactive Authorization Endpoint. If present, the Wallet SHOULD use this endpoint to obtain authorization as defined in (#interactive-authorization-endpoint).
+* `require_interactive_authorization_request`: OPTIONAL. A boolean indicating whether the Authorization Server only accepts an Authorization Request for Credential issuance via the Interactive Authorization Endpoint defined in [#interactive-authorization-endpoint]. If omitted, the default value is `false`. This parameter MUST NOT be present if `interactive_authorization_endpoint` is omitted. Note that the presence of `interactive_authorization_endpoint` is sufficient for a Wallet to determine that it can use the Interactive Authorization Endpoint.
 
 Additional Authorization Server metadata parameters MAY be defined and used,
 as described in [@!RFC8414].
@@ -3255,7 +3257,7 @@ established by [@!RFC8414].
 * Metadata Name: `interactive_authorization_endpoint`
 * Metadata Description: URL of the Authorization Server's Interactive Authorization Endpoint. This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Authorization Server does not support the Interactive Authorization Endpoint.
 * Change Controller: OpenID Foundation Digital Credentials Protocols Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
-* Reference: (#interactive-authorization-request) of this specification
+* Reference: (#as-metadata) of this specification
 
 ## OAuth Dynamic Client Registration Metadata Registry
 
@@ -3457,3 +3459,5 @@ The technology described in this specification was made available from contribut
 
    * Initial draft created with same text as 1.0 Final
    * Add back Interactive Authorization Endpoint text that was removed from the 1.0 draft
+   * add require_interactive_authorization_request to AS metadata
+   * add interactive_authorization_endpoint to AS metadata section


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1762,7 +1762,9 @@ See (#additional-issuer-metadata-examples) for additional examples of Credential
...
+* `pre-authorized_grant_anonymous_access_supported`: OPTIONAL. A boolean indicating whether the Credential Issuer accepts a Token Request with a Pre-Authorized Code but without a `client_id`. The default is `false`.
+* `interactive_authorization_endpoint`: OPTIONAL. URL of the Authorization Server's Interactive Authorization Endpoint. This URL MUST use the https scheme and MAY contain port, path, and query parameter components. If omitted, the Authorization Server does not support the Interactive Authorization Endpoint. If present, the Wallet SHOULD use this endpoint to obtain authorization as defined in (#interactive-authorization-endpoint).
+* `require_interactive_authorization_requests`: OPTIONAL. A boolean indicating whether the Authorization Server accepts Authorization Request only via the Interactive Authorization Endpoint defined in (#interactive-authorization-endpoint). If omitted, the default value is `false`. This parameter MUST NOT be present if `interactive_authorization_endpoint` is omitted. Note that the presence of `interactive_authorization_endpoint` is sufficient for a Wallet to determine that it MAY use the Interactive Authorization Endpoint.
```
```suggestion
* `require_interactive_authorization_requests`: OPTIONAL. A boolean indicating whether the Authorization Server accepts Authorization Request only via the Interactive Authorization Endpoint defined in (#interactive-authorization-endpoint). If omitted, the default value is `false`. This parameter MUST NOT be present if `interactive_authorization_endpoint` is omitted. Note that the presence of `interactive_authorization_endpoint` is sufficient for a Wallet to determine that it can use the Interactive Authorization Endpoint.
```

charsleysa (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1762,7 +1762,9 @@ See (#additional-issuer-metadata-examples) for additional examples of Credential
...
+* `pre-authorized_grant_anonymous_access_supported`: OPTIONAL. A boolean indicating whether the Credential Issuer accepts a Token Request with a Pre-Authorized Code but without a `client_id`. The default is `false`.
+* `interactive_authorization_endpoint`: OPTIONAL. URL of the Authorization Server's Interactive Authorization Endpoint. This URL MUST use the https scheme and MAY contain port, path, and query parameter components. If omitted, the Authorization Server does not support the Interactive Authorization Endpoint. If present, the Wallet SHOULD use this endpoint to obtain authorization as defined in (#interactive-authorization-endpoint).
+* `require_interactive_authorization_requests`: OPTIONAL. A boolean indicating whether the Authorization Server accepts Authorization Request only via the Interactive Authorization Endpoint defined in (#interactive-authorization-endpoint). If omitted, the default value is `false`. This parameter MUST NOT be present if `interactive_authorization_endpoint` is omitted. Note that the presence of `interactive_authorization_endpoint` is sufficient for a Wallet to determine that it can use the Interactive Authorization Endpoint.
```
paulbastian: Could you get the same effect by omitting the "normal" authorization endpoint?

awoie: I think you are right! For some reason, I thought that IAE used the `authorization_endpoint` similar to PAR. Because we are referring to PAR. But unlike PAR, the IAE flow uses the IAE endpoint only. I believe PAR required the flag in the AS metadata because it uses the `authorization_endpoint` as well for the follow-up request.

RFC8414 says:
> authorization_endpoint
      URL of the authorization server's authorization endpoint
      [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)].  This is REQUIRED unless no grant types are supported
      that use the authorization endpoint.

In this case, it should be fine to omit the `authorization_endpoint`.

jogu: > Could you get the same effect by omitting the "normal" authorization endpoint?

Not as clearly I think - although thinking about this more I think this all gets weird. It certainly feels like it's pointless listing a `require_interactive_authorization_requests` if there's say an authorization endpoint or CIBA or device code flow or ... are supported.

It's also somewhat problematic if the authorization server is a general purpose one that also serves non-VCI use cases - then clients would presumably need to ignore `require_interactive_authorization_requests`.

Maybe this?

```suggestion
* `require_interactive_authorization_requests`: OPTIONAL. A boolean indicating whether the Authorization Server accepts Authorization Request that request Credential issuance only via the Interactive Authorization Endpoint defined in (#interactive-authorization-endpoint). If omitted, the default value is `false`. This parameter MUST NOT be present if `interactive_authorization_endpoint` is omitted. Note that the presence of `interactive_authorization_endpoint` is sufficient for a Wallet to determine that it can use the Interactive Authorization Endpoint.
```

jogu (approved):

jogu:
Discussed on today's WG call; Raj offered to review, Paul to re-review.

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1815,7 +1815,9 @@ See (#additional-issuer-metadata-examples) for additional examples of Credential
 
 This specification also defines a new OAuth 2.0 Authorization Server metadata [@!RFC8414] parameter to publish whether the Authorization Server that the Credential Issuer relies on for authorization supports anonymous Token Requests with the Pre-Authorized Grant Type. It is defined as follows:
 
-* `pre-authorized_grant_anonymous_access_supported`: OPTIONAL. A boolean indicating whether the Credential Issuer accepts a Token Request with a Pre-Authorized Code but without a `client_id`. The default is `false`. 
+* `pre-authorized_grant_anonymous_access_supported`: OPTIONAL. A boolean indicating whether the Credential Issuer accepts a Token Request with a Pre-Authorized Code but without a `client_id`. The default is `false`.
+* `interactive_authorization_endpoint`: OPTIONAL. URL of the Authorization Server's Interactive Authorization Endpoint. This URL MUST use the https scheme and MAY contain port, path, and query parameter components. If omitted, the Authorization Server does not support the Interactive Authorization Endpoint. If present, the Wallet SHOULD use this endpoint to obtain authorization as defined in (#interactive-authorization-endpoint).
```
paulbastian: Shouldn't the AS metadata also declare the interaction_types_required or interaction_types_supported? The pattern that the Wallet is sending interaction_types_supported seems uncommon to OAuth from my perspective. I would appreciate that as some form of authentication negotiation over the AS metadata

paulbastian (changes requested):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1815,7 +1815,9 @@ See (#additional-issuer-metadata-examples) for additional examples of Credential
...
+* `require_interactive_authorization_requests`: OPTIONAL. A boolean indicating whether the Authorization Server accepts Authorization Request that request Credential issuance only via the Interactive Authorization Endpoint defined in (#interactive-authorization-endpoint). If omitted, the default value is `false`. This parameter MUST NOT be present if `interactive_authorization_endpoint` is omitted. Note that the presence of `interactive_authorization_endpoint` is sufficient for a Wallet to determine that it can use the Interactive Authorization Endpoint.
```
paulbastian: We have a SHOULD recommendation, that IAE is preferred over normal auth endpoint. Shouldn't the default value be true then?

awoie: The SHOULD recommendation helps wallets to decided whether to use the `authorization_endpoint` or the `interactive_authorization_endpoint` whereas the `require_interactive_authorization_requests` just says that this is the only option. The semantics are different and not entirely related.

openid-4-verifiable-credential-issuance-1_0.md
awoie: Yep, this was discussed here https://github.com/openid/OpenID4VCI/issues/598 but we decided that a dedicated error code is sufficient. We could add an AS metadata for this in the future. IMO, a separate issue. You could ask to reopen the issue if you think this feature is required.

awoie:
> may not be directly related to this PR, but I think if we touch the AS metadata in this PR it should be complete. Was this previously discussed?

Yes, see above (https://github.com/openid/OpenID4VCI/issues/598)

openid-4-verifiable-credential-issuance-1_0.md
deshmukhrajvardhan: suggestion:
"
`require_interactive_authorization_requests`: OPTIONAL. A boolean indicating whether the Authorization Server only accepts Authorization Request for Credential issuance via the Interactive Authorization Endpoint defined in [#interactive-authorization-endpoint].
"

deshmukhrajvardhan: Also, should it be `require_interactive_authorization_request` instead of `require_interactive_authorization_requests`?

awoie: Applied your suggestions. Thanks!

Sakurann:
WG discussion:
- suggestion is to merge this and open a new issue/PR on Paul's concern with more metadata. @paulbastian to look into it and confirm

PR #618, opened 2025-08-12
add example for wallet attestation jwt

### LINKED ISSUE #608: As example signed wallet attestation
> @paulbastian @c2bo @tplooker any chance you guys can easily generate complete examples?

Shouldn't be that hard, but I think adding one such example would be sufficient. Do you want me to add a PR that adds such an example to the new appendix for additional examples (https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#appendix-I) ?

_Originally posted by @c2bo in https://github.com/openid/OpenID4VCI/issues/601#issuecomment-3159850339_

c2bo:
Closes #608 

Adds an example of a signed Wallet Attestation JWT and slightly changes the existing non-normative example.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -702,10 +702,10 @@ Note: In case a Wallet Attestation is required by the Authorization Server, it h
 
 The initial request to the Interactive Authorization Endpoint is formed and sent in the same way as PAR request as defined in Section 2.1 of [@!RFC9126]. The contents of the request are the same as in a regular Authorization Request as defined in (#credential-authz-request), with the following addition:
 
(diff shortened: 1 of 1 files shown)


jogu:
> I agree that it doesn't really help in the `cnf` claim, but implementation should be able to parse `use` for keys in metadata etc. and if I recall correctly checking in those cases is also heavily encouraged in security related profiles like FAPI (not entirely sure, need to check)?
> 
> I can adjust this example (inside cnf) if we think that makes sense - any other opinions?

It's weird. I can't see any use of having a `use` claim, but also it's present in the example: https://datatracker.ietf.org/doc/html/rfc7800#section-3.2

My take would be that FAPI doesn't require `use` to be present as there's no key selection going on - there can only be a single key in `cnf` and it is only used for signature verification as far as I know...

Not sure there's any need to change the example, though probably we need to have a conformance test for that tries both variants :'(

PR #619, opened 2025-08-12
editorial: fixes IAE initial request example

awoie:
Editorial: fixes IAE initial request example. Interaction types have to be percent encoded.


jogu (approved):
Good catch!

For anyone wondering, https://url.spec.whatwg.org/#percent-encoded-bytes is I believe the canonical reference for why you need to encode the comma.

tplooker (approved):

Sakurann (approved):

PR #626, opened 2025-08-13
Clarify who picks wallet attestation 'sub'

### LINKED ISSUE #611: Selection of 'client_id' when using wallet attestations
The spec currently says:

> The sub claim of the Wallet Attestation JWT is picked by the Wallet Provider and represents the client_id of the Wallet instance.

I don't think "picked by the Wallet Provider"  is really true/generic. In some cases the client id will I think be assigned by the AS as a result of registering the client id, in other cases I think it might somehow be dictated by the x5c certificate being used. In some cases (to allow pre-generation of Wallet attestations) it would need to be a value assigned at the ecosystem level.

jogu:
Remove a bit of text that could be read to imply that the wallet provider has a free choice as to the 'sub' value used.

closes #611

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -2971,7 +2971,7 @@ The following is a non-normative example of a Wallet Attestation:
 
 To use the Wallet Attestation towards the Authorization Server, the Wallet MUST generate a proof of possession according to Section 5.2 "Client Attestation PoP JWT" of Attestation-Based Client Authentication.
 
-The `sub` claim of the Wallet Attestation JWT is picked by the Wallet Provider and represents the `client_id` of the Wallet instance. For privacy reasons, this value is the same across Wallet instances of that Wallet Provider, see (#walletattestation-sub) for more details.
+The `sub` claim of the Wallet Attestation JWT is the `client_id` of the Wallet instance. For privacy reasons, this value is the same across Wallet instances of that Wallet Provider, see (#walletattestation-sub) for more details.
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2971,7 +2971,7 @@ The following is a non-normative example of a Wallet Attestation:
 
 To use the Wallet Attestation towards the Authorization Server, the Wallet MUST generate a proof of possession according to Section 5.2 "Client Attestation PoP JWT" of Attestation-Based Client Authentication.
 
-The `sub` claim of the Wallet Attestation JWT is picked by the Wallet Provider and represents the `client_id` of the Wallet instance. For privacy reasons, this value is the same across Wallet instances of that Wallet Provider, see (#walletattestation-sub) for more details.
+The `sub` claim of the Wallet Attestation JWT is the `client_id` of the Wallet instance. For privacy reasons, this value is the same across Wallet instances of that Wallet Provider, see (#walletattestation-sub) for more details.
```
charsleysa, line 2974: doesn't this somewhat conflict with [15.4.4. ](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#section-15.4.4)[Wallet Attestation Subject](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-wallet-attestation-subject)

charsleysa: on further thought, I think the text is sufficient

Issue #620, opened 2025-08-12
IAE URL is not compatible with Origin/expected_origin


awoie:
Two comments:
1. since expected_origins is optional (or conditional), should we add "if present"?
2. expected_origins can only contain origins, not URLs. While they can match, they don't always match. Is the idea that the expected_origins is converted by the wallet before matching?

_Originally posted by @martijnharing in https://github.com/openid/OpenID4VCI/pull/602#discussion_r2269368694_

Also

https://github.com/openid/OpenID4VCI/pull/602#discussion_r2269424470

Current definition of `Origin` and `expected_origins` in OID4VP clashes with this normative statement. `expected_origins` contains origin values and is conditional based on if the request was signed. If somebody wants to write a parser for requests, this would be prone to errors since with the current text both, URLs and origins, are allowed.

tlodderstedt:
closing this issues as it is superceded by #672

PR #629, opened 2025-08-13
fix: just do derived origin binding

awoie:
This PR makes the change to do derived origin binding instead of full URL endpoint binding for audience/session transcript/domain.


Sakurann:
WG discussion. Close this PR, open a new one once there is agreement on the direction of derived origin binding. Most likely we need to work on origin terminology throughout the spec. there is an issue #620 to discuss that

awoie:
Closing since we have a separate issue for the underlying potential issue https://github.com/openid/OpenID4VCI/issues/620

PR #637, opened 2025-08-25
Clarifications regarding mismatches in number of proofs, batch size, and number of issued credentials

joelposti:
I discovered that mismatches between the number of proofs and `batch_size`, and between the number of proofs and issued credentials are somewhat unspecified. It was not entirely clear what the issuer or the wallet should do in case of a mismatch.

In the first commit I added the word ‘issued’ to a sentence in section ‘10.1. Notification Request’ to make it clearer that the wallet should not treat the overall issuance flow failed, if it received fewer credentials than it expected from the issuer (ie. wallet received fewer credentials than it sent keys).

In the second commit I added paragraph to section ‘8.2. Credential Request’ as guidance on what the issuer needs to do, if the `proofs` parameters contains more proofs than the issuer expects. My suggestion is that the issuer must ignore the extra proofs. As an alternative, I think its fine for the issuer to be stricter and return a credential error response in those cases.

As an additional note while reading the various sentences and sections in the standard I noticed that the terminology with regards to the proofs and the keys that they prove could be better. In some parts of the standard it seems to be presumed that `one proof = one key bound to one credential`. For example this bit about `batch_size`: ‘Integer value specifying the maximum array size for the proofs parameter in a Credential Request.’ This presumption is problematic because one `attestation` proof and one `jwt` proof type with an embedded key attestation can contain many keys and thus the presumption becomes invalid. I will propose a new term Proven Key in another pull request.


Sakurann:
i am ok adding a non-normative implementation considerations about this - opened a new PR #643 given the timeline for 1.0.

PRs a very very welcome, but please try open an issue first. thank you!

PR #642, opened 2025-09-02
Move Encrypted Messages to the top level

### LINKED ISSUE #638: Move Encrypted Messages Section
Currently, Encrypted Messages is under the Deferred Credential Endpoint section, but applies to both Credential Endpoint and Deferred Credential Endpoint. 

It should either be
  1) at a top level (to be referenced by both)
  2) be in Credential Endpoint section (as Deferred Credential Endpoint text often refers back to text in that section).

GarethCOliver:
Resolve #638

It wasn't clear from the comment as to which way the working group wished to resolve this. Moved it to the top level in this PR, mostly so that it could also be referenced by other sections (such as IAR) in the future, but happy to move it instead if people prefer.

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1168,8 +1168,8 @@ Cache-Control: no-store
 
 In the event the Credential Issuer can no longer issue the credential(s), the `credential_request_denied` error code as defined in (#credential-request-errors) should be used in response to a request. A wallet upon receiving this error SHOULD stop making requests to the deferred credential endpoint for the given `transaction_id`. 
 
-# Encrypted Messages {#encrypted-messages}
-Encryption of Request and Response Messages is performed as follows:
+# Encrypted Credential Requests and Responses {#encrypted-messages}
+Encryption of requests and responses for the Credential and Deferred Credential Endpoints is performed as follows:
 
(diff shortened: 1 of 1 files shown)


Sakurann (approved):

tplooker (approved):

c2bo (approved):

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1168,7 +1168,7 @@ Cache-Control: no-store
 
 In the event the Credential Issuer can no longer issue the credential(s), the `credential_request_denied` error code as defined in (#credential-request-errors) should be used in response to a request. A wallet upon receiving this error SHOULD stop making requests to the deferred credential endpoint for the given `transaction_id`. 
 
-## Encrypted Messages {#encrypted-messages}
+# Encrypted Messages {#encrypted-messages}
```
jogu: A more specific heading might help now it's a top level heading:

```suggestion
# Encrypted Credential Requests and Responses {#encrypted-messages}
```

openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1168,7 +1168,7 @@ Cache-Control: no-store
...
 Encryption of Request and Response Messages is performed as follows:
```
```suggestion
Encryption of requests and responses for the Credential and Deferred Credential Endpoints is performed as follows:
```

PR #643, opened 2025-09-02
add implementation considerations for the batch issuance of the credentials

Sakurann:
resolves #637

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1575,6 +1575,10 @@ Credential Refresh can be initiated by the Wallet independently from the Credent
 
 It is up to the Credential Issuer whether to update both the signature and the claim values, or only the signature.
 
+## Batch Issuing Credentials
+
+When the `batch_credential_issuance` parameter is not present in the Issuer metadata, it is at the discretion of the Credential Issuer to determine the number of the Credentials issued in the Credential Response, regardless of number of proofs contained in the `proofs` parameter in the Credential Request.
+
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1575,6 +1575,10 @@ Credential Refresh can be initiated by the Wallet independently from the Credent
 
 It is up to the Credential Issuer whether to update both the signature and the claim values, or only the signature.
 
+## Batch Issuing Credentials
+
+When the `batch_credential_issuance` parameter is not present in the Issuer metadata, it is at the discretion of the Credential Issuer to determine the number of the Credentials issued in the Credential Response, regardless of number of proofs contained in the `proofs` parameter in the Credential Request.
```
jogu: I'm not really sure what we're trying to clarify here, but I think the "When" adds extra confusion as the rest of the sentence is true when `batch_credential_issuance` is present too I believe. So this sentence would at least be clearer I think:

```suggestion
It is at the discretion of the Credential Issuer to determine the number of the Credentials issued in the Credential Response, regardless of number of proofs contained in the `proofs` parameter in the Credential Request.
```

jogu: This sentence would be a further simplification to the language that I think still conveys the same meaning:

```suggestion
The Credential Issuer determines the number of the Credentials issued in the Credential Response, regardless of number of proofs/keys contained in the `proofs` parameter in the Credential Request.
```

jogu (approved):

tplooker (approved):

PR #649, opened 2025-09-09
clarify PKCE session binding

### LINKED ISSUE #562: [Editorial] PKCE is not device bound
https://github.com/openid/OpenID4VCI/blob/main/openid-4-verifiable-credential-issuance-1_0.md?plain=1#L1417

mentions that `not bound to a certain device (as the Authorization Code Flow does with PKCE)`
PKCE doesn't bound the auth code to the device if I understand it right

https://datatracker.ietf.org/doc/html/rfc7636#section-4.1

I think it prevents MITM and reuse of the auth code for an attack as the code challenge and verifier is required in addition.

Suggestion:
`The Pre-Authorized Code Flow is vulnerable to the replay of the Pre-Authorized Code, because by design, it doesn't have additional challenge and verification (as the Authorization Code Flow does with PKCE)`

please feel free to suggest other wording

Sakurann:
resolves #562

### DIFF
--- a/openid-4-verifiable-credential-issuance-1_0.md
+++ b/openid-4-verifiable-credential-issuance-1_0.md
@@ -1481,7 +1481,7 @@ The Credential Issuer MUST ensure the release of any privacy-sensitive data in C
 
 ### Replay Prevention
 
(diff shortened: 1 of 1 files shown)


c2bo (approved):

Sakurann:
wg discussion. will merge as part of preparation for 1.0 publication given there are more than 3 approvals

PR #628, opened 2025-08-13
Fixes #627: "normative" typo in IAE section

awoie:
Fixes #627: "normative" typo in IAE section

Issue #627, opened 2025-08-13
`openid4vp_request` param in IAE "interaction required" response is wrong


awoie:
The following is in the spec:
> The Interactive Authorization Request, which is used to submit the OpenID4VP Authorization Response MUST satisfy the requirements set out in [Section 6.1.2](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#follow-up-request). In addition to these requirements, the request MUST also contain the openid4vp_request request parameter. The value of the openid4vp_request request parameter is a JSON-encoded object that encodes the OpenID4VP Authorization Response parameters.

There is probably a typo and should be fixed like this:
> The Interactive Authorization Request, which is used to submit the OpenID4VP Authorization Response MUST satisfy the requirements set out in [Section 6.1.2](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#follow-up-request). In addition to these requirements, the request MUST also contain the openid4vp_response request parameter.  The value of the openid4vp_response request parameter is a JSON-encoded object that encodes the OpenID4VP Authorization Response parameters. 

Also the example needs to be updated:
```http
POST /iar HTTP/1.1
Host: server.example.com
OAuth-Client-Attestation: eyJ...
OAuth-Client-Attestation-PoP: eyJ...
Content-Type: application/x-www-form-urlencoded

auth_session=wxroVrBY2MCq4dDNGXACS
&openid4vp_presentation=...
```

PR #624, opened 2025-08-13
Improvements to text about PKCE / IAR vs PAR / security BCP.

jogu:
Clarify that authorization code flow section now that we have a non-frontchannel method (IAE) to obtain authorization_code.

Clarify that the recommendation to use PAR means PAR or IAE.

closes #613

PR #623, opened 2025-08-13
Fix for Issue 607

tplooker:
Fixes #607

PR #657, opened 2025-09-22
Initial 1 1 draft

### LINKED ISSUE #607: Make clear client auth required at IAR endpoint
does IAR mandate client authentication? feels like that is not clear enough from the current spec text in section 6??

_Originally posted by @Sakurann in https://github.com/openid/OpenID4VCI/issues/601#issuecomment-3159815926_

jogu:
Copies the 1.0 final text, change version, change title, add doc history back, remove text about erratas

Apply 3 of the open IAE PRs against 1.0 that appeared to have consensus (#624, #623, #628)

closes #607 #613 #627


jogu:
Two things I wasn't 100% sure about:

1. I restarted the draft numbering at 01 (not sure if it should have been 00, but pretty sure it shouldn't continue on from where 1.0 ended)

2. The file published to github pages doesn't have a 1.0 / 1.1 version number in it, so I made it so we just publish 1.1 to the same url as we were publishing 1.0. I could add a new url for 1.1 instead?

jogu:
Discussed on APAC call this morning: Would be nice if what ends up in git history in the repo clearly shows the text from 1.0 and then has commits for all the changes. Joseph can open PR that has just the first commit included.

PR #660, opened 2025-09-25
Changes to update 1.1 spec to be a 1.1 draft

jogu:
Change version, change title, add doc history back, remove text about erratas, update github action to publish 1.1 (instead of 1.0).


c2bo:
Does it make sense to keep the old 1.0 markdown next to the 1.1? Should we at least move it to some folder to avoid confusion?

PR #678, opened 2025-11-11
Rephrase conditions to provide `nonce` in proof types based on presence of Nonce endpoint

### LINKED ISSUE #676: key attestation in the jwt proof header should be able to be pre-generated...
The intent of having both attestation and jwt proof types was the following:
A. jwt proof type: `key_attestation` in the jwt proof header can be pre-generated (attested keys in the attestation can also be pre-generated), and jwt proof signed using one of the attested keys needs to be generated fresh since it contains nonce from the issuer.
B. attestation proof type: all attested keys can be pre-generated, and key attestation needs to be generated fresh since it contains nonce from the issuer.

However, there is this text for key_attestation in the jwt proof header, which defeats the intent above because key attestation in jet proof type cannot be pre-generated since it needs to contain a nonce:
> key_attestation: OPTIONAL. JOSE Header containing a key attestation as described in Appendix D. If the Credential Issuer provided a c_nonce, the nonce claim in the key attestation MUST be set to a server-provided c_nonce.

original issue #438 pointed to [a nonce claim in key attestation jwt itself](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-1_1-wg-draft.html#appendix-D.1-4.1.2.7) but the PR #501 added clarification to [key attestation claim in the jwt header itself](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-1_1-wg-draft.html#appendix-F.1-2.1.2.6)

I think we have two options (still investigating feasibility of both; not advocating for any of these at this point):
1. do an errata of VCI and fix the above by removing "If the Credential Issuer provided a c_nonce, the nonce claim in the key attestation MUST be set to a server-provided c_nonce."
3. define a new proof type in VCI 1.1 that achieves the purpose A above that jwt proof type was supposed to achieve
4. define a new proof type in HAIP 1.0 that achieves the purpose A above that jwt proof type was supposed to achieve

### LINKED ISSUE #677: "If the Credential Issuer provided a c_nonce" doesn't really make sense
We should replace all 3 instances with "If the Credential Issuer has a nonce_endpoint", the current text is a hangover from when c_nonce was provided in credential/token endpoint responses. (Although one of the instances may be removed by #676 )

awoie:
Potentially fixes #677 , potentially fixes #676 

Note that I created the PR based on https://github.com/openid/OpenID4VCI/issues/676#issuecomment-3501443791.

IMO, one implication is that for `nonce` claims in a `key_attestation` in a `jwt` proof, it means, the wallet decides whether to include it which is how I interpret the current version of the spec but wanted to point this out in case it is not obvious for readers of this PR. If the issuer insists on the presence which is unlikely, it could still provide a nonce error. To improve this behaviour, we could define a dedicated issuer metadata parameter, e.g., `require_nonce_in_key_attesatation_in_jwt_proof` in a backward compatible way to improve this behaviour.

### DIFF
--- a/1.0/openid-4-verifiable-credential-issuance-1_0.md
+++ b/1.0/openid-4-verifiable-credential-issuance-1_0.md
@@ -2708,7 +2708,7 @@ When a W3C Verifiable Presentation as defined by [@VC_DATA_2.0] or [@VC_DATA] se
   * `cryptosuite`: REQUIRED. If Credential Issuer metadata is provided, the value MUST match one of the entries in the `proof_signing_alg_values_supported` metadata parameter.
   * `proofPurpose`: REQUIRED. MUST be set to `authentication`.
   * `domain`: REQUIRED. MUST be set to the Credential Issuer Identifier.  
-  * `challenge`: OPTIONAL. The value type of this claim MUST be a string, where the value is a server-provided `c_nonce`. It MUST NOT be used otherwise. It MUST be present when the Issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
+  * `challenge`: OPTIONAL. The value type of this claim MUST be a string, where the value is a server-provided `c_nonce`. It MUST be present when the Issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
 
 The Credential Issuer MUST validate that the W3C Verifiable Presentation used as a proof is actually signed with a key in the possession of the Holder.
 
--- a/1.0/openid-4-verifiable-credential-issuance-1_0.md
+++ b/1.0/openid-4-verifiable-credential-issuance-1_0.md
@@ -1047,7 +1047,7 @@ If the Wallet is requesting the issuance of a Credential that is not supported b
   * `unknown_credential_configuration`: Requested Credential Configuration is unknown.
   * `unknown_credential_identifier`: Requested Credential identifier is unknown.
   * `invalid_proof`: The `proofs` parameter in the Credential Request is invalid: (1) if the field is missing, or (2) one of the provided key proofs is invalid, or (3) if at least one of the key proofs does not contain a `c_nonce` value (refer to (#nonce-response)).
-  * `invalid_nonce`: The `proofs` parameter in the Credential Request uses an invalid nonce: at least one of the key proofs contains an invalid `c_nonce` value. The wallet should retrieve a new `c_nonce` value (refer to (#nonce-endpoint)).
+  * `invalid_nonce`: The `proofs` parameter in the Credential Request uses an invalid nonce: at least one of the key proofs contains an invalid `c_nonce` value. The Wallet should retrieve a new `c_nonce` value (refer to (#nonce-endpoint)).
   * `invalid_encryption_parameters`: This error occurs when the encryption parameters in the Credential Request are either invalid or missing. In the latter case, it indicates that the Credential Issuer requires the Credential Response to be sent encrypted, but the Credential Request does not contain the necessary encryption parameters.
   * `credential_request_denied`: The Credential Request has not been accepted by the Credential Issuer. The Wallet SHOULD treat this error as unrecoverable, meaning if received from a Credential Issuer the Credential cannot be issued. 
 * `error_description`: OPTIONAL. The `error_description` parameter MUST be a human-readable ASCII [@!USASCII] text, providing any additional information used to assist the Client implementers in understanding the occurred error. The values for the `error_description` parameter MUST NOT include characters outside the set `%x20-21 / %x23-5B / %x5D-7E`.
@@ -2630,14 +2630,14 @@ The JWT MUST contain the following elements:
   * `kid`: OPTIONAL. JOSE Header containing the key ID. If the Credential is to be bound to a DID, the `kid` refers to a DID URL which identifies a particular key in the DID Document that the Credential is to be bound to. It MUST NOT be present if `jwk` or `x5c` is present.
   * `jwk`: OPTIONAL. JOSE Header containing the key material the new Credential is to be bound to. It MUST NOT be present if `kid` or `x5c` is present.
   * `x5c`: OPTIONAL. JOSE Header containing at least one certificate where the first certificate contains the key that the Credential is to be bound to, additional certificates may also be present. It MUST NOT be present if `kid` or `jwk` is present.
-  * `key_attestation`: OPTIONAL. JOSE Header containing a key attestation as described in (#keyattestation). If the Credential Issuer provided a `c_nonce`, the `nonce` claim in the key attestation MUST be set to a server-provided `c_nonce`.
+  * `key_attestation`: OPTIONAL. JOSE Header containing a key attestation as described in (#keyattestation). If the `nonce` claim is present in the key attestation, its value MUST be set to a server-provided `c_nonce` from the Nonce Endpoint as defined in (#nonce-endpoint). Note that including a `nonce` claim is left to the Wallet. In some environments, a `nonce` is unnecessary because the key material in the `key_attestation` already provides sufficient entropy and freshness. Omitting the `nonce` also enables pre-generation of attestations prior to interacting with a specific Issuer. If the Issuer returns an `invalid_nonce` error, this can be interpreted as an indication that the Issuer expects a `nonce` claim in the `key_attestation`.
   * `trust_chain`: OPTIONAL. JOSE Header containing an [@!OpenID.Federation] Trust Chain. This element MAY be used to convey key attestation, metadata, metadata policies, federation Trust Marks and any other information related to a specific federation, if available in the chain. When used for signature verification, the header parameter `kid` MUST be present.
 
 * in the JWT body,
   * `iss`: OPTIONAL (string). The value of this claim MUST be the `client_id` of the Client making the Credential request. This claim MUST be omitted if the access token authorizing the issuance call was obtained from a Pre-Authorized Code Flow through anonymous access to the token endpoint.
   * `aud`: REQUIRED (string). The value of this claim MUST be the Credential Issuer Identifier.
   * `iat`: REQUIRED (number). The value of this claim MUST be the time at which the key proof was issued using the syntax defined in [@!RFC7519].
-  * `nonce`: OPTIONAL (string). The value type of this claim MUST be a string, where the value is a server-provided `c_nonce`. It MUST be present when the issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
+  * `nonce`: OPTIONAL (string). The value type of this claim MUST be a string, where the value is a server-provided `c_nonce`. It MUST be present when the Issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
 
 The Credential Issuer MUST validate that the JWT used as a proof is actually signed by a key identified in the JOSE Header through either `kid`, `jwk` or `x5c` element.
 
@@ -2708,7 +2708,7 @@ When a W3C Verifiable Presentation as defined by [@VC_DATA_2.0] or [@VC_DATA] se
   * `cryptosuite`: REQUIRED. If Credential Issuer metadata is provided, the value MUST match one of the entries in the `proof_signing_alg_values_supported` metadata parameter.
   * `proofPurpose`: REQUIRED. MUST be set to `authentication`.
   * `domain`: REQUIRED. MUST be set to the Credential Issuer Identifier.  
-  * `challenge`: REQUIRED when the Credential Issuer has provided a `c_nonce`. It MUST NOT be used otherwise. String, where the value is a server-provided `c_nonce`. It MUST be present when the issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
+  * `challenge`: REQUIRED when the Credential Issuer has a Nonce Endpoint as defined in (#nonce-endpoint). It MUST NOT be used otherwise. String, where the value is a server-provided `c_nonce`.
 
 The Credential Issuer MUST validate that the W3C Verifiable Presentation used as a proof is actually signed with a key in the possession of the Holder.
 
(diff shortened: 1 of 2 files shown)


paulbastian (approved):
I can't judge for DI, but the fix on jwt proof_type looks right

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -3033,7 +3033,7 @@ The JWT MUST contain the following elements:
   * `kid`: OPTIONAL. JOSE Header containing the key ID. If the Credential is to be bound to a DID, the `kid` refers to a DID URL which identifies a particular key in the DID Document that the Credential is to be bound to. It MUST NOT be present if `jwk` or `x5c` is present.
   * `jwk`: OPTIONAL. JOSE Header containing the key material the new Credential is to be bound to. It MUST NOT be present if `kid` or `x5c` is present.
   * `x5c`: OPTIONAL. JOSE Header containing at least one certificate where the first certificate contains the key that the Credential is to be bound to, additional certificates may also be present. It MUST NOT be present if `kid` or `jwk` is present.
-  * `key_attestation`: OPTIONAL. JOSE Header containing a key attestation as described in (#keyattestation). If the Credential Issuer provided a `c_nonce`, the `nonce` claim in the key attestation MUST be set to a server-provided `c_nonce`.
+  * `key_attestation`: OPTIONAL. JOSE Header containing a key attestation as described in (#keyattestation). If present, the `nonce` claim in the key attestation MUST be set to a server-provided `c_nonce`.
```
```suggestion
  * `key_attestation`: OPTIONAL. JOSE Header containing a key attestation as described in (#keyattestation). If the `nonce` claim is present in the key attestation, its value MUST be set to a server-provided `c_nonce`.
```

jogu (changes requested):
We still have the part here to solve: https://github.com/openid/OpenID4VCI/issues/676#issuecomment-3498762032 (or we should raise a new issue for that, I don't have a strong opinion which).

I think we should apply the change to 1.0 as well.

babisRoutis:
Hi @awoie 

Please consider updating the definition of `nonce` claim of `key_attestation` found in appendix E.

Currently the definition is

> nonce: OPTIONAL. String that represents a nonce provided by the Issuer to prove that a key attestation was freshly generated


IMO this could be update to include

> MUST be present if the attestation is used  with `attestation` Proof type, if credential issuer provides a nonce endpoint

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -3113,7 +3113,7 @@ When a W3C Verifiable Presentation as defined by [@VC_DATA_2.0] or [@VC_DATA] se
   * `cryptosuite`: REQUIRED. If Credential Issuer metadata is provided, the value MUST match one of the entries in the `proof_signing_alg_values_supported` metadata parameter.
   * `proofPurpose`: REQUIRED. MUST be set to `authentication`.
   * `domain`: REQUIRED. MUST be set to the Credential Issuer Identifier.  
-  * `challenge`: REQUIRED when the Credential Issuer has provided a `c_nonce`. It MUST NOT be used otherwise. String, where the value is a server-provided `c_nonce`. It MUST be present when the issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
+  * `challenge`: REQUIRED when the Credential Issuer has a Nonce Endpoint as defined in (#nonce-endpoint). It MUST NOT be used otherwise. String, where the value is a server-provided `c_nonce`.
```
GarethCOliver: It's somewhat inconsistent we say this is REQUIRED (but only when the Issuer has a Nonce Endpoint) while above for nonce we say it is OPTIONAL (but required when the Issuer has a Nonce Endpoint).

awoie:
```suggestion
  * `challenge`: OPTIONAL. The value type of this claim MUST be a string, where the value is a server-provided `c_nonce`. It MUST NOT be used otherwise. It MUST be present when the Issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
```

1.0/openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2708,7 +2708,7 @@ When a W3C Verifiable Presentation as defined by [@VC_DATA_2.0] or [@VC_DATA] se
   * `cryptosuite`: REQUIRED. If Credential Issuer metadata is provided, the value MUST match one of the entries in the `proof_signing_alg_values_supported` metadata parameter.
   * `proofPurpose`: REQUIRED. MUST be set to `authentication`.
   * `domain`: REQUIRED. MUST be set to the Credential Issuer Identifier.  
-  * `challenge`: REQUIRED when the Credential Issuer has provided a `c_nonce`. It MUST NOT be used otherwise. String, where the value is a server-provided `c_nonce`. It MUST be present when the issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
+  * `challenge`: REQUIRED when the Credential Issuer has a Nonce Endpoint as defined in (#nonce-endpoint). It MUST NOT be used otherwise. String, where the value is a server-provided `c_nonce`.
```
awoie:
```suggestion
  * `challenge`: OPTIONAL. The value type of this claim MUST be a string, where the value is a server-provided `c_nonce`. It MUST NOT be used otherwise. It MUST be present when the Issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
 ```

**Review by awoie — COMMENTED** [review]
Applied @GarethCOliver suggestion. Please review again.

As a general thought, we might use the opportunity to clean up the language a bit more and also include a table such as the one that @babisRoutis provided to make things more explicit.

File: 1.1/openid-4-verifiable-credential-issuance-1_1.md (hunk 2 above)
Comment (awoie) [reply 1.1/openid-4-verifiable-credential-issuance-1_1.md]: I applied your suggestion.

File: 1.0/openid-4-verifiable-credential-issuance-1_0.md [hunk 4]
```diff
@@ -2708,7 +2708,7 @@ When a W3C Verifiable Presentation as defined by [@VC_DATA_2.0] or [@VC_DATA] se
   * `cryptosuite`: REQUIRED. If Credential Issuer metadata is provided, the value MUST match one of the entries in the `proof_signing_alg_values_supported` metadata parameter.
   * `proofPurpose`: REQUIRED. MUST be set to `authentication`.
   * `domain`: REQUIRED. MUST be set to the Credential Issuer Identifier.  
-  * `challenge`: REQUIRED when the Credential Issuer has provided a `c_nonce`. It MUST NOT be used otherwise. String, where the value is a server-provided `c_nonce`. It MUST be present when the issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
+  * `challenge`: OPTIONAL. The value type of this claim MUST be a string, where the value is a server-provided `c_nonce`. It MUST NOT be used otherwise. It MUST be present when the Issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
```
awoie:
```suggestion
  * `challenge`: OPTIONAL. The value type of this claim MUST be a string, where the value is a server-provided `c_nonce`. It MUST be present when the Issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -3113,7 +3113,7 @@ When a W3C Verifiable Presentation as defined by [@VC_DATA_2.0] or [@VC_DATA] se
   * `cryptosuite`: REQUIRED. If Credential Issuer metadata is provided, the value MUST match one of the entries in the `proof_signing_alg_values_supported` metadata parameter.
   * `proofPurpose`: REQUIRED. MUST be set to `authentication`.
   * `domain`: REQUIRED. MUST be set to the Credential Issuer Identifier.  
-  * `challenge`: REQUIRED when the Credential Issuer has provided a `c_nonce`. It MUST NOT be used otherwise. String, where the value is a server-provided `c_nonce`. It MUST be present when the issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
+  * `challenge`: OPTIONAL. The value type of this claim MUST be a string, where the value is a server-provided `c_nonce`. It MUST NOT be used otherwise. It MUST be present when the Issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
```
awoie:
```suggestion
  * `challenge`: OPTIONAL. The value type of this claim MUST be a string, where the value is a server-provided `c_nonce`. It MUST be present when the Issuer has a Nonce Endpoint as defined in (#nonce-endpoint).
```

andprian:
@awoie the changes you made are in multiple places and describe how the wallet is expected to behave. I think we also need this kind on explanation in [Clause 7. Nonce Endpoint](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-nonce-endpoint ). It would be useful for the Issuer's side to have the big picture here. If he provides a nonce endpoint, explain how this is expected to work, when it is mandatory to use this endpoint and when not.

Sakurann:
discussed in the WG call:
- @andprian please open an issue and we can do another PR, but let's merge this PR first
- @c2bo @jogu @babisRoutis @GarethCOliver can you please re-review/approve?

Sakurann (approved):

c2bo (approved):

fkj (approved):

Sakurann:
@jogu put request for changes which i think have been addressed. please re-review:
>> We still have the part here to solve: https://github.com/openid/OpenID4VCI/issues/676#issuecomment-3498762032 (or we should raise a new issue for that, I don't have a strong opinion which).
>
> we now have the following text: "In some environments, a `nonce` is unnecessary because the key material in the `key_attestation` already provides sufficient entropy and freshness. Omitting the `nonce` also enables pre-generation of attestations prior to interacting with a specific Issuer. If the Issuer returns an `invalid_nonce` error, this can be interpreted as an indication that the Issuer expects a `nonce` claim in the `key_attestation`."

>> I think we should apply the change to 1.0 as well.
>
> has been applied

1.0/openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -2641,14 +2641,14 @@ The JWT MUST contain the following elements:
...
-  * `key_attestation`: OPTIONAL. JOSE Header containing a key attestation as described in (#keyattestation). If the Credential Issuer provided a `c_nonce`, the `nonce` claim in the key attestation MUST be set to a server-provided `c_nonce`.
+  * `key_attestation`: OPTIONAL. JOSE Header containing a key attestation as described in (#keyattestation). If the `nonce` claim is present in the key attestation, its value MUST be set to a server-provided `c_nonce` from the Nonce Endpoint as defined in (#nonce-endpoint). Note that including a `nonce` claim is left to the Wallet. In some environments, a `nonce` is unnecessary because the key material in the `key_attestation` already provides sufficient entropy and freshness. Omitting the `nonce` also enables pre-generation of attestations prior to interacting with a specific Issuer. If the Issuer returns an `invalid_nonce` error, this can be interpreted as an indication that the Issuer expects a `nonce` claim in the `key_attestation`.
```
andprian, line 2644: Initially there was a suggestion to have a parameter `require_nonce_in_key_attestation_in_jwt_proof` and I believe that this is necessary to have, otherwise the wallet cannot know what the issuer expects without trial and error.

andprian, line 2644: `this can be interpreted as an indication` -> can it be interpreted in another way as well?

andprian, line 2644: Considering that the issuer can still enforce the presence of the nonce in the key_attestation, it is not actually true that including a nonce claim is left to the Wallet. Saying this is left to the wallet makes one believe that the request will succeed in both cases.

brentzundel:
@jogu can you re-review?

brentzundel:
@jogu still waiting for re-review.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -3209,14 +3209,14 @@ The JWT MUST contain the following elements:
...
+  * `key_attestation`: OPTIONAL. JOSE Header containing a key attestation as described in (#keyattestation). If the `nonce` claim is present in the key attestation, its value MUST be set to a server-provided `c_nonce` from the Nonce Endpoint as defined in (#nonce-endpoint). Note that including a `nonce` claim is left to the Wallet. In some environments, a `nonce` is unnecessary because the key material in the `key_attestation` already provides sufficient entropy and freshness. Omitting the `nonce` also enables pre-generation of attestations prior to interacting with a specific Issuer. If the Issuer returns an `invalid_nonce` error, this can be interpreted as an indication that the Issuer expects a `nonce` claim in the `key_attestation`.
   * `trust_chain`: OPTIONAL. JOSE Header containing an [@!OpenID.Federation] Trust Chain. This element MAY be used to convey key attestation, metadata, metadata policies, federation Trust Marks and any other information related to a specific federation, if available in the chain. When used for signature verification, the header parameter `kid` MUST be present.
```
mickrau, line 3213: I don’t like the use of **JOSE Header** in the parameter descriptions; the correct term would be **JOSE Header parameter**. However, that may be unnecessary, since the context already makes it clear that these are parameters of the JOSE Header. Nonetheless, we should state their concrete types explicitly—just as we do for the body.

For `key_attestation`, it is unclear whether the value must be encoded as a JSON string (JWT Compact Serialization) or whether it may also be provided as a JSON object (JWT JSON Serialization).

mickrau (changes requested):

brentzundel:
Discussion from WG: @jogu still needs to re-review, also @awoie please respond to review from @mickrau and @andprian

brentzundel:
Discussed today: @awoie will respond to review comments, @jogu will re-review

brentzundel:
Discussed today: @awoie will respond to review comments, @jogu will re-review

PR #686, opened 2026-01-15
Add invalid_tx_code error code to pre-authz code flow

### LINKED ISSUE #474: Define dedicated error code for tx code in pre-auth code flow
The specification defines several OID4VCI-specific `error` codes for the credential response. For improved user experience, it would be beneficial to introduce a dedicated `error` code to indicate when the `tx_code` provided in the token request is invalid or no longer valid. Currently, we only have `invalid_grant`.

### Rationale
Some implementations may want to distinguish between an invalid pre-authorization code and an invalid transaction code (`tx_code`). In the case of an invalid pre-authorization code, an `invalid_grant` error would be appropriate, prompting the wallet to stop asking the user for a `tx_code`. However, if the `tx_code` is incorrect, it would be preferable for the wallet to allow the user to try again, rather than forcing them to return to the issuer for a new pre-authorization code—a process that can be unnecessarily complex or time-consuming.

To support this distinction, a new `invalid_tx_code` error should be defined. This would enable the wallet to clearly identify and respond to an invalid or missing `tx_code`, improving usability and error handling. That said, we still expect some implementations to return `invalid_grant` from the token endpoint once a maximum number of failed attempts (with invalid/wrong `tx_code`) is reached and the pre-authz code was invalidated to protect against brute forcing the token endpoint.

### Proposal

Add a new `error` code to the Token Error Response for the Pre-Authorization Code Flow:

`invalid_tx_code`:
- *(-copied from `invalid_grant`-)* The Authorization Server expects a Transaction Code in the Pre-Authorized Code Flow but the Client provides the wrong Transaction Code.

Update existing definition for `invalid_grant` that talks about tx codes.

`invalid_grant`:
- The End-User provides the wrong Pre-Authorized Code or the Pre-Authorized Code has expired *(-new-)* **or is no longer valid**.
- *(-removed transaction code language-)*

Add security considerations on brute forcing the `tx_code` and how this can be mitigated, e.g., by invalidating the pre-authz code after max number of failed attempts exceeded.

awoie:
Fixes #474

### DIFF
--- a/1.1/openid-4-verifiable-credential-issuance-1_1.md
+++ b/1.1/openid-4-verifiable-credential-issuance-1_1.md
@@ -3474,3 +3474,4 @@ The technology described in this specification was made available from contribut
    * Add back Interactive Authorization Endpoint text that was removed from the 1.0 draft
    * add require_interactive_authorization_request to AS metadata
    * add interactive_authorization_endpoint to AS metadata section
+   * add invalid_tx_code to Pre-Authz Code Flow
--- a/1.1/openid-4-verifiable-credential-issuance-1_1.md
+++ b/1.1/openid-4-verifiable-credential-issuance-1_1.md
@@ -1162,12 +1162,15 @@ The following additional clarifications are provided for some of the error codes
 
 `invalid_grant`:
 
-- The Authorization Server expects a Transaction Code in the Pre-Authorized Code Flow but the Client provides the wrong Transaction Code.
-- The End-User provides the wrong Pre-Authorized Code or the Pre-Authorized Code has expired.
+- The End-User provides the wrong Pre-Authorized Code or the Pre-Authorized Code has expired or is no longer valid.
 
 `invalid_client`:
 
 - The Client tried to send a Token Request with a Pre-Authorized Code without a Client ID but the Authorization Server does not support anonymous access.
+
+`invalid_tx_code`:
+
+- The Authorization Server expects a Transaction Code in the Pre-Authorized Code Flow but the Client provides the wrong Transaction Code.
 
 Below is a non-normative example of a Token Error Response:
 
@@ -1907,6 +1910,16 @@ The Pre-Authorized Code Flow is vulnerable to the replay of the Pre-Authorized C
 An attacker might leverage the Credential issuance process and the End-User's trust in the Wallet to phish Transaction Codes sent out by a different service that grant the attacker access to services other than Credential issuance. The attacker could set up a Credential Issuer site and in parallel to the issuance request, trigger transmission of a Transaction Code to the End-User's phone from a service other than Credential issuance, e.g., from a payment service. The End-User would then be asked to enter this Transaction Code into the Wallet and since the Wallet sends this Transaction Code to the Token Endpoint of the Credential Issuer (the attacker), the attacker would get access to the Transaction Code, and access to that other service.
 
 In order to cope with that issue, the Wallet is RECOMMENDED to interact with trusted Credential Issuers only. In that case, the Wallet would not process a Credential Offer with an untrusted issuer URL. The Wallet MAY also show the End-User the endpoint of the Credential Issuer it will be sending the Transaction Code to and ask the End-User for confirmation.
+
+### Transaction Code Guessing
+
+When the Pre-Authorized Code Flow is used together with a Transaction Code (`tx_code`), the Transaction Code is typically short, low-entropy, and intended for one-time use. As a result, it may be susceptible to online guessing or brute-force attacks if an attacker can repeatedly submit Token Requests using the same Pre-Authorized Code.
+
+To mitigate this risk, the Authorization Server SHOULD limit the number of failed Transaction Code verification attempts associated with a Pre-Authorized Code or issuance transaction. Once a configurable maximum number of failed attempts is exceeded, the Authorization Server SHOULD invalidate the Pre-Authorized Code and reject further Token Requests for that transaction.
+
+Transaction Codes SHOULD be short-lived and SHOULD be treated as single-use. Upon successful verification, a Transaction Code SHOULD NOT be accepted again.
+
+When a valid Pre-Authorized Code is presented with an incorrect Transaction Code, the Authorization Server SHOULD return the `invalid_tx_code` error. If the Pre-Authorized Code has expired, has been invalidated (including due to too many failed attempts), or is otherwise no longer valid, the Authorization Server SHOULD return the `invalid_grant` error.
 
 ## Credential Lifecycle Management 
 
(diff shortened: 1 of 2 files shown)


1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1908,6 +1911,16 @@ An attacker might leverage the Credential issuance process and the End-User's tr
 
 In order to cope with that issue, the Wallet is RECOMMENDED to interact with trusted Credential Issuers only. In that case, the Wallet would not process a Credential Offer with an untrusted issuer URL. The Wallet MAY also show the End-User the endpoint of the Credential Issuer it will be sending the Transaction Code to and ask the End-User for confirmation.
 
+### Transaction Code Guessing
+
+When the Pre-Authorized Code Flow is used together with a Transaction Code (`tx_code`), the Transaction Code is typically short, low-entropy, and intended for one-time use. As a result, it may be susceptible to online guessing or brute-force attacks if an attacker can repeatedly submit Token Requests using the same Pre-Authorized Code.
+
+To mitigate this risk, the Authorization Server SHOULD limit the number of failed Transaction Code verification attempts associated with a Pre-Authorized Code or issuance transaction. Once a configurable maximum number of failed attempts is exceeded, the Authorization Server SHOULD invalidate the Pre-Authorized Code and reject further Token Requests for that transaction.
```
paulbastian, line 1938: Should this be a must?

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1908,6 +1911,16 @@ An attacker might leverage the Credential issuance process and the End-User's tr
...
+
+Transaction Codes SHOULD be short-lived and SHOULD be treated as single-use. Upon successful verification, a Transaction Code SHOULD NOT be accepted again.
```
paulbastian, line 1940: Again should this be a MUST?

paulbastian (approved):
This makes a lot of sense, please consider to strengthen the normative language.

1.1/openid-4-verifiable-credential-issuance-1_1.md
awoie: It is again a question of breaking and non-breaking. MUST would technically trigger a breaking change whereas SHOULD is just a recommendation although a strong one.

1.1/openid-4-verifiable-credential-issuance-1_1.md
awoie: It is again a question of breaking and non-breaking. MUST would technically trigger a breaking change whereas SHOULD is just a recommendation although a strong one. I would make it a MUST if we can @jogu @tplooker @Sakurann ?

1.1/openid-4-verifiable-credential-issuance-1_1.md
Sakurann:
```suggestion
To mitigate this risk, the Authorization Server SHOULD limit the number of failed Transaction Code verification attempts associated with a Pre-Authorized Code or issuance transaction. Once a configurable maximum number of failed attempts is exceeded, the Authorization Server MUST invalidate the Pre-Authorized Code and reject further Token Requests for that transaction.
```

personal opinion. logically, recommending AS to limit the number of attempts is a new optional feature in 1.1, so saying that if that is implemented, mandating how to do it seems ok to me.

Sakurann: WG discussion. this should be in errata

1.1/openid-4-verifiable-credential-issuance-1_1.md
Sakurann: WG discussion. SHOULD is preferred here because turning this into a must can break some 1.0 implementations

fkj: As also brought up by @paulbastian in the WG meeting yesterday, it would be beneficial to explain the circumstances where it is OK not to do it. In my opinion, there are no such circumstances except that you already have a system implementing 1.0 where you didn't do it that you cannot change for some reason.

jogu: Discussed on WG calls today:

If a credential offer involved something sent by post does the pre-authorized code or transaction codes need to be long lived?

awoie: there is already a requirement that the pre-authorized code must be short-lived without further defining what short-lived means. i would expect the transaction codes to be aligned with this, or even shorter lived.

we discussed this in the wg last week, and nobody could bring up a case where the transaction code should be long-lived.

1.1/openid-4-verifiable-credential-issuance-1_1.md
awoie: Added to errata.

fkj (approved):
I was convinced during the WG meeting that the transaction code lifetime issue is "superseded" by the pre-authorized code lifetime already being "short-lived", whatever that means for the specific application. So I'm OK with just saying transaction codes SHOULD also be short-lived.

c2bo (approved):

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1162,13 +1162,16 @@ The following additional clarifications are provided for some of the error codes
...
 
-- The Authorization Server expects a Transaction Code in the Pre-Authorized Code Flow but the Client provides the wrong Transaction Code.
-- The End-User provides the wrong Pre-Authorized Code or the Pre-Authorized Code has expired.
+- The End-User provides the wrong Pre-Authorized Code or the Pre-Authorized Code has expired or is no longer valid.
 
 `invalid_client`:
 
 - The Client tried to send a Token Request with a Pre-Authorized Code without a Client ID but the Authorization Server does not support anonymous access.
 
+`invalid_tx_code`:
```
paulbastian, line 1176: I believe it makes sense to differentiate error codes, where the AS allows for a retry of sending Token Request with a new Transaction Code or whether this terminates the issuance session. Is the intention to allows retry for invalid_tx_code?

awoie: @paulbastian It is expected that `invalid_grant` would terminate the issuance session. Does this address your concern?

paulbastian: So is the intention to send invalid_tx_code as long as retry of transaction code is allowed and send invalid_grant if too many tries have been made due to brutr force prevention?

awoie: @paulbastian Yes, that is the idea. Do you have any concerns?

paulbastian: No, I would just phrase it more explicitly

Sakurann:
WG call
@paulbastian raised a point that this error code can be also used when correct tx_code was entered too many times. PR might follow.

PR #695, opened 2026-01-27
Use expected URL instead of expected origins for IAE flow

### LINKED ISSUE #672: Type definition of `expected_origins` clashes with IAE endpoint URL
For the IAE flow, we are using the IAE endpoint URL for the `expected_origins` for signed requests. The definition of `expected_origins` is that it contains Origins, not arbitrary URLs. OID4VP implementations might have issues with validating the value if it contains a request URL but an Origin value is expected.

Potential solutions:
1. (clean but potentially duplication) use a different field than `expected_origins`, e.g., `expected_urls`. This approach might need to require more spec changes and duplication of some parts of OID4VP in OID4VCI.
2. (clean but possibly same security concerns as 3.) use derived origin for `expected_origins` as proposed in this PR  https://github.com/openid/OpenID4VCI/pull/602. This approach loses some precision which might have an impact on security.
3. (ugly but quick) Just say that in this case `expected_origins` can contain arbitrary URLs including path and query string. Some implementations might cut the non-Origin parts of the URL if they share code paths with DC API handling. Not sure if this could lead to some form of mix-up attack if a host (or Origin) is serving multiple issuer services.

### LINKED ISSUE #698: IAE Binding for SD-JWT VC is inconsistent with other definitions
The SD-JWT VC IAE binding defines:
> the aud claim in the Key Binding JWT MUST be set to the derived Origin

The W3C IAE binding defines:
> the aud claim value MUST be set to the Interactive Authorization Endpoint

I think this should all use the IAE endpoint instead of the derived origin If I understood recent changes correctly?

awoie:
Fixes #672
Fixes #698

Need feedback on whether `expected_url` should be plural/ an array. Currently, it is a single value since there is only one IAE endpoint and follow-up requests have to use the IAE endpoint and not a dynamic URL provided by the interactive authorization response that requires interaction. If this should change any reasons (e.g. reusability), I would need to update the PR.

@GarethCOliver if this should be a resuable mechanism, do we need to change the binding the IAE endpoint sections? Currently, they are tightly coupled to IAE by using prefixes etc.


1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -3609,3 +3615,4 @@ The technology described in this specification was made available from contribut
    * use derived origin for `expected_origins` in IAE flow
```
TimoGlastra: Should this be removed, since the `expected_url` overwrites this change right? 

But similar things could be said about `iar-post` to `iae_post`. IAR was not there in 1.0, so it's a change before draft 1 was published, and if you're just looking at draft 1 compared to 1.0 it doesn't make a lot of sense.

awoie: That is true. I will discuss with the chairs and if can be removed, they should be removed.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -884,7 +885,7 @@ The exact architecture and the deployment of the Issuer's OpenID4VP Verifier is
 
 When processing the request the following logic applies:
 
-  1. If `expected_origins` is present, the Wallet MUST ensure that `expected_origins` contains the derived Origin as defined above.
+  1. If `expected_url` is present, the Wallet MUST compare the value in this parameter to the URL of the follow-up request to detect replay of the request from a malicious Verifier. If the URL does not match value of the `expected_url`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```
TimoGlastra: From reading this text, I'm not really sure what a wallet should do.

Before this change you were required to derive the origin from the IAE endpoint and compared this to the `expected_origins`, correct?

After this change you take the exact IAE endpoint and compare that against the `expected_url`? The text above mentions "string that contains the endpoint of the Verifier that is returning the request" which is a bit vague to me, and different from what was done before. Shouldn't it be the Issuer returning the request? And then the OID4VP request has a link to the issuer url (based on `expected_url`).

- Is it correct Verifier should be Issuer?
- Is it possible to define "endpoint of the Verifier that is returning the request" more clearly (is it the IAE endpoint?)

awoie: IMO, in the IAE case, the IAE endpoint will always have to mach the `expected_url` parameter.

> From reading this text, I'm not really sure what a wallet should do.
> 
> Before this change you were required to derive the origin from the IAE endpoint and compared this to the `expected_origins`, correct?

Correct but this has security implications. Also note that the previous IAE text was a draft anyways.

> 
> After this change you take the exact IAE endpoint and compare that against the `expected_url`? The text above mentions "string that contains the endpoint of the Verifier that is returning the request" which is a bit vague to me, and different from what was done before. Shouldn't it be the Issuer returning the request? And then the OID4VP request has a link to the issuer url (based on `expected_url`).
> 
> * Is it correct Verifier should be Issuer?

We can call it issuer instead. I wanted the text more generic because some folks noted they wanted that feature to be more general and applicable to any endpoints the wallet is interaction with.

> * Is it possible to define "endpoint of the Verifier that is returning the request" more clearly (is it the IAE endpoint?)

Yes, IMO, it will always be the IAE endopint.

Le's discuss with @GarethCOliver what he prefers.

TimoGlastra: Keeping it generic in design is okay, but it would be good to describe somewhere explicitly that in the case of IAE it MUST be the IAE endpoint.

awoie: I believe we agreed that we will create a dedicated issue for generic design. I will keep the parameter name generic though

1.1/openid-4-verifiable-credential-issuance-1_1.md
awoie: I removed the history entry.

1.1/openid-4-verifiable-credential-issuance-1_1.md
awoie: However, I made the language future-proof in case we forget to update this section once we made the section more generic.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -824,7 +824,8 @@ Additional keys are defined based on the type of interaction, as shown next.
 If `type` is set to `openid4vp_presentation`, as shown in the following example, the response MUST further include an `openid4vp_request` parameter containing an OpenID4VP Authorization Request. The contents of the request is the same as for requests passed to the Digital Credentials API (see Appendix A.2 and Appendix A.3 of [@!OpenID4VP]), except as follows:
 
 * The `response_mode` MUST be either `iae_post` for unencrypted responses or `iae_post.jwt` for encrypted responses. These modes are used to indicate to the Wallet to return the response back to the same Interactive Authorization Endpoint.
-* If `expected_origins` is present, it MUST contain only the derived Origin of the Interactive Authorization Endpoint as defined in Section 4 in [@RFC6454]. For example, the derived Origin from `https://example.com/iae` is `https://example.com`.
+* The `expected_origins` parameter MUST NOT be present.
+* When signed requests defined in A.3.2 of [@!OpenID4VP] are used, an addtional `expected_url` parameter MUST be present. The `expected_url` parameter is a non-empty string that contains the endpoint of the Verifier that is returning the request.
```
awoie:
```suggestion
* When signed requests defined in A.3.2 of [@!OpenID4VP] are used, an addtional `expected_url` parameter MUST be present. The `expected_url` parameter is a non-empty string that contains the endpoint of the Verifier that is returning the request. In the `openid4vp_presentation` interaction type, `expected_url` MUST be set to the Interactive Authorization Endpoint.
```

awoie:
This PR needs more reviews. 

Could you please review again @TimoGlastra

1.1/openid-4-verifiable-credential-issuance-1_1.md
TimoGlastra:
```suggestion
  1. If `expected_url` is present, the Wallet MUST compare the value in this parameter to the URL of the follow-up request to detect replay of the request from a malicious Verifier. If the URL does not match the value of the `expected_url`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
TimoGlastra: I think this should also be updated to use `expected_url`?

```suggestion
 1. The Wallet is able to detect that a request is not presented by the party that initiated the Interactive Authorization Request. In the case of the (#iae-require-presentation) interaction with a signed Presentation request, this is achieved by the Wallet verifying the `expected_url` parameter in the request, which contains the Interactive Authorization Endpoint that initiated the request.
```

TimoGlastra: To address #698, this should be updated to:

```suggestion
To bind the Interactive Authorization Endpoint to a Verifiable Presentation using the Credential Format defined in this section, the `aud` claim in the Key Binding JWT MUST be set to the MUST be set to the Interactive Authorization Endpoint, prefixed with `iae:` (e.g., `iae:https://example.com/iae`).
```

TimoGlastra (commented):
I think it's missing some required changes to fully remove all references to "Derived origin"

1.1/openid-4-verifiable-credential-issuance-1_1.md
awoie: For some reason I didn't commit all changes. I committed and pushed them now.

awoie: For some reason I didn't commit all changes. I committed and pushed them now.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2851,7 +2852,7 @@ The following is a non-normative example of a Credential Response containing a C
 
 ### Interactive Authorization Endpoint Binding {#iae-binding-sd-jwt-vc}
 
-To bind the Interactive Authorization Endpoint to a Verifiable Presentation using the Credential Format defined in this section, the `aud` claim in the Key Binding JWT MUST be set to the derived Origin (as defined in (#iae-require-presentation)) of the Interactive Authorization Endpoint, prefixed with `iae:` (e.g., `iae:https://example.com`).
+To bind the Interactive Authorization Endpoint to a Verifiable Presentation using the Credential Format defined in this section, the `aud` claim in the Key Binding JWT MUST be set to the  Interactive Authorization Endpoint, prefixed with `iae:` (e.g., `iae:https://example.com`).
```
TimoGlastra:
```suggestion
To bind the Interactive Authorization Endpoint to a Verifiable Presentation using the Credential Format defined in this section, the `aud` claim in the Key Binding JWT MUST be set to the Interactive Authorization Endpoint, prefixed with `iae:` (e.g., `iae:https://example.com`).
```

TimoGlastra (approved):

c2bo (commented):
I found the description/terminology a bit confusing: request vs Interactive  Authorization Response in our terminology unless i misunderstood something. This is the current text:

> The `expected_url` parameter is a non-empty string that contains the endpoint of the Verifier that is returning the request. In the `openid4vp_presentation` interaction type, `expected_url` MUST be set to the Interactive Authorization Endpoint.

> If `expected_url` is present, the Wallet MUST compare the value in this parameter to the URL of the follow-up request to detect replay of the request from a malicious Verifier.

> this is achieved by the Wallet verifying the `expected_url` parameter in the request, which contains the Interactive Authorization Endpoint that initiated the request.

Given the terminology, wouldn't something like this be more fitting?

> the `expected_url` parameter is a non-empty string that contains the endpoint of the Authorization Server that is returning the Interactive Authorization Response.

> If `expected_url` is present, the Wallet MUST compare the value in this parameter to the URL of the follow-up Interactive Authorization Request to detect replay of the request from a malicious Authorization Server/Verifier.

> this is achieved by the Wallet verifying the `expected_url` parameter in the Interactive Authorization Response, which contains the Interactive Authorization Endpoint that initiated the Interactive Authorization Response.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -824,7 +824,8 @@ Additional keys are defined based on the type of interaction, as shown next.
...
 
 * The `response_mode` MUST be either `iae_post` for unencrypted responses or `iae_post.jwt` for encrypted responses. These modes are used to indicate to the Wallet to return the response back to the same Interactive Authorization Endpoint.
-* If `expected_origins` is present, it MUST contain only the derived Origin of the Interactive Authorization Endpoint as defined in Section 4 in [@RFC6454]. For example, the derived Origin from `https://example.com/iae` is `https://example.com`.
+* The `expected_origins` parameter MUST NOT be present.
+* When signed requests defined in A.3.2 of [@!OpenID4VP] are used, an addtional `expected_url` parameter MUST be present. The `expected_url` parameter is a non-empty string that contains the endpoint of the Verifier that is returning the request. In the `openid4vp_presentation` interaction type, `expected_url` MUST be set to the Interactive Authorization Endpoint.
```
awoie: @c2bo is this better?
```suggestion
* When signed requests defined in A.3.2 of [@!OpenID4VP] are used, an additional `expected_url` parameter MUST be present. The `expected_url` parameter is a non-empty string that contains the endpoint of the Authorization Server that is returning the Interactive Authorization Response. In the `openid4vp_presentation` interaction type, `expected_url` MUST be set to the Interactive Authorization Endpoint.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -884,7 +885,7 @@ The exact architecture and the deployment of the Issuer's OpenID4VP Verifier is
 
 When processing the request the following logic applies:
 
-  1. If `expected_origins` is present, the Wallet MUST ensure that `expected_origins` contains the derived Origin as defined above.
+  1. If `expected_url` is present, the Wallet MUST compare the value in this parameter to the URL of the follow-up request to detect replay of the request from a malicious Verifier. If the URL does not match the value of the `expected_url`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```
awoie: @c2bo is this better?
```suggestion
  1. If `expected_url` is present, the Wallet MUST compare the value in this parameter to the URL of the follow-up Interactive Authorization Request to detect replay of the request from a malicious Authorization Server/Verifier. If the URL does not match the value of the `expected_url`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1006,7 +1007,7 @@ This may lead to the malicious Authorization Server gaining access to Credential
 
 Custom extensions ((#iae-custom-extensions)) MUST ensure that this attack is prevented by ensuring one or both of the following:
 
- 1. The Wallet is able to detect that a request is not presented by the party that initiated the Interactive Authorization Request. In the case of the (#iae-require-presentation) interaction with a signed Presentation request, this is achieved by the Wallet verifying the `expected_origins` parameter in the request, which contains the derived Origin of the Interactive Authorization Endpoint that initiated the request.
+ 1. The Wallet is able to detect that a request is not presented by the party that initiated the Interactive Authorization Request. In the case of the (#iae-require-presentation) interaction with a signed Presentation request, this is achieved by the Wallet verifying the `expected_url` parameter in the request, which contains the Interactive Authorization Endpoint that initiated the request.
```
awoie: @c2bo is this better?
```suggestion
 1. The Wallet is able to detect that a request is not presented by the party that initiated the Interactive Authorization Request. In the case of the (#iae-require-presentation) interaction with a signed Presentation request, this is achieved by the Wallet verifying the `expected_url` parameter in the Interactive Authorization Response, which contains the Interactive Authorization Endpoint that initiated the Interactive Authorization Response.
```

Sakurann:
@c2bo can you please respond to @awoie 's suggestions?

c2bo:
> @c2bo can you please respond to @awoie 's suggestions?

I think those changes help clarify / reduce the confusion between request & IA response, thanks @awoie!

fkj (commented):
Looks good to me with the new suggestions.

c2bo (approved):
looks good to me with the proposed changes

brentzundel:
@awoie if you accept @c2bo 's changes, then I believe we can merge this PR.

brentzundel:
@awoie, please also address merge conflicts.

mickrau:
@awoie @brentzundel 
i solved the merge conflicts in [mickrau/fix-672](/openid/OpenID4VCI/tree/mickrau/fix-672)  

Feel free to merge it into this PR

Update: After speaking with @paulbastian, I went ahead and merged it myself.

mickrau (changes requested):
I fixed the conflicts, but i'm not really satisfied that we're introducing a new VP related variable in this spec and gaining so little from it.
Correct me if I'm wrong, but we get a little more precision (uri instead of origin) but this only addresses potential privacy issues when:

different AS instances use the same origin and one instance is evil
the Authorization Response is not encrypted.
I would prefer to keep it simple and recommend to use response encryption (in HAIP)

brentzundel:
discussion in WG: we need to re-review this PR to be sure it still makes sense after building IAE on top of the first-party apps draft.

brentzundel:
@fkj will review

paulbastian (changes requested):
I don't think this is necessary. We came to the conclusion in redirect_uri that origins are sufficient. If you are concerned with multi-tenant environments, please use subdomains

brentzundel:
No comments since marked pending close, closing

Issue #194, opened 2023-12-31
Update to W3C VC DMv2


David-Chadwick:
The current specification is based on the W3C VC DM v1. 

The W3C VC DM v2 is about to be published as a Candidate Recommendation. It contains a number of differences to the v1 DM. Importantly all VC DM v2 VCs must have an `@context` property. More importantly, securing them using SD-JWT is indicated in a different way to securing a non-W3C VC with SD-JWT. Specifically the typ is set to vc+ld+json+sd-jwt (in the latest draft version of the specification), rather than vc+sd-jwt which is used by the SD-JWT VC draft.

Therefore a number of changes will be needed to the OpenID4VCI specification if it is to cater for the W3C VC DM v2.

This issue is a placeholder to indicate that some changes will be needed in order to accommodate the W3C VC DM v2.

selfissued:
The spec actually refers to both VCDM 1.1 and VCDM 2.0 at present.  I agree that we will probably want to do a harmonization pass once VCDM 2.0 is a Candidate Recommendation.

babisRoutis:
> 
>  Importantly all VC DM v2 VCs must have an `@context` property. More importantly, securing them using SD-JWT is indicated in a different way to securing a non-W3C VC with SD-JWT. Specifically the typ is set to vc+ld+json+sd-jwt (in the latest draft version of the specification), rather than vc+sd-jwt which is used by the SD-JWT VC draft.
> 

@David-Chadwick 
Given that SD-JWT-VC is a credential format, not related to W3C VC DM (1.1 or 2.0) and given that VCI has already different appendixes for them, I don't understand your point.
Why would somebody confuse W3C VC 2.0 that use SD-JWT, with SD-JWT-VC?

David-Chadwick:
They already do confuse the two!

Sakurann:
> Why would somebody confuse W3C VC 2.0 that use SD-JWT, with SD-JWT-VC?

+1. the payload is very different.

Sakurann:
we'll probably need to discuss if we want to replace VCDM v1.1 credential format profile with VCDM v2.0 (pretty big change) or add a new credential format profile for VCDM v2.0. technically, the former is probably more appropriate because VCDM 2.0 is supposed to replace v1.1, but I think it will be more confusing because it's not like all v1.1 implementations will upgrade to v2.0 at the same time, so I do see the benefit of keeping VCDM v1.1 format profile

David-Chadwick:
[...]

PR #699, opened 2026-01-29
add credential format profile for vcdm 2 with jose

TimoGlastra:
This partially addresses #194, but only focuses on the VCDM 2.0 with JOSE (vc+jwt and vc+sd-jwt), but doesn't include the JSON-LD-based or CBOR based variants.

It might make sense to add them all at once, but we have only implemented / experimented with the JOSE-based variants.

Since there was already an issue where contribution was asked, I decided to create a PR directly instead of opening the discussion in the issue again.

Based on comments from @jogu in #194 I added a new profile rather than extending the existing VCDM profile.

### DIFF
--- a/1.1/examples/authorization_details_vc_jwt.json
+++ b/1.1/examples/authorization_details_vc_jwt.json
@@ -1,0 +1,11 @@
+[
+  {
+    "type": "openid_credential",
+    "credential_configuration_id": "UniversityDegreeCredential",
+    "claims": [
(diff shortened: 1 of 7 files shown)


bc-pi (changes requested):
Image generally agree here

brentzundel:
Discussed in WG meeting today. Decision was to close, continue work in the issue, and work on this in a version if the spec beyond 1.1

brentzundel:
no objections since being marked `pending close`. Closing.

PR #706, opened 2026-02-06
Make auth_session optional if other binding mechanism exists.

### LINKED ISSUE #690: Stuttgart 3: align example and text on auth_session in the response
3. Section 6.2.1 requires an auth_session parameter, but the example in Section 6.2.1.2 (Redirect to Web) does not contain an auth_session value. Not returning the auth_session value seems reasonable as the actual authorization endpoint might return an auth_session value that the client can use for a follow-up request to the IAE.

fkj:
The auth_session parameter is not needed in the all Interaction Required Responses since some interaction types already include a way to associate the next request with the ongoing auth sequence.

This makes the requirement for auth_session conditional on there not being some other mechanism for binding the request to the auth sequence.
It also adds a description of why the auth_session is not necessary in the redirect_to_web case and the custom example.

This is a suggestion to fix #690, but we could also go other ways.

### DIFF
--- a/1.1/openid-4-verifiable-credential-issuance-1_1.md
+++ b/1.1/openid-4-verifiable-credential-issuance-1_1.md
@@ -808,9 +808,13 @@ Depending on this assessment, the response from the Interactive Authorization En
 ### Interaction Required Response {#iae-interaction-required-response}
 
 By setting `status` to `require_interaction` in the response, the Authorization Server requests an additional user interaction.
-In this case, the following keys MUST be present in the response as well:
+In this case, the following key MUST be present in the response as well:
 
 * `type`: REQUIRED. String indicating which type of interaction is required, as defined below. The Authorization Server MUST NOT set this to a value that was not included in the `interaction_types_supported` parameter sent by the Wallet.
+
+The Authorization Server MUST provide a mechanism to associate the next request by this Wallet with the ongoing authorization request sequence.
+If no other mechanism to associate the next request by this Wallet with the ongoing authorization request sequence is defined by the type of interaction, the following key MUST be present in the response as well:
+
 * `auth_session`: REQUIRED. String containing a value that allows the Authorization Server to associate subsequent requests by this Wallet with the ongoing authorization request sequence. Wallets SHOULD treat this value as an opaque value. The value returned MUST be distinct for each interactive authorization response.
 
 The Wallet MUST include the most recently received `auth_session` in follow-up requests to the Interactive Authorization Endpoint.
@@ -825,6 +829,8 @@ If `type` is set to `openid4vp_presentation`, as shown in the following example,
 
 * The `response_mode` MUST be either `iae-post` for unencrypted responses or `iae-post.jwt` for encrypted responses. These modes are used to indicate to the Wallet to return the response back to the same Interactive Authorization Endpoint.
 * If `expected_origins` is present, it MUST contain only the derived Origin of the Interactive Authorization Endpoint as defined in Section 4 in [@RFC6454]. For example, the derived Origin from `https://example.com/iae` is `https://example.com`.
+
+The response MUST include the key `auth_session` to associate the next request by this Wallet with the ongoing authorization request sequence.
 
 The following is a non-normative example of an unsigned Authorization Request:
 
@@ -939,6 +945,8 @@ In this case, the Authorization server MUST include the key `request_uri` in the
 The Wallet MUST use the `request_uri` value to build an Authorization Request as defined in Section 4 of [@!RFC9126] and complete the rest of the authorization process as defined there.
 The Authorization Server MAY include the `expires_in` key as defined in [@!RFC9126].
 
+Since the `request_uri` allows the Authorization Server to associate the Authorization Request with the ongoing authorization request sequence, no `auth_session` is needed.
+
 Non-normative Example:
 
 ```
@@ -975,6 +983,7 @@ It is RECOMMENDED to use this extension point instead of modifying the OAuth pro
 See (#iae-security) for additional security considerations.
 
 In the following non-normative example, this extension point is used to read the Betelgeuse Intergalactic ID card through an NFC interface in the Wallet. A token called `biic_token` is used to start the process.
+It is assumed that the `biic_token` is used by the Authorization Server to associate the next request by this Wallet with the ongoing authorization request sequence, and no `auth_session` is thus needed.
 
 ```
 HTTP/1.1 200 OK
--- a/1.1/openid-4-verifiable-credential-issuance-1_1.md
+++ b/1.1/openid-4-verifiable-credential-issuance-1_1.md
@@ -945,7 +945,7 @@ In this case, the Authorization server MUST include the key `request_uri` in the
 The Wallet MUST use the `request_uri` value to build an Authorization Request as defined in Section 4 of [@!RFC9126] and complete the rest of the authorization process as defined there.
 The Authorization Server MAY include the `expires_in` key as defined in [@!RFC9126].
 
-Since the `request_uri` allows the Authorization Server to associate the Authorization Request with the ongoing authorization request sequence, no `auth_session` is needed.
+Since the `request_uri` allows the Authorization Server to associate the Authorization Request with the ongoing authorization request sequence, the Authorization Server MUST omit `auth_session` parameter in the response. The `auth_session` will be returned in the redirect back to the Wallet if required.
 
 Non-normative Example:
 
--- a/1.1/openid-4-verifiable-credential-issuance-1_1.md
+++ b/1.1/openid-4-verifiable-credential-issuance-1_1.md
@@ -813,9 +813,14 @@ In this case, the following key MUST be present in the response as well:
 * `type`: REQUIRED. String indicating which type of interaction is required, as defined below. The Authorization Server MUST NOT set this to a value that was not included in the `interaction_types_supported` parameter sent by the Wallet.
 
 The Authorization Server MUST provide a mechanism to associate the next request by this Wallet with the ongoing authorization request sequence.
-If no other mechanism to associate the next request by this Wallet with the ongoing authorization request sequence is defined by the type of interaction, the following key MUST be present in the response as well:
+The following key MAY be present in the response to provide such a mechanism:
 
 * `auth_session`: REQUIRED. String containing a value that allows the Authorization Server to associate subsequent requests by this Wallet with the ongoing authorization request sequence. Wallets SHOULD treat this value as an opaque value. The value returned MUST be distinct for each interactive authorization response.
+
+A definition of a custom type of interaction MUST include exactly one of the following:
+
+1. A normative requirement that the `auth_session` key MUST be included in the Interaction Required Response.
+2. A definition of a mechanism to associate the next request by the Wallet with the ongoing authorization request sequence.
 
 The Wallet MUST include the most recently received `auth_session` in follow-up requests to the Interactive Authorization Endpoint.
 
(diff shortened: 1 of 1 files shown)


Sakurann:
i like the direction. I think we need to specify behavior for the wallet and explicitly say that the wallet must include auth_session in the subsequent request if it was included by the issuer in the IAE response. I didn't see us say something like that already, but i might have missed.

this means that it is completely up to the issuer to bind these sessions and these session binding is opaque to the wallet. i assume that's ok and does not open any security gaps as we have discussed this in the WG calls number of times, but wanted to note it one last time.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -939,6 +945,8 @@ In this case, the Authorization server MUST include the key `request_uri` in the
 The Wallet MUST use the `request_uri` value to build an Authorization Request as defined in Section 4 of [@!RFC9126] and complete the rest of the authorization process as defined there.
 The Authorization Server MAY include the `expires_in` key as defined in [@!RFC9126].
 
+Since the `request_uri` allows the Authorization Server to associate the Authorization Request with the ongoing authorization request sequence, no `auth_session` is needed.
```
Vanderkast:
```suggestion
Since the `request_uri` allows the Authorization Server to associate the Authorization Request with the ongoing authorization request sequence, the Authorization Server MAY omit `auth_session` parameter in the response.
```

Note: here I assume that it is AS to decide how to associate subsequent IAE requests to the ongoing auth session.

jogu: Discussed on today's WG:

We definitely don't need auth_session here, and the Wallet wouldn't ever use it if is here. "MAY omit" hence sounds wrong, suggestions welcome. "MUST omit" might work. 

Joseph's suggestion:

> auth_session will be returned in the redirect if required so is not passed here.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -808,9 +808,13 @@ Depending on this assessment, the response from the Interactive Authorization En
 ### Interaction Required Response {#iae-interaction-required-response}
 
 By setting `status` to `require_interaction` in the response, the Authorization Server requests an additional user interaction.
-In this case, the following keys MUST be present in the response as well:
+In this case, the following key MUST be present in the response as well:
 
 * `type`: REQUIRED. String indicating which type of interaction is required, as defined below. The Authorization Server MUST NOT set this to a value that was not included in the `interaction_types_supported` parameter sent by the Wallet.
+
+The Authorization Server MUST provide a mechanism to associate the next request by this Wallet with the ongoing authorization request sequence.
+If no other mechanism to associate the next request by this Wallet with the ongoing authorization request sequence is defined by the type of interaction, the following key MUST be present in the response as well:
```
jogu: Discussed on today's WG call:

Could we be more explicit that each interaction type must state whether auth_session is included in the response or not?

1.1/openid-4-verifiable-credential-issuance-1_1.md
fkj: How about:
```suggestion
Since the `request_uri` allows the Authorization Server to associate the Authorization Request with the ongoing authorization request sequence, no `auth_session` is needed in this response. The `auth_session` will be returned in the redirect back to the Wallet if required.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
fkj: How about adding this text in the `iae-custom-extensions` section:

> A definition of a custom type of interaction MUST include exactly one of the following:
> 1. A normative requirement that the `auth_session` key MUST be included in the Interaction Required Response.
> 2. A definition of a mechanism to associate the next request by the Wallet with the ongoing authorization request sequence.

I know this is a little bit meta since it imposes requirements on the normative text of extension specs, but I guess that was your intention here @jogu?

1.1/openid-4-verifiable-credential-issuance-1_1.md
jogu: LGTM.

1.1/openid-4-verifiable-credential-issuance-1_1.md
jogu: Discussed on today's WG call - no objections to the approach @fkj suggests, let's try updating the PR and get some reviews.

1.1/openid-4-verifiable-credential-issuance-1_1.md
Vanderkast: I would suggest to be more explicit:

```suggestion
Since the `request_uri` allows the Authorization Server to associate the Authorization Request with the ongoing authorization request sequence, the Authorization Server MUST omit `auth_session` parameter in the response. The `auth_session` will be returned in the redirect back to the Wallet if required.
```

Otherwise, LGTM.

1.1/openid-4-verifiable-credential-issuance-1_1.md
Vanderkast: Should we normatively say what do the Wallet MUST (or MAY) do when it receives unexpected `auth_session` according to the interaction?

E.g. when AS adds `auth_session` in redirect_to_web interaction required response, the Wallet:
a) ignores that and goes with the redirect;
b) aborts auth process;

fkj: I think this should be up to each interaction type to define, but I don't have a strong opinion. Perhaps there is some interaction type where it makes sense to get either. I will update the PR to reflect what I think is the current consensus and we can go from there based on reviews.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -808,11 +808,20 @@ Depending on this assessment, the response from the Interactive Authorization En
...
 By setting `status` to `require_interaction` in the response, the Authorization Server requests an additional user interaction.
-In this case, the following keys MUST be present in the response as well:
+In this case, the following key MUST be present in the response as well:
 
 * `type`: REQUIRED. String indicating which type of interaction is required, as defined below. The Authorization Server MUST NOT set this to a value that was not included in the `interaction_types_supported` parameter sent by the Wallet.
+
+The Authorization Server MUST provide a mechanism to associate the next request by this Wallet with the ongoing authorization request sequence.
+The following key MAY be present in the response to provide such a mechanism:
+
 * `auth_session`: REQUIRED. String containing a value that allows the Authorization Server to associate subsequent requests by this Wallet with the ongoing authorization request sequence. Wallets SHOULD treat this value as an opaque value. The value returned MUST be distinct for each interactive authorization response.
```
Vanderkast:
```suggestion
* `auth_session`: REQUIRED under conditions defined by the interaction type. String containing a value that allows the Authorization Server to associate subsequent requests by this Wallet with the ongoing authorization request sequence. Wallets SHOULD treat this value as an opaque value. The value returned MUST be distinct for each interactive authorization response.
```

fkj: Another option:
```suggestion
* `auth_session`: REQUIRED if specified by the interaction type. String containing a value that allows the Authorization Server to associate subsequent requests by this Wallet with the ongoing authorization request sequence. Wallets SHOULD treat this value as an opaque value. The value returned MUST be distinct for each interactive authorization response.
```

fkj: @Vanderkast Would you be OK with going with my suggestion for this one or do you have a strong opinion on the exact wording?

Vanderkast: @fkj, your option is OK for me.

1.1/openid-4-verifiable-credential-issuance-1_1.md
Vanderkast: I remember we re-visited the following on one of the meetings, but I don't remember resolution.

Assume interaction type defines a mechanism to associate the next request by the Wallet with the ongoing authorization request sequence. Thus AS MUST omit `auth_session` in the IAE response.
1) Should interaction type explicitly define how the Wallet MUST/SHOULD act if AS sends `auth_session`? or
2) should we globally define that the Wallet MUST abort transaction? or
3) should we globally define that the Wallet MUST ignore `auth_session` and proceed with the defined mechanism?

Vanderkast (approved):
I have one [open question](https://github.com/openid/OpenID4VCI/pull/706#discussion_r2959159037) left.

Besides that, LGTM

Thank you, @fkj

1.1/openid-4-verifiable-credential-issuance-1_1.md
fkj: I'm not sure this actually was resolved. I think there are interoperability and UX pros and cons to each of the possibilities.

At the last WG meeting, I asked for anyone who is going to define their own interaction types to consider how the options might affect the interaction types they imagine.

GarethCOliver (approved):

brentzundel:
@fkj now that #736 has been merged, please take care of merge conflicts and otherwise prepare this PR for consideration by the WG.

fkj:
I have rebased this PR on top of the new main including #736, and I believe it is ready for review.

1.1/openid-4-verifiable-credential-issuance-1_1.md
mickrau:
```suggestion
```
remove `auth_session` in example

mickrau:
```suggestion
```
remove auth_session in example

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -783,15 +783,25 @@ The response to an Authorization Challenge Request is an HTTP message with the c
  3. an error as defined in Section 5.2.2 of [@!I-D.ietf-oauth-first-party-apps], including the additional error codes defined in (#ia-error-response).
 
 ### Interaction Required Response {#ia-interaction-required-response}
-
 By setting `error_code` to `insufficient_authorization` in the response with HTTP response code 401 `Unauthorized`, the Authorization Server requests an additional user interaction.
```
c2bo, line 788: Not sure why the empty line was deleted (should be there in markdown)

```suggestion

By setting `error_code` to `insufficient_authorization` in the response with HTTP response code 401 `Unauthorized`, the Authorization Server requests an additional user interaction.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
fkj: Something is weird with github so I couldn't apply this, but I removed it in 84ac45b.

fkj: Something is weird with github so I couldn't apply this, but I removed it in 84ac45b.

1.1/openid-4-verifiable-credential-issuance-1_1.md
fkj: Something is weird with github so I couldn't apply this, but I added it back in 84ac45b (and another missing one as well).

c2bo (approved):
We should probably add a small clarification on Wallet behaviour if `auth_session` is present but shouldn't be (as previously asked). I am fine with this being in the mechanisms, but that should be added then? Apart from that lgtm

brentzundel:
> We should probably add a small clarification on Wallet behaviour if `auth_session` is present but shouldn't be (as previously asked). I am fine with this being in the mechanisms, but that should be added then? Apart from that lgtm

From the discussion today: The two options seems to be ignoring `auth_session` in this case, or failing.  There is also the consideration of failing this step vs failing the whole session. Failing the whole session may be expensive in terms of time, energy, or actual money.

In general there seems to be support for ignoring.

@c2bo can you add text for this?

brentzundel:
or perhaps @fkj could add text, since he originated the PR?

fkj:
> We should probably add a small clarification on Wallet behaviour if `auth_session` is present but shouldn't be (as previously asked). I am fine with this being in the mechanisms, but that should be added then? Apart from that lgtm

I added text saying it MUST be ignored in b0781ae. Open for discussion/reviews.

brentzundel:
multiple positive review, requests for changes addressed, merging

PR #712, opened 2026-02-25
Add URNs for IAE types to prevent collisions

### LINKED ISSUE #685: Prevent collisions when people extend IAE interaction types
The "type" field defined here https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-1_1-wg-draft.html#section-6.2.1 we expect people to define extra values potentially custom to ecosystems.

We probably need to document an approach for avoiding collisions.

One approach would be an IANA registry (though OIDF specs cannot create IANA registries). Another approach would be saying that non-OIDF definitions should use a urn format that presents collisions, e.g. incorporating a reverse domain name.

c2bo:
Fixes #685 

- Adds requirement to use URNs for IAE types to prevent naming collisions
- Request registration for the "openid" URN namespace
- Define valid URNs for the 2 IAE types defined in this document

I am not that confident on the proposed IANA considerations section to request registration of the namespace `urn:openid` (since it encompasses all of OpenID). I made some proposals that I believe are reasonable such as WGs managing their own sub-namespace, but please review that section properly and provide feedback.

### DIFF
--- a/1.1/openid-4-verifiable-credential-issuance-1_1.md
+++ b/1.1/openid-4-verifiable-credential-issuance-1_1.md
@@ -708,12 +708,12 @@ Note: In case a Wallet Attestation is required by the Authorization Server, it h
 
 The initial request to the Interactive Authorization Endpoint is formed and sent in the same way as PAR request as defined in Section 2.1 of [@!RFC9126]. The contents of the request are the same as in a regular Authorization Request as defined in (#credential-authz-request), with the following addition:
 
-`interaction_types_supported`: REQUIRED. Comma-separated list of strings indicating the types of interactions that the Wallet supports. The order of the values is not significant. The following values are defined by this specification:
+`interaction_types_supported`: REQUIRED. Comma-separated list of strings indicating the types of interactions that the Wallet supports. The order of the values is not significant. Values MUST be valid URNs. The following values are defined by this specification:
 
 * `urn:openid:dcp:iae:openid4vp_presentation`: Indicates that the Wallet supports an OpenID4VP Presentation interaction, as defined in (#iae-require-presentation).
 * `urn:openid:dcp:iae:redirect_to_web`: Indicates that the Wallet supports a redirect to a web-based interaction, as defined in (#iae-redirect-to-web).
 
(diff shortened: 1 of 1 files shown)


c2bo:
@selfissued could i kindly ask you for a review especially on the IANA registration request for the "openid" URN namespace?

fkj (approved):
As we discussed in the WG call yesterday, we will probably need an "official" OIDF opinion on whether we should register a URN namespace (which is already in use, unregistered, in e.g. CIBA) and what the governance for it should be.

Otherwise, I think this looks good (with our without registration of the namespace).

edit to add: in the most recent WG call, there was consensus on registering the namespace. We do not need anything else to proceed with that.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -3544,6 +3542,52 @@ in the IANA "Uniform Resource Identifier (URI) Schemes" registry [@IANA.URI.Sche
...
+
+### Purpose
+
+The Namespace Identifier (NID) "openid" will be used to identify all types of digital resources defined by the OpenID Foundation. These might include resources defined and used within protocols or standards themselves.
+
+### Syntax
+
+The syntax for the openid URN namestring is defined using the ABNF below:
+
+namestring    = "urn:openid:" NSS
```
```suggestion
namestring = "urn:openid:" NSS
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -3544,6 +3542,52 @@ in the IANA "Uniform Resource Identifier (URI) Schemes" registry [@IANA.URI.Sche
 * Change Controller: OpenID Foundation Digital Credentials Protocols Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
 * Reference: (#client-metadata-retrieval) of this specification
 
+## Uniform Resource Names (URN) Namespaces Registry
```
Sakurann, line 3545: based on what have you drafted this section?

Sakurann (approved):

1.1/openid-4-verifiable-credential-issuance-1_1.md
c2bo: I looked at a bunch of examples at https://www.iana.org/assignments/urn-namespaces/urn-namespaces.xhtml (and the RFC).

Vanderkast (approved):
LGTM

1.1/openid-4-verifiable-credential-issuance-1_1.md
jogu: For others, Christian is referring to https://datatracker.ietf.org/doc/html/rfc8141#appendix-A when he says RFC I believe.

paulbastian (approved):

GarethCOliver (approved):

Issue #300, opened 2024-04-18
Add a mechanism for notifying the wallet about certain events


awoie:
ODI4VCI would benefit from a mechanism that allows issuers to send notifications to the wallet or its backend to enable:

- The ability to notify the wallet of specific credential issuance process-related events, e.g., to reduce polling requests at the deferred issuance endpoint.
- The ability to inform the wallet or its backend about certain risk-related events, e.g., fraud, key compromise.
- The ability to notify the wallet of specific credential lifecycle-related events, e.g., a credential has been revoked independently from the status mechanism and associated TTL windows, new credential types can be requested, or the underlying credential dataset has changed, which is relevant for many credential types such as age-over-18, marriage, etc.

A potential solution could be based on the OIDF Shared Signals and related IETF standards work as presented at IIW. The slides can be found [here](https://docs.google.com/presentation/d/1Uc5ZxQWdaLzzAWjA5R5LpbgDBYuuz9sofMouTOPiPvM/).

The solution should support both push and pull notification mechanisms. While the pull mechanism is easier to implement, it has certain limitations. For example, if a wallet must maintain relationships with several issuers, polling event endpoints from those issuers introduces a significant load on the mobile device.

peppelinux:
awesome!

the pull method is pratically what can be achieved using [oauth status attestations](https://peppelinux.github.io/draft-demarco-oauth-status-attestations/draft-demarco-oauth-status-attestations.html)

the push method raises to me some privacy concerns, since the wallet backed would be informed about which credential the wallet instance (holder) has obtained and their status

I'm really excited about your ideas anyway, since I'm wondering as well about how the shared signals may comes into place. In my context they may have sense between the VCI and the authentic sources. When an authentic sources changes the status of the data, previously used by a VCI for the issuance of a credential, the authentic source would signal this change (while the VCI should therefore revoke the credential).

the things I really like is that RISC seems [quite ready](https://openid.net/specs/openid-risc-profile-specification-1_0.html#rfc.section.2.3) for the main use cases I have in mind.

awoie:
> awesome!
> 
> the pull method is pratically what can be achieved using [oauth status attestations](https://peppelinux.github.io/draft-demarco-oauth-status-attestations/draft-demarco-oauth-status-attestations.html), because the wallet instance automatically request and obtains the (relevant) updates about the status of its credentials

[...]

PR #721, opened 2026-03-18
Metadata in credential response

### LINKED ISSUE #421: Add display metadata to the CredentialResponse

c2bo:
Closes #421 

Not really certain about the `only_metadata` parameter in the Credential Request.


1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1258,9 +1258,10 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
...
 * `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If this request element is not present, the corresponding credential response returned is not encrypted.
-    * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
-    * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
-    * `zip`: OPTIONAL. JWE [@!RFC7516] `zip` algorithm [@!RFC7518] for compressing Credential Responses prior to encryption. If absent then compression MUST not be used.
+  * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
+  * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
+  * `zip`: OPTIONAL. JWE [@!RFC7516] `zip` algorithm [@!RFC7518] for compressing Credential Responses prior to encryption. If absent then compression MUST not be used.
```
c2bo, line 1263: Since it doesn't easily show on the github diff: this is only a fix for indentation (3->2 spaces)

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1380,13 +1381,18 @@ If the Credential Response is not encrypted, the media type of the response MUST
...
 * `interval`: REQUIRED if `transaction_id` is present. Contains a positive number that represents the minimum amount of time in seconds that the Wallet SHOULD wait after receiving the response before sending a new request to the Deferred Credential Endpoint. It MUST NOT be used if the `credentials` parameter is present.
 * `notification_id`: OPTIONAL. String identifying one or more Credentials issued in one Credential Response. It MUST be included in the Notification Request as defined in (#notification). It MUST not be used if the `credentials` parameter is not present.
+* `metadata`: OPTIONAL. Object that contains additional metadata specific to the issued Credential(s).
+  * `display`: OPTIONAL. Object that contains additional display information specific to the issued Credential(s). The definitions and contained parameters for this Object are identical to the `display` parameter as defined under the `credential_metadata` parameter in Credential Issuer Metadata (see (#credential-issuer-metadata)). Any values contained here overwrite existing values for display that were contained in the Credential Issuer Metadata.
```
TimoGlastra: What does "Any values contained here overwrite existing values for display that were contained in the Credential Issuer Metadata." mean? Should you merge the display arrays based on e.g. the locale, or does it overwrite the whole display array? The "Any values contained" indicates to me it overwrites based on values within the array, not the whole array.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1380,13 +1381,18 @@ If the Credential Response is not encrypted, the media type of the response MUST
...
+  * `custom`: OPTIONAL. Object that contains additional metadata specific to the issued Credenetial(s). This is an extension point that allows the addition of ecosystem-specific metadata during issuance. Wallets that do not understand the contained parameters MUST ignore them.
```
TimoGlastra: Should the `custom` claim also be added to the Credential Issuer Metadata? What if a value is added to both credential issuer metadata and credential response metadata, should it be added top-level in the issuer metadata, but in the credential response it should be under the `custom` key?

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1258,9 +1258,10 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
...
-    * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
-    * `zip`: OPTIONAL. JWE [@!RFC7516] `zip` algorithm [@!RFC7518] for compressing Credential Responses prior to encryption. If absent then compression MUST not be used.
+  * `jwk`: REQUIRED. Object containing a single public key as a JWK used for encrypting the Credential Response.
+  * `enc`: REQUIRED. JWE [@!RFC7516] `enc` algorithm [@!RFC7518] for encrypting Credential Responses.
+  * `zip`: OPTIONAL. JWE [@!RFC7516] `zip` algorithm [@!RFC7518] for compressing Credential Responses prior to encryption. If absent then compression MUST not be used.
+* `only_metadata`: OPTIONAL. Boolean value that if present and set to `true` signals that the Wallet does not expect any credentials in the Credential Response. The Wallet requests only updates to the metadata of previously issued Credentials as identified by the `credential_identifier` or `credential_configuration_id` parameters. If set to `true`, then the `proofs` parameter MUST NOT be present. If the parameter is absent, the value is treated as `false`.
```
TimoGlastra: I feel like this is reusing the credential endpoint for a completely different purpose. Implementation wise it will have completely different logic I assume? Wouldn't it make more sense to introduce a dedicated endpoint for this?

c2bo: Yeah, that is why I also said I didn't really confident about that part - I was trying to create an initial proposal that reflects the discussion we had in the WG, which was similar to this

1.1/openid-4-verifiable-credential-issuance-1_1.md
c2bo: They way I'd see metadata is in sort of a hierarchy:
- Issuer metadata: most general form level of metadata
- Optional format-specific metadata
- Metadata returned in Credential response

Where Issuer metadata is the default, gets overwritten by format-specific metadata if it exists (only for the values it contains), gets overwritten by credential response (only values that exist in the credential response)

From an implementation perspective:
- Take everything from the issuer metadata
- Check for format-specific metadata and for each value found, overwrite your current metadata construct
- Check Credential Response and for each value found, overwrite your current metadata construct

1.1/openid-4-verifiable-credential-issuance-1_1.md
c2bo: Yeah, I wasn't sure on custom, we have 2 options imho:
- A clearly defined extension point like `custom`
- People just add their own parameters to `display`

I think the first option is a bit cleaner for implementations, that's why I went with that as a starting point

leecam: I think the extended display data will be standardized alongside the credential specific schema. I think it applies to all levels of the display data hierarchy

For example EMVCo is standardizing the Digital Payment Credential schema for payment cards. As part of this spec it also defines the additional display metadata. 

Maybe we need an identifier here like a vct. e.g:
```json
"display" : {
  "name": "Lee's Visa Card",
  "logo": "....",
  "com.emvco.dpc.v1": {
    "dpc_specific_data": "foo"
  }
} 
```
This way wallets can know where this display data is coming from and have versioning etc..
And authors of credential specs now know what they need to define this structure and name it in their spec document.

c2bo: Changed it to that variant and tried to roughly explain how extending the default parameters should work

c2bo:
Changed it to the variant, where it's just "normal" `credential_metadata` and added an explanation w/ examples of the mechanism.
Does this work for you @leecam, @TimoGlastra ?

TimoGlastra:
I like this addition 👍

c2bo:
@Wicpar, could I also ask for a review from you since you provided feedback in the corresponding issue (can't request review directly since you aren't part of the gh organization)?

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1258,9 +1258,10 @@ A Client makes a Credential Request to the Credential Endpoint by sending the fo
...
 * `proofs`: OPTIONAL. Object providing one or more proof of possessions of the cryptographic key material to which the issued Credential instances will be bound to. The `proofs` parameter contains exactly one parameter named as the proof type in (#proof-types), the value set for this parameter is a non-empty array containing parameters as defined by the corresponding proof type.
 * `credential_response_encryption`: OPTIONAL. Object containing information for encrypting the Credential Response. If this request element is not present, the corresponding credential response returned is not encrypted.
```
Wicpar, line 1260: There is no mechanism to request an issuer signature of the response equivalent to the jwt response of the issuer metadata.

1.1/openid-4-verifiable-credential-issuance-1_1.md
Wicpar: I think it's a bit awkward having credential metadata distributed like this as well.

Wicpar (commented):
Thanks a lot @c2bo for letting me chime in.

In a strict review sense there is a missing way to request the response as a JWT signed by the issuer, attesting its provenance. In the context of PSD2 (banking) legislation, all display elements are the responsibility of the bank, the issuer. This means issuers must ensure that there is a way for wallets to ensure that during all presentations a verification of the persisted data is done before displaying them, in addition to app integrity to ensure the display has happened correctly according to the issuer's definition.

--- 
Now in a larger scope does this feature really make sense as it is ?

The original need of the feature is to be able to display card art when receiving the credential, however at that point, one has already the sd-jwt-vc credential ready to decode, and fetch the metadata, and also already has had access to the issuer metadata which also has the credential metadata. It clearly is a nice to have, but does it warrant yet another metadata source ? For that part if it is truly the wish of VCI to provide self contained displayable responses it would make sense to do it like this, but the `only_metadata` seems overreaching.

Another unrelated need, is the ability to verify, update and securely store potentially heavy metadata objects — common to all credentials of that type — in a high assurance environment where all data must have a known and verified use and provenance.

For that purpose Credential Metadata, as in any metadata that that is common to and lives along the credential type for the length of its life, should be a first class object that belongs to the credential type, and should be able to have a format that can be verified and attributed to the same trust chain as the credential itself. Display is a major part of the consent, and should be securable if required by standards relying on this specification.

I think creating a dedicated endpoint to query metadata like `credential_metadata_uri` based on the same behavior of the credential endpoint would make a lot more sense, allowing to get clean, signed and/or encrypted credential metadata.

I would even go to deprecating the credential_metadata in the credential_configurations, the issuer metadata request in principle should only advertise data immediately useful to the issuance, and localized display metadata is sure to grow out of proportion way too easily in real world implementations.

In our use-case we do need to be able to get a minimal credential metadata  JWT with x5c based on the same trust chain as the credential, and with the sub being the credential type identifier (or a stringified, canonicalized version of it), or another way to have it be tied to the credential in an independently verifiable manner.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1819,17 +1823,17 @@ This specification defines the following Credential Issuer Metadata parameters:
...
+  * `credential_metadata`: OPTIONAL. Object containing information relevant to the usage and display of issued Credentials. Credential Format-specific mechanisms can overwrite the information in this object to convey Credential metadata. Format-specific mechanisms, such as SD-JWT VC display metadata are always preferred by the Wallet over the information in this object, which serves as the default fallback. The Credential Response can also overwrite the information in this object to convey additional metadata specific to a specific Credential issuance flow. Metadata contained in the Credential Response is always preferred by the Wallet over other information. Default parameters are defined here, but additional ones MAY be added. See (#display-metadata-considerations) for implementation cosinderations on credential metadata. Below is a non-exhaustive list of parameters that MAY be included:
```
fkj:
```suggestion
  * `credential_metadata`: OPTIONAL. Object containing information relevant to the usage and display of issued Credentials. Credential Format-specific mechanisms can overwrite the information in this object to convey Credential metadata. Format-specific mechanisms, such as SD-JWT VC display metadata are always preferred by the Wallet over the information in this object, which serves as the default fallback. The Credential Response can also overwrite the information in this object to convey additional metadata specific to a specific Credential issuance flow. Metadata contained in the Credential Response MUST be preferred by the Wallet over other information. Default parameters are defined here, but additional ones MAY be added. See (#display-metadata-considerations) for implementation cosinderations on credential metadata. Below is a non-exhaustive list of parameters that MAY be included:
```

fkj:
```suggestion
  * `credential_metadata`: OPTIONAL. Object containing information relevant to the usage and display of issued Credentials. Credential Format-specific mechanisms can overwrite the information in this object to convey Credential metadata. Format-specific mechanisms, such as SD-JWT VC display metadata are always preferred by the Wallet over the information in this object, which serves as the default fallback. The Credential Response can also overwrite the information in this object to convey additional metadata specific to a specific Credential issuance flow. Metadata contained in the Credential Response is always preferred by the Wallet over other information. Default parameters are defined here, but additional ones MAY be added. See (#display-metadata-considerations) for implementation considerations on credential metadata. Below is a non-exhaustive list of parameters that MAY be included:
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2016,6 +2020,80 @@ It is up to the Credential Issuer whether to update both the signature and the c
 
 The Credential Issuer determines the number of the Credentials issued in the Credential Response, regardless of number of proofs/keys contained in the `proofs` parameter in the Credential Request.
 
+## Credential Metadata {#display-metadata-considerations}
+
+Credential metadata is provided by the Credential Issuer, but can be conveyed using different mechanisms. Defined mechanisms to provide display metadata are:
+
+* Credentiel Issuer metadata may contain credenetial metadata as defined in (#credential-issuer-parameters)
```
fkj:
```suggestion
* Credential Issuer metadata may contain credential metadata as defined in (#credential-issuer-parameters)
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2016,6 +2020,80 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+* Credential Formats may define their own mechanisms for metadata
+* Credential Response may also contain display metadata as defined in (#credential-response)
+
+Credential metadata provided via the Credential Issuer metadata should be interpreted as the most generic form of metadata and as a general fallback solution. If defined and present, Credential Format specific metadata data overwrites the values from the Credential Issuer metadata. If present, metadata present in the Credential Response overwrites existing values.
+
+(#credential-issuer-parameters) defines a set of default Credential metadata parameters, but additional ones can be defined and added by profiles or extensions. Other standardization organizations or ecosystems defining extensions to the Credential metadata parameters SHOULD do so by defining a collision-resistant parameter that contains an Object with all parameters they are defining.
+
+Below is a non-normative example how credential metadata present in the Credential Issuer metadata and the Credential Response would result would be merged:
```
fkj:
```suggestion
Below is a non-normative example of how credential metadata present in the Credential Issuer metadata and the Credential Response would be merged:
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2016,6 +2020,80 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+      },
+      "locale": "en-US",
+      "background_color": "#12107c",
+      "text_color": "#FFFFFF"
+    }
+  ]
+}
+```
+
+Credential Netadata contained in the Credenteial Response:
```
fkj:
```suggestion
Credential Metadata contained in the Credential Response:
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2016,6 +2020,80 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+```json
+{
+  "display": [
+    {
+      "logo": {
+        "uri": "data:image/png;base64,F00==",
+      },
+      "com.emvco.dpc.v1": {
```
fkj: Should switch to something generic instead of a real company.

fkj (commented):
I agree with the previous reviewers that it might be best to define a new general endpoint for this. I think it's important to understand what the use cases for being able to update metadata are.
So far, I've seen these two:
1. Obtain signed credential metadata (not only for display, but also in general) for compliance with specific laws, e.g. PSD2. (This is @Wicpar's use case as I understand it).
2. Update credential display name/art/whatever in case the issuer changes branding or similar. We noted this in the NOBID LSP.

Wicpar:
> I agree with the previous reviewers that it might be best to define a new general endpoint for this. I think it's important to understand what the use cases for being able to update metadata are. So far, I've seen these two:

There is also the need to be able to add new locales and update visuals during the lifetime of the credential.
Our use case adds a new `transaction_data_types` parameter that might need to add new transaction data types supported by the credential when standards evolve. It is too heavy to gracefully be provided in the issuer metadata, we have multiple megabyte payloads for a dozen credentials in our test setup issuer metadata right now.

1.1/openid-4-verifiable-credential-issuance-1_1.md
c2bo: combined:

```suggestion
  * `credential_metadata`: OPTIONAL. Object containing information relevant to the usage and display of issued Credentials. Credential Format-specific mechanisms can overwrite the information in this object to convey Credential metadata. Format-specific mechanisms, such as SD-JWT VC display metadata are always preferred by the Wallet over the information in this object, which serves as the default fallback. The Credential Response can also overwrite the information in this object to convey additional metadata specific to a specific Credential issuance flow. Metadata contained in the Credential Response MUST be preferred by the Wallet over other information. Default parameters are defined here, but additional ones MAY be added. See (#display-metadata-considerations) for implementation considerations on credential metadata. Below is a non-exhaustive list of parameters that MAY be included:
```

c2bo:
I won't be able to join the DCP call tomorrow morning, but I'd like to have a discussion on the endpoint part in the WG before we are changing towards a new endpoint.

My understanding was that the last time this was discussed, the point of refreshing metadata was briefly brought up and the initial reaction was to allow repeated calls to the credential endpoint with some sort of flag that this request only updates metadata. I don't disagree with moving this to somewhere else, but I'd like to have rough consensus in a WG meeting first. From my perspective, it would make more sense to use something like feature proposed in #300 for the updates though instead of a new endpoint?

Wicpar:
#300 would be useful for the notification part but doesn't really fulfill the need of having a locale-filtered and signed credential metadata object, which hash allows proper identification of the delivered instance (not possible if the credentials are in there with it).

fkj:
> #300 would be useful for the notification part but doesn't really fulfill the need of having a locale-filtered and signed credential metadata object, which hash allows proper identification of the delivered instance (not possible if the credentials are in there with it).

Why would it not be possible to verify signed metadata along with a credential? Because you want the metadata to be a fixed "template" that can be verified statically? I don't understand why you could not sign a second piece of metadata later, e.g. to add a new locale. What exactly do you need to verify for each credential instance?

c2bo:
> #300 would be useful for the notification part but doesn't really fulfill the need of having a locale-filtered and signed credential metadata object, which hash allows proper identification of the delivered instance (not possible if the credentials are in there with it).

Locale-based filtering might be a bit harder, but the issuer could just store the preference sent during the credential issuance flow and keep using that. We could definitely use this channel to convey signed information (JWTs). I am a bit reluctant to introduce a new/dedicated endpoint for this if we would likely move towards the notification based variant mid-/longterm -> That is exactly what I would like to better understand in a WG discussion.

Wicpar:
> Why would it not be possible to verify signed metadata along with a credential?

it is possible, the issue comes when needing to verify the hash returned in the kb-jwt that is the signature of the metadata to replay when verifying a transaction after the fact. if it has the credentials along with it, it is not the same as without, it will increase the storage requirements quite a bit as all hashes must have a stored metadata object as it was issued

> Because you want the metadata to be a fixed "template" that can be verified statically? 

not fixed but it must be verified against the credential and the wallet unit in an independent manner for each different version that was issued.

> I don't understand why you could not sign a second piece of metadata later, e.g. to add a new locale. 

you could, it's just very inconvenient having hashes multiplied by the number of issuances to store, it becomes quadratic instead of linear. based on user adoption.

> What exactly do you need to verify for each credential instance?

PSD2 requires the bank to prove the user has been properly informed and has given consent, it is critical that the presentation can be verifiably replayed after the fact, that every single display detail be provably as it was at that time, with the assumption that the certified wallet indeed respects the spec as the audit certified.
since display is determined by metadata, it must be signed and linked to the proof, like the credential.

the goal is that the kb-jwt + credential + metadata + wallet unit allow the bank to prove the transaction went as defined without user cooperation. for that a dedicated signed credential metadata is necessary, having credentials in there sometimes complicates things significantly.

It is possible to do by calling the metadata endpoint again to get only the metadata, let me think about it a bit more. i jumped the gun a bit when saying it is not possible rather than impractical, sorry about that.

Wicpar:
> Locale-based filtering might be a bit harder, but the issuer could just store the preference sent during the credential issuance flow and keep using that. 

that is not appropriate as the user preference in the wallet might change at any point in time and should trigger a refresh of the metadata in the relevant locales

> We could definitely use this channel to convey signed information (JWTs). I am a bit reluctant to introduce a new/dedicated endpoint for this if we would likely move towards the notification based variant mid-/longterm -> That is exactly what I would like to better understand in a WG discussion.

my worry about a pure notification approach is that it would remove control from the wallet unit on how to refresh the data

fkj:
> it is possible, the issue comes when needing to verify the hash returned in the kb-jwt that is the signature of the metadata to replay when verifying a transaction after the fact. if it has the credentials along with it, it is not the same as without, it will increase the storage requirements quite a bit as all hashes must have a stored metadata object as it was issued

I wonder if this could be solved by returning the metadata in a nested JWT. It could then be verified independently.

> you could, it's just very inconvenient having hashes multiplied by the number of issuances to store, it becomes quadratic instead of linear. based on user adoption.

I don't understand this. Would it not be one hash per issued credential metadata instance?

> since display is determined by metadata, it must be signed and linked to the proof, like the credential.

Only the metadata that was actually used in the particular presentation needs to be linked to the proof, right? Not every single metadata set present.


On the notification approach, I agree that it is useful but not sufficient. The wallet may also want to initiate fetching new metadata if the user e.g. changes language on their device.

Wicpar:
> I wonder if this could be solved by returning the metadata in a nested JWT. It could then be verified independently.

yes, it's just a bit weird to do nested jwts, also how would the request params for the inner jwts look like, it just seems a bit odd.


> I don't understand this. Would it not be one hash per issued credential metadata instance?

only if the credential response is not in there, hashing anything else than the signed JWT introcuces too much uncertainty that has to be compensated with extra standard definition. so ideally we would have a clean credential metadata only JWT

> Only the metadata that was actually used in the particular presentation needs to be linked to the proof, right? Not every single metadata set present.

Yes if a wallet has a metadata object for FR and one for DE it would just insert the hash of the one used in the KB-JWT. But it cannot divide up a metadata object as the signature must be part of the hash for the proof, so that it can be queried by hash in the issuer database alongside the credential and WUA if there is a case that needs proof without user cooperation.

> On the notification approach, I agree that it is useful but not sufficient. The wallet may also want to initiate fetching new metadata if the user e.g. changes language on their device.

👍

c2bo:
Following the WG discussion today, I removed the update metadata part. The PR now only contains the addition of optional `credential_metadata` in the Credential Response.

Update to be discussed/solved in another PR.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2016,6 +2017,82 @@ It is up to the Credential Issuer whether to update both the signature and the c
 
 The Credential Issuer determines the number of the Credentials issued in the Credential Response, regardless of number of proofs/keys contained in the `proofs` parameter in the Credential Request.
 
+## Credential Metadata {#display-metadata-considerations}
+
+Credential metadata is provided by the Credential Issuer, but can be conveyed using different mechanisms. Defined mechanisms to provide display metadata are:
+
+* Credential Issuer metadata may contain credential metadata as defined in (#credential-issuer-parameters)
+* Credential Formats may define their own mechanisms for metadata
+* Credential Response may also contain display metadata as defined in (#credential-response)
```
fkj:
```suggestion
* Credential Response may also contain metadata as defined in (#credential-response)
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2016,6 +2017,82 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+Credential metadata provided via the Credential Issuer metadata should be interpreted as the most generic form of metadata and as a general fallback solution. If defined and present, Credential Format specific metadata data overwrites the values from the Credential Issuer metadata. If present, metadata present in the Credential Response overwrites existing values.
```
fkj: Remove some "duplicate" words:
```suggestion
Credential metadata provided via the Credential Issuer metadata should be interpreted as the most generic form of metadata and as a general fallback solution. If defined and present, Credential Format specific metadata overwrites the values from the Credential Issuer metadata. If present, metadata in the Credential Response overwrites existing values.
```

fkj: I think this would clarify a bit:
```suggestion
Credential metadata provided via the Credential Issuer metadata should be interpreted as the most generic form of metadata and as a general fallback solution. If defined and present, Credential Format specific metadata data overwrites the values from the Credential Issuer metadata. If present, metadata present in the Credential Response overwrites all other existing values.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2016,6 +2017,82 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+      "com.example.domain": {
+        "some_text": "something_really_important",
+        "more_text": "also_really_important"
+      }
+    }
+  ]
+}
+```
+
+Note that `com.emvco.dpc.v1` in the above example shows how an extension would be defined and used.
```
fkj: Forgot to change?
```suggestion
Note that `com.example.domain` in the above example shows how an extension would be defined and used.
```

fkj (commented):
It looks like we are using a mix of "Credential Metadata" and "Credential metadata". We should decide if it is a Capital Letter Concept or not and be consistent

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1819,17 +1820,17 @@ This specification defines the following Credential Issuer Metadata parameters:
...
+  * `credential_metadata`: OPTIONAL. Object containing information relevant to the usage and display of issued Credentials. Credential Format-specific mechanisms can overwrite the information in this object to convey Credential metadata. Format-specific mechanisms, such as SD-JWT VC display metadata are always preferred by the Wallet over the information in this object, which serves as the default fallback. The Credential Response can also overwrite the information in this object to convey additional metadata specific to a specific Credential issuance flow. Metadata contained in the Credential Response MUST be preferred by the Wallet over other information. Default parameters are defined here, but additional ones MAY be added. See (#display-metadata-considerations) for implementation considerations on credential metadata. Below is a non-exhaustive list of parameters that MAY be included:
```
```suggestion
  * `credential_metadata`: OPTIONAL. Object containing information relevant to the usage and display of issued Credentials. Credential Format-specific mechanisms can overwrite the information in this object to convey Credential metadata and is preferred by the Wallet over the information in this object, which serves as the default fallback. Additionally, the Credential Response can overwrite the information in this object and MUST be preferred by the Wallet over other information. See (#display-metadata-considerations) for implementation considerations on credential metadata. Below is a default, non-exhaustive, extensible list of parameters that MAY be included:
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
paulbastian: Do you mean Signed metadata? There is no such possibility other than TLS, but that's not the focus of this PR

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1380,10 +1380,11 @@ If the Credential Response is not encrypted, the media type of the response MUST
...
 * `interval`: REQUIRED if `transaction_id` is present. Contains a positive number that represents the minimum amount of time in seconds that the Wallet SHOULD wait after receiving the response before sending a new request to the Deferred Credential Endpoint. It MUST NOT be used if the `credentials` parameter is present.
 * `notification_id`: OPTIONAL. String identifying one or more Credentials issued in one Credential Response. It MUST be included in the Notification Request as defined in (#notification). It MUST not be used if the `credentials` parameter is not present.
+* `credential_metadata`: OPTIONAL. Object that contains additional metadata specific to the issued Credential(s). The definitions and contained parameters for this Object are identical to the `credential_metadata` parameter as defined in Credential Issuer Metadata (see (#credential-issuer-metadata)).
```
```suggestion
* `credential_metadata`: OPTIONAL. Object that contains additional metadata specific to the issued Credential(s). The definitions and contained parameters for this Object are identical to the `credential_metadata` parameter as defined in Credential Issuer Metadata (see (#credential-issuer-metadata)) See (#display-metadata-considerations) for implementation considerations on credential metadata.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2016,6 +2017,82 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+* Credential Issuer metadata may contain credential metadata as defined in (#credential-issuer-parameters)
+* Credential Formats may define their own mechanisms for metadata
+* Credential Response may also contain metadata as defined in (#credential-response)
+
+Credential metadata provided via the Credential Issuer metadata should be interpreted as the most generic form of metadata and as a general fallback solution. If defined and present, Credential Format specific metadata overwrites the values from the Credential Issuer metadata. If present, metadata in the Credential Response overwrites existing values.
+
+(#credential-issuer-parameters) defines a set of default Credential metadata parameters, but additional ones can be defined and added by profiles or extensions. Other standardization organizations or ecosystems defining extensions to the Credential metadata parameters SHOULD do so by defining a collision-resistant parameter that contains an Object with all parameters they are defining.
+
+Below is a non-normative example of how credential metadata present in the Credential Issuer metadata and the Credential Response would be merged:
```
```suggestion
Below is a non-normative example of how credential metadata from the Credential Issuer metadata and the Credential Response would be merged:
```

paulbastian (approved):

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2016,6 +2017,82 @@ It is up to the Credential Issuer whether to update both the signature and the c
 
 The Credential Issuer determines the number of the Credentials issued in the Credential Response, regardless of number of proofs/keys contained in the `proofs` parameter in the Credential Request.
 
+## Credential Metadata {#display-metadata-considerations}
+
+Credential metadata is provided by the Credential Issuer, but can be conveyed using different mechanisms. Defined mechanisms to provide display metadata are:
+
+* Credential Issuer metadata may contain credential metadata as defined in (#credential-issuer-parameters)
```
GarethCOliver:
```suggestion
* Credential Issuer metadata MAY contain credential metadata as defined in (#credential-issuer-parameters)
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2016,6 +2017,82 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+* Credential Formats may define their own mechanisms for metadata
```
GarethCOliver:
```suggestion
* Credential Formats MAY define their own mechanisms for metadata
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2016,6 +2017,82 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+* Credential Response may also contain metadata as defined in (#credential-response)
```
GarethCOliver:
```suggestion
* Credential Response MAY also contain metadata as defined in (#credential-response)
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -2016,6 +2017,82 @@ It is up to the Credential Issuer whether to update both the signature and the c
...
+
+Credential metadata provided via the Credential Issuer metadata should be interpreted as the most generic form of metadata and as a general fallback solution. If defined and present, Credential Format specific metadata overwrites the values from the Credential Issuer metadata. If present, metadata in the Credential Response overwrites existing values.
```
GarethCOliver:
```suggestion
Credential metadata provided via the Credential Issuer metadata SHOULD be interpreted as the most generic form of metadata and as a general fallback solution. If defined and present, Credential Format specific metadata overwrites the values from the Credential Issuer metadata. If present, metadata in the Credential Response overwrites existing values.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
GarethCOliver: You can probably squint and say that this is trusted the same because the credential endpoint is specified in the (signed) issuer metadata (even if the credential metadata is no longer directly signed).

1.1/openid-4-verifiable-credential-issuance-1_1.md
GarethCOliver: Given the way we have written Deferred Credential Response we need to update the text here to allow credential_metadata to be returned (as we don't just specify  Deferred Credential Response === Credential Response).

1.1/openid-4-verifiable-credential-issuance-1_1.md
Sakurann: this is out of scope of this PR, please open a separate issue.

Sakurann:
DCP WG discussion:
- @Wicpar can you please open a separate issue about metadata update. so that we can have a discussion there and do another follow up PR. current PR does not specify signing of the display metadata.
- current PR already removes the update metadata part. Please re-review so that we can merge this as a starting point.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1819,17 +1820,17 @@ This specification defines the following Credential Issuer Metadata parameters:
...
-  * `credential_metadata`: OPTIONAL. Object containing information relevant to the usage and display of issued Credentials. Credential Format-specific mechanisms can overwrite the information in this object to convey Credential metadata. Format-specific mechanisms, such as SD-JWT VC display metadata are always preferred by the Wallet over the information in this object, which serves as the default fallback. Below is a non-exhaustive list of parameters that MAY be included:
+  * `credential_metadata`: OPTIONAL. Object containing information relevant to the usage and display of issued Credentials. Credential Format-specific mechanisms can overwrite the information in this object to convey Credential metadata and is preferred by the Wallet over the information in this object, which serves as the default fallback. Additionally, the Credential Response can overwrite the information in this object and MUST be preferred by the Wallet over other information. See (#display-metadata-considerations) for implementation considerations on credential metadata. Below is a default, non-exhaustive, extensible list of parameters that MAY be included:
```
```suggestion
  * `credential_metadata`: OPTIONAL. Object containing information relevant to the usage and display of issued Credentials. Credential Format-specific mechanisms can overwrite the information in this object to convey Credential metadata and is preferred by the Wallet over the information in this object, which serves as the default fallback. Additionally, the Credential Response MUST be preferred by the Wallet over any other information, effectively overwriting the information in this object and any Credential Format-specific mechanisms. See (#display-metadata-considerations) for implementation considerations on credential metadata. Below is a default, non-exhaustive, extensible list of parameters that MAY be included:
```
based on the question @martijnharing raised in dcp wg call. intent being

> credential_metadata itself is optional, so if it is not present, issuer_metadata if source of truth, but if it is there, it overwrites other sources.

Wicpar:
> * @Wicpar can you please open a separate issue about metadata update. so that we can have a discussion there and do another follow up PR. current PR does not specify signing of the display metadata.

Done in #735

fkj (approved):

1.1/openid-4-verifiable-credential-issuance-1_1.md
c2bo: added  `credentia_metadata` to the deferred credential response (optional) as well - good catch!

EsTharian (approved):

GarethCOliver (approved):

PR #753, opened 2026-06-16
First draft of OpenId4VCI-S2S

GarethCOliver:
This is the first draft of the s2s variant of credential issuance.

Current version is to be submitted as the base contribution accepted by DCP WG. Issues opened for tracking future work using server-to-server tag. 

Before this is ready to review there needs to be:
 - Flow Diagrams
 - Example flows for Credential Lifecycle

Notable missing features:

 - Atomic Credential Sets
 - Specified VerificationData for common uses
 - Reverification
 - Schemas

Additionally, before the first implementors draft a convergence effort with existing VCI needs to be undertaken including:
  - Credential Offer
  - IAE
  - Credential Endpoint
  - Notifications
  - Encryption
  - Issuer and Credential Metadata


s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+
+## Rate Limiting {#rate-limiting}
+
+Clients receiving HTTP 429 MUST honor the `Retry-After` header. Exponential backoff is RECOMMENDED: 0–1s, 1.1–5s, 5.1–30s (max 3 retries).
+
+## Identifier Model {#identifier-model}
+
+The API uses the following identifiers:
+
+- SessionId: An identifier associated with a particular Wallet Instance, persistent throughout the process.
```
GarethCOliver: Wallet Client Instance
```suggestion
- SessionId: An identifier associated with a particular Wallet Client Instance, persistent throughout the process.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+
+## Wallet Initiated (Negotiated Verification) {#wallet-initiated}
+
+1. Holder, in the Wallet Client Instance, selects a Credential to Issue where the required VerificationData is unknown.
+1. The Wallet Client generates a WSK and calls `verification/initiate` on the Issuer Server via the Wallet Server.
+1. The Issuer validates the verificationNonce and returns a status of `ADDITIONAL_INFO_REQUIRED` along with a VerificationDataRequest.
+1. The Wallet Client collects the requested VerificationData from the Holder.
+1. The Wallet Client signs-then-encrypts the VerificationData and sends it to the Wallet Server, which calls `verification/supplement` on the Issuer Server.
+1. The Issuer decrypts the clientEncryptedData and verifies it is signed by the previously established WSK
+1. The Issuer asynchronously verifies the VerificationData
```
dpostnikov: Issuer returns a status of `PENDING` or something like that? before Async processing.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+fullname="Gareth Oliver"
+organization="Google"
+    [author.address]
+    email = "gco@google.com"
+
+%%%
+
+.# Abstract
+
+This specification defines an API for the issuance and management of Verifiable Digital Credentials between a Wallet Server and Issuer Server, while enabling sensitive user data to remain confidential between a Wallet Client Instance and the Issuer.
```
Sakurann:
```suggestion
This specification defines an API for the issuance and management of Verifiable Credentials between a Wallet Server and Issuer Server, while keeping sensitive user data confidential between a Wallet Client Instance and the Issuer without revealing it to the Wallet Server.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+
+{mainmatter}
+
+# Introduction {#introduction}
+
+This specification defines the contract between a Wallet Server and one or more
+Credential Issuer Servers for credential issuance and lifecycle management. It
+is credential format agnostic so Credentials can be of any format, including,
+but not limited to, IETF SD-JWT VC [@I-D.ietf-oauth-sd-jwt-vc], ISO mdoc
+[@ISO.18013-5], and W3C VCDM [@VC_DATA_2.0].
```
Sakurann:
```suggestion
This specification defines a set of APIs used by the Wallet Server
for credential issuance and lifecycle management. It
is credential format agnostic. Credentials can be of any format, including,
but not limited to, IETF SD-JWT VC [@I-D.ietf-oauth-sd-jwt-vc], ISO mdoc
[@ISO.18013-5], and W3C VCDM [@VC_DATA_2.0].
```

Sakurann: "contract" did not feel descriptive enough

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
Sakurann: do we want to change terminology to Verifiable Digital Credentials..?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+when, and only when, they appear in all capitals, as shown here.
+
+# Terminology {#terminology}
+
+
+This specification defines the following terms. In the case where a term has a definition that differs, the definition below is authoritative for this specification.
+
+
+Credential (or Verifiable Digital Credential (VDC)):
+:  An instance of a Credential Configuration with a particular Credential Dataset, that is signed by an Issuer and can be cryptographically verified. An Issuer may provide multiple Credentials as separate instances of the same Credential Configuration and Credential Dataset but with different cryptographic values. In this specification, the term "Verifiable Digital Credential" is also referred to as "Credential".
```
Sakurann: why redefine what is already defined in VCI..? because calling it a VDC is very important?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+:  Credential Issuer's description of a particular kind of Credential that the Credential Issuer is offering to issue. A Credential Configuration is identified by a Credential Configuration Identifier string that is unique to an Issuer.
+
+Credential Format:
+:  Data Model used to create and represent Credential information. This format defines how various pieces of data within a Verifiable Digital Credential are organized and encoded, ensuring that the Verifiable Digital Credential can be consistently understood, processed, and verified by different systems.
+
+Credential Issuer (or Issuer):
+:  An entity that issues Verifiable Credentials. In the context of this specification, the Credential Issuer implements the Issuer Endpoints.
+
+CredentialInstanceId:
+:  A unique identifier at an Issuer for a particular Credential Dataset on a particular Wallet Client Instance.
```
Sakurann: this looks like a parameter name... I'd suggest either define "Credential Instance Identifier" or remove CredentialInstanceId from terminology.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+:  Data that is presented to a specific Verifier, derived from one or more Verifiable Credentials that can be from the same or different Credential Issuers. It can be of any Credential Format.
+
+SessionId:
+:  A globally unique identifier associated with a particular Wallet Client Instance, persistent throughout the issuance process.
+
+Verification:
+:  Verification in this specification refers to the process by which an Issuer determines that a Wallet Client Instance is authorized to have particular Credentials issued to it.
+
+VerificationId:
+:  An identifier generated by the Issuer for a particular verification session. Must be unique for a given SessionId.
```
Sakurann, line 77: same comment as for CredentialInstanceId. looks like a parameter name... I'd suggest either define "verification Identifier" or remove it from terminology.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+:  An entity that receives Verifiable Credentials and has control over them to present them to the Verifiers as Presentations.
+
+Presentation:
+:  Data that is presented to a specific Verifier, derived from one or more Verifiable Credentials that can be from the same or different Credential Issuers. It can be of any Credential Format.
+
+SessionId:
+:  A globally unique identifier associated with a particular Wallet Client Instance, persistent throughout the issuance process.
+
+Verification:
+:  Verification in this specification refers to the process by which an Issuer determines that a Wallet Client Instance is authorized to have particular Credentials issued to it.
```
Sakurann, line 74: Verification is way too generic as a term to be defined. either be more specific "Wallet Client Instance Verification", but i'd rather not define this as a term

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+:  A unique identifier at an Issuer for a particular Credential Dataset on a particular Wallet Client Instance.
+
+Holder:
+:  An entity that receives Verifiable Credentials and has control over them to present them to the Verifiers as Presentations.
+
+Presentation:
+:  Data that is presented to a specific Verifier, derived from one or more Verifiable Credentials that can be from the same or different Credential Issuers. It can be of any Credential Format.
+
+SessionId:
+:  A globally unique identifier associated with a particular Wallet Client Instance, persistent throughout the issuance process.
```
Sakurann: same comment as for CredentialInstanceId. looks like a parameter name... I'd suggest either define "Session Identifier", which is way too obvious to be defined IMHO or remove it from terminology.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in BCP 14 [@!RFC2119] [@!RFC8174]
+when, and only when, they appear in all capitals, as shown here.
+
+# Terminology {#terminology}
+
+
+This specification defines the following terms. In the case where a term has a definition that differs, the definition below is authoritative for this specification.
```
Sakurann, line 55: why each of the following terms from VCI/VP are redefined?: Credential, Credential Configuration, Credential Issuer, Credential Format, Holder, Presentation, Wallet, etc. 

Please do not redefine and simply say: 
> This specification uses the terms "XXX" defined by VCI , "ZZZ" defined by VP, etc.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+:  A key type provided by a Wallet Client Instance for the Issuer to encrypt sensitive data to.
+
+Wallet Provider:
+:  The entity who provides the technical solutions for the Wallet Client and the Wallet Server.
+
+Wallet Server:
+:  A server component that provides infrastructure to facilitate improved reliability and security of credential issuance and lifecycle management. In the context of this specification it is responsible for implementing the Wallet Server Endpoints, and communicating with the Wallet Client.
+
+Wallet Signing Key (WSK):
+:  An asymmetric key pair generated by a Wallet Client Instance used as a proof-of-origin for payload generated on the client for the Issuer.
```
Sakurann, line 98: would not define this in terminology. it is an artifact in the spec. if we follow this, we would need to define key attestation, wallet attestation, etc. in the terminology of VCI.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
Sakurann: Wallet Client, Wallet Client Instance, Wallet Server are terms that seem to be most valuable to be defined in this spec

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+
+# Overview  {#overview}
+
+This specification defines endpoints on both the Wallet Server and Credential
+Issuer Server to perform credential issuance and lifecycle management. This
+specification is designed for use with an architecture involving 3 components:
```
Sakurann, line 104: is there a reason why `Credential Issuer Server` is not defined in terminology? why can't it just be `Credential Issuer`? even in VCI, Issuer is assumed to be a server.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+      |  |             |  Out of |             |  | OpenID4VCI | |  Server   | |
+      |  +-------------+  Scope  +-------------+  |  S2S API   | +-----------+ |
+      |                                           |            |               |
+      +-------------------------------------------+            +---------------+
+~~~
+{: #architecture-diagram title="Architecture Overview"}
+
+The goal of this protocol is to enable a Wallet Server to improve the reliability, security and management of Credentials issued to the Holder while using application-layer encryption and wallet client authentication to minimize the sensitive user data unnecessarily exposed to the Wallet Server. Due to their distributed nature, Client applications are notoriously difficult to update, have poor connectivity, slower analytics and are a less trusted environment. An Issuer Server could work around these issues, as the provider of both the Client and the Server, the Wallet Provider can be in a better position to alleviate these problems.
+
+At a high level the protocol works in two phases; Verification and Credential Management.
```
Sakurann:
```suggestion
At a high level the protocol works in two phases; Verification of the Wallet Client Instance by the Issuer and Credential Management post-initial issuance of the Credential.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+
+Verification is as follows:
+
+- The Wallet Server and Issuer Server mutually authenticate with each other using mTLS, prior to any endpoint interaction.
+- As part of initiating verification of the Wallet Client Instance, the Wallet Client establishes an attested public key (WSK) which is used to authenticate payloads originating on Wallet Client.
+- VerificationData is collected on the Wallet Client and encrypted to the Issuer Server to demonstrate the
```
Sakurann: Please do not use parameter names in the overview before defining the parameter?
```suggestion
- Data used for verification is collected on the Wallet Client and encrypted to the Issuer Server to demonstrate the
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+      |  |   Wallet    |         |   Wallet    |  |            | +-----------+ |
+      |  |   Client    |<------->|   Server    |<--------------->|  Issuer   | |
+      |  |             |  Out of |             |  | OpenID4VCI | |  Server   | |
+      |  +-------------+  Scope  +-------------+  |  S2S API   | +-----------+ |
+      |                                           |            |               |
+      +-------------------------------------------+            +---------------+
+~~~
+{: #architecture-diagram title="Architecture Overview"}
+
+The goal of this protocol is to enable a Wallet Server to improve the reliability, security and management of Credentials issued to the Holder while using application-layer encryption and wallet client authentication to minimize the sensitive user data unnecessarily exposed to the Wallet Server. Due to their distributed nature, Client applications are notoriously difficult to update, have poor connectivity, slower analytics and are a less trusted environment. An Issuer Server could work around these issues, as the provider of both the Client and the Server, the Wallet Provider can be in a better position to alleviate these problems.
```
Sakurann: feels like this belongs in the introduction tbh 

```suggestion
This protocol enables a Wallet Server to improve reliability and security, and have a reliable lifecycle management of Credentials issued to the Holder. Application-layer encryption and wallet client authentication is used to minimize the unnecessary exposure of sensitive user data to the Wallet Server. Due to their distributed nature, Client applications are notoriously difficult to update, have poor connectivity, slower analytics and are a less trusted environment. An Issuer Server could work around these issues, as the provider of both the Client and the Server, the Wallet Provider can be in a better position to alleviate these problems.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+
+At a high level the protocol works in two phases; Verification and Credential Management.
+
+Verification is as follows:
```
Sakurann:
```suggestion
Issuer verifies Wallet Client Instance in the following steps:
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+
+- The Wallet Server and Issuer Server mutually authenticate with each other using mTLS, prior to any endpoint interaction.
+- As part of initiating verification of the Wallet Client Instance, the Wallet Client establishes an attested public key (WSK) which is used to authenticate payloads originating on Wallet Client.
```
Sakurann:
```suggestion
- As part of initiating verification of the Wallet Client Instance, the Wallet Client establishes an attested public key (WSK) which is later used to authenticate payloads originating on Wallet Client.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+- VerificationData is collected on the Wallet Client and encrypted to the Issuer Server to demonstrate the
+  - The Wallet Server MAY also contribute additional VerificationData
+  - This process is potentially asynchronous, while the Issuer evaluates the outcome.
```
Sakurann: either mark this as a to-do, or be more assertive
```suggestion
  - This process is can be asynchronous, while the Issuer evaluates the outcome.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+
+After the Wallet Client Instance has been authenticated and is determined as authorized to have the particular Credential Instance. This works as follows:
```
Sakurann:
```suggestion
After the Wallet Client Instance has been authenticated and is authorized to have the particular Credential Instance. This works as follows:
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,904 @@
...
+
+- The authorized CredentialInstanceIds per Credential DataSet are retrieved from the Provision Credential endpoint. Each Credential Instance MAY be managed independently.
```
Sakurann: I'd rather we define all endpoints upfront before using them in the sentence

Sakurann: also, should be `Credential Provisioning Endpoint` to be more descriptive

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+Issuance begins with a Verification phase. This phase serves two purposes: 
+
+ - Establishing a Wallet Signing Key (WSK) with the Issuer that is under control of the Holder of the Wallet Client Instance.
+ - Authenticate the Holder of the Wallet Client Instance to the Issuer, and determine what Credential Datasets it is authorized to have issued to it.
+
+This is done as follows: 
+
+1. The Wallet Client Instance generates a WSK, and the Wallet Server initiates the Verification on the Issuer Server.
+  1. The Wallet provides a stable unique Session Id associated with the WSK and the Issuer provides a stable Verification Id to reference this session.
```
deshmukhrajvardhan, line 141: should this be "Wallet Client Instance"?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Done

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Could be, or could be the Wallet Server that generates it and provides it to the Wallet Client. It's just a UUID, so doesn't really matter.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+fullname="Gareth Oliver"
+organization="Google"
+    [author.address]
+    email = "gco@google.com"
+
+%%%
+
+.# Abstract
+
+This specification defines an API for the issuance and management of Verifiable Digital Credentials between a Wallet Server and Issuer Server, while enabling sensitive user data to remain confidential between a Wallet Client Instance and the Issuer.
```
GarethCOliver:
```suggestion
This specification defines an API for the issuance and management of Verifiable Digital Credentials between a Wallet Server and Issuer Server, while keeping sensitive user data confidential between a Wallet Client Instance and the Issuer (i.e., not revealing it to the Wallet Server).
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+  1. Examples include digital credential presentations, auth on web, wallet collected documents and liveness/selfie checks. 
+1. The Wallet Server optionally provides additional Verification Data and calls the Issuer Server.
+  1. As an optimization, Verification Data can be provided in the initial call.
+  1. Examples include risk and fraud signals or server collected verification data and evaluations. 
+1. The Issuer Server decrypts and verifies the Verification Data using the WSK, and evaluates it along with any Wallet Server provided Verification Data. 
+  1. The Issuer can perform the evaluation asynchronously to support evaluations that are not immediate (such as human reviews). 
+  1. The Issuer Server can optionally repeat this process to trigger the collection of additional Verification.
+1. After reaching a verdict the Issuer Server updates the verification status and notifies the Wallet Server of the change.
+
+On successful completion the Issuer Server now has a WSK that can be used to authenticate payloads as originating on a particular Wallet Client Instance, and have established which Credentials can be issued.
```
Sakurann:
```suggestion
On successful completion the Issuer Server now has a WSK that can be used to authenticate payloads as originating on a particular Wallet Client Instance, and have authenticated the Holder. 
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+  1. As an optimization, Verification Data can be provided in the initial call.
+  1. Examples include risk and fraud signals or server collected verification data and evaluations. 
+1. The Issuer Server decrypts and verifies the Verification Data using the WSK, and evaluates it along with any Wallet Server provided Verification Data. 
+  1. The Issuer can perform the evaluation asynchronously to support evaluations that are not immediate (such as human reviews). 
+  1. The Issuer Server can optionally repeat this process to trigger the collection of additional Verification.
+1. After reaching a verdict the Issuer Server updates the verification status and notifies the Wallet Server of the change.
+
+On successful completion the Issuer Server now has a WSK that can be used to authenticate payloads as originating on a particular Wallet Client Instance, and have established which Credentials can be issued. 
+
+The Wallet retrieves the credentials as follows:
```
Sakurann, line 154: what's the purpose of this phase?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+~~~
+{: #architecture-diagram title="Architecture Overview"}
+
+The goal of this protocol is to enable a Wallet Server to improve the reliability, security and management of Credentials issued to the Holder while using application-layer encryption and wallet client authentication to minimize the sensitive user data unnecessarily exposed to the Wallet Server. Due to their distributed nature, Client applications are notoriously difficult to update, have poor connectivity, slower analytics and are a less trusted environment. An Issuer Server could work around these issues, but as the provider of both the Client and the Server, the Wallet Provider can be in a better position to alleviate these problems.
+
+At a high level the protocol works as follows: 
+
+In advance of any communication, the Wallet Server and Issuer Server authenticate via mTLS, which is used at all endpoints. The Wallet Server is responsible for authenticating the Wallet Client and ensuring it is in a good state before communicating with the Issuer Server. 
+
+Issuance begins with a Verification phase. This phase serves two purposes:
```
Sakurann, line 132: is there issuer metadata retrieval step? How does the issuer know which credential wallet instance is asking for to know which verification data is needed?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+  1. The Issuer can perform the evaluation asynchronously to support evaluations that are not immediate (such as human reviews). 
+  1. The Issuer Server can optionally repeat this process to trigger the collection of additional Verification.
+1. After reaching a verdict the Issuer Server updates the verification status and notifies the Wallet Server of the change.
+
+On successful completion the Issuer Server now has a WSK that can be used to authenticate payloads as originating on a particular Wallet Client Instance, and have established which Credentials can be issued. 
+
+The Wallet retrieves the credentials as follows:
+
+1. The Wallet Server retrieves the Credential Instance Identifiers from the Wallet Server using the Session Id and Verification Id. 
+1. The Wallet Client generates proofs for Presentation Keys and signs-then-encrypts them using the WSK. The Wallet Client also creates an encryption key and signs it with the WSK.
```
Sakurann: > signs-then-encrypts them using the WSK. The Wallet Client also creates an encryption key and signs it with the WSK. 

WSK is used for signing and encryption? but then another encryption key is generated?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+The Wallet retrieves the credentials as follows:
+
+1. The Wallet Server retrieves the Credential Instance Identifiers from the Wallet Server using the Session Id and Verification Id. 
+1. The Wallet Client generates proofs for Presentation Keys and signs-then-encrypts them using the WSK. The Wallet Client also creates an encryption key and signs it with the WSK. 
+1. The Wallet Server retrieves a batch of Credentials using the Credential Instance Identifiers and the client payload. 
+  1. The Credentials can be retrieved asynchronously by the Issuer Server returning a PENDING state. 
+1. The Issuer verifies the keys originated on the correct client using the WSK, validates the proofs and creates the Credentials. The Credentials are encrypted using the Wallet Encryption Key before being sent back to the Wallet.
+  1. Credential Metadata, such as display can optionally be returned as well.
+
+This process is repeated to refresh the Credentials and to update them. Credentials can be managed through the following processes:
```
Sakurann:
```suggestion
This process is repeated to refresh the Credentials and to update them. Post initial issuance, Credentials lifecycle can be managed through the following processes:
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+- The Issuer Server and Wallet Server support bi-directional notification channels to prompt actions such as updates and retrievals. 
+
+# Endpoints {#endpoints}
+
+This specification defines a series of endpoints on both the Wallet Server and the Issuer Server. Endpoints.
+
+| **Endpoint**              | **Path**                 | **Implemented By** | **Description**                                                                                          |
+|---------------------------|--------------------------|--------------------|----------------------------------------------------------------------------------------------------------|
+| Verification Initiate     | /verification/initiate   | Issuer             | Starts the verification process for a Wallet Client Instance for a particular set of credential configurations. |
+| Verification Supplement   | /verification/supplement | Issuer             | Provides additional verification data for an ongoing verification session.                               |
```
Sakurann, line 176: why can't this be the same endpoint like with IAE?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+| Get Verification Status   | /verification/status     | Issuer             | Queries the current status of a verification session.                                                    |
+| Verification Notification | /verification/notify     | Wallet             | Allows the Issuer to notify the Wallet of a verification status change.                                  |
+| Verification Cancellation | /verification/cancel     | Issuer             | Cancels an ongoing verification session.                                                                 |
+| Provision Credentials     | /credential/provision    | Issuer             | Fetches credential instance IDs after verification approval.                                             |
+| Get Credential            | /credential/get          | Issuer             | Retrieves the verifiable credentials for a particular credential instance.                               |
```
Sakurann, line 181: what's the difference of these two endpoints?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+|---------------------------|--------------------------|--------------------|----------------------------------------------------------------------------------------------------------|
+| Verification Initiate     | /verification/initiate   | Issuer             | Starts the verification process for a Wallet Client Instance for a particular set of credential configurations. |
+| Verification Supplement   | /verification/supplement | Issuer             | Provides additional verification data for an ongoing verification session.                               |
+| Get Verification Status   | /verification/status     | Issuer             | Queries the current status of a verification session.                                                    |
+| Verification Notification | /verification/notify     | Wallet             | Allows the Issuer to notify the Wallet of a verification status change.                                  |
+| Verification Cancellation | /verification/cancel     | Issuer             | Cancels an ongoing verification session.                                                                 |
```
Sakurann, line 179: why can't this be one endpoint implemented by Both like Credential management endpoint? like Verification management endpoint or something?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+- Decrypt the outer JWE.
+- Verify the signature of the JWS matches the previously established Wallet Signing Key.
+- Verify that the `aud` matches the base64url encoded JWK SHA-256 thumbprint used to encrypt the outer JWE.
+
+If this process fails for any received Client Encrypted Payload the Issuer MUST reject the request.
+
+### KeyData
+
+All keys are communicated between the Wallet and Issuer using this common key structure.
```
Sakurann, line 270: is this JWK..? `Field` where?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+Implementations MAY define their own key types. It is RECOMMENDED that collision-resistant values are used, such as r-DNS or urns.
+
+**TODO**: add post quantum
+
+**TODO**: required curves
+
+## Content Reference {#content-reference}
+
+This section details how large content can be passed `by-reference` to a downloadable URL, rather than in-band.
```
Sakurann: we usually use `s for parameter names. 

```suggestion
This section details how large content can be passed by-reference to a downloadable URL, rather than in-band.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+SessionId MUST be globally unique. CredentialInstanceId MUST be unique for an Issuer. VerificationId and CredentialInstanceIds MUST be unique for a given SessionId.
+
+## Verification Endpoints {#verification-endpoints}
+
+The Verification endpoints are used to determine that a particular Wallet Client Instance authorized to have a Credential. This encompasses Identity Proofing, (e.g., document authentication, liveness detection, facial matching for government IDs), eligibility verification (e.g., fraud scoring, account standing checks for payment cards, access badges) as well as transferring Verification from another context (pre-auth code, device migration).
+
+### Engagement
+
+There are five different flows we are considering here:
```
Sakurann:
```suggestion
There are five different flows we are considering in this specification:
```

please bring this section higher. content-wise, this section is closer to the overview and relationship between the wallet and the issuer

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+1. **Wallet Initiated (Preknown Verification)**: Holder in a Wallet Client Instance, where the Issuer's required VerificationData is known in advance.
+1. **Wallet Initiated (Negotiated Verification)**: Holder in a Wallet Client Instance, where the Issuer's required Verification Data is unknown
+1. **Issuer Initiated (Authed Holder)**: Holder is in an Issuer Surface (e.g. Issuer App) where the holder is already authenticated. The Holder wishes to `push` one or more credentials to a Wallet Client Instance of their choosing.
+1. **Issuer Initiate (Unauthed Holder)**: Holder is in an Issuer Surface (e.g. Issuer Website) but are not currently (sufficiently) authenticated. The Holder wishes to initiate a Wallet Client Instance and begin Issuance of some Credentials.
+1. **Device Migration**: Holder in a new Wallet Client Instance, but has previously performed Verification in another Wallet Client Instance.
+
+All engagement is modeled as the Wallet receiving a `Credential Offer` payload from the Issuer.
```
Sakurann: Is "Credential Offer" here user in the same way as in VCI? worth mentioning earlier that that "Credential Offer" definition of VCI applies then

```suggestion
All engagement is modeled as the Wallet receiving a Credential Offer payload from the Issuer.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+This could be obtained:
+
+- Via a front-channel, such as DC API create, a deep-link or a QR code.
+- Via a back-channel call from the Issuer Server to the Wallet Server
+  - Though this is currently out of scope of this specification
```
Sakurann, line 346: why?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
dpostnikov: "protocol"?  It's really a sequence of API calls, no?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+ - Establishing a Wallet Signing Key (WSK) with the Issuer that is under control of the Holder of the Wallet Client Instance.
+ - Authenticate the Holder of the Wallet Client Instance to the Issuer, and determine what Credential Datasets it is authorized to have issued to it.
+
+This is done as follows: 
+
+1. The Wallet Client Instance generates a WSK, and the Wallet Server initiates the Verification on the Issuer Server.
+  1. The Wallet provides a stable unique Session Id associated with the WSK and the Issuer provides a stable Verification Id to reference this session.
+1. The Wallet Client Instance collects Verification Data and signs-then-encrypts it using the WSK and the Issuer Encryption Key.
+  1. Examples include digital credential presentations, auth on web, wallet collected documents and liveness/selfie checks.
```
dpostnikov, line 143:
```suggestion
  1. Examples include digital credential presentations, authentication on web, wallet-collected documents and liveness/selfie checks. 
```

dpostnikov: Why "on web"?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+ - Establishing a Wallet Signing Key (WSK) with the Issuer that is under control of the Holder of the Wallet Client Instance.
+ - Authenticate the Holder of the Wallet Client Instance to the Issuer, and determine what Credential Datasets it is authorized to have issued to it.
+
+This is done as follows: 
+
+1. The Wallet Client Instance generates a WSK, and the Wallet Server initiates the Verification on the Issuer Server.
+  1. The Wallet provides a stable unique Session Id associated with the WSK and the Issuer provides a stable Verification Id to reference this session.
+1. The Wallet Client Instance collects Verification Data and signs-then-encrypts it using the WSK and the Issuer Encryption Key.
+  1. Examples include digital credential presentations, auth on web, wallet collected documents and liveness/selfie checks. 
+1. The Wallet Server optionally provides additional Verification Data and calls the Issuer Server.
```
dpostnikov:
```suggestion
1. The Wallet Server calls The Issuer Server, passing Wallet Client-supplied Verification Data and, optionally, providing additional Wallet Server-supplied Verification Data.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
dpostnikov: Actually, it's probably both: "a set of APIs and a protocol".

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+The following Credential Status states are defined by this specification.
+
+| **Status**       | **Description**                                 |
+|------------------|-------------------------------------------------|
+| ACTIVE           | Credential is active and usable                 |
+| SUSPENDED_ISSUER | Suspended by the credential issuer              |
+| SUSPENDED_WALLET | Suspended by the wallet operator                |
+| UNLINKED         | Credential removed from device                  |
+| IN_PROCESS       | Provisioning or lifecycle operation in progress |
```
dpostnikov, line 703: What does it mean? Cannot be used?

INACTIVE, INACTIVE_IN_PROGRESS, or PENDING?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+| CANCELLED        | Credential permanently invalidated              |
+| UNKNOWN          | Status cannot be determined                     |
```
dpostnikov, line 705: Can it not find a credential? 404?

Or is it that the Wallet Client hasn't been live or synced for X period of time?

Or?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+| lastStatusChangeTime   | String (HTTP-date)        | **OPTIONAL**: Date and time when the status was last updated         |
+| lastStatusChangeReason | String                    | **OPTIONAL**: Machine-readable reason for the status change          |
+| extensions             | Object                    | **OPTIONAL**: Implementation-specific extension fields               |
+
+The following Credential Status states are defined by this specification.
+
+| **Status**       | **Description**                                 |
+|------------------|-------------------------------------------------|
+| ACTIVE           | Credential is active and usable                 |
+| SUSPENDED_ISSUER | Suspended by the credential issuer              |
```
dpostnikov, line 700: Is it different from REVOKED?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+| requestedDeviceCount       | integer >=1 (default 1)           | **OPTIONAL**: Number of devices the Wallet Server intends to provision for this verification session. The issuer uses this as input when determining `authorizedDeviceCount` in the approval response. If omitted, defaults to 1.                                |
+| locale                     | string                            | **OPTIONAL**: Local of the Wallet Client Instance to use for localization. When absent, the Issuer chooses the locale.                                                                                                                                          |
+| extensions                 | object                            | **OPTIONAL**: Implementation-specific extension fields. Reserved for proprietary data elements needed by specific deployments.                                                                                                                                   |
+
+#### Response Body Schema:
+
+The Response Body schema contains AllOf the fields in the VerificationStatus schema.
```
dpostnikov:
```suggestion
The Response Body schema contains all of the fields in the VerificationStatus schema.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+The API uses header-based semantic versioning. The Client MUST include the desired API version in the `Api-Version` request header using the semantic version (e.g. `1.0.0`).
+
+| **Header**             | **Direction** | **Type**                          | **Description**                                                                       |
+|------------------------|---------------|-----------------------------------|---------------------------------------------------------------------------------------|
+| Api-Version            | Request       | String (semantic version)         | Required. Semantic version requested by the client (e.g., 1.0.0)                      |
+| Api-Version            | Response      | String (semantic version)         | Required. The version used to process the request                                     |
+| Api-Supported-Versions | Response      | String (comma-separated versions) | Comma-separated list of versions the server supports                                  |
+| Sunset                 | Response      | String (HTTP-date)                | HTTP-date as defined in [@!RFC8594] after which the requested version will be removed |
```
dpostnikov, line 210: "Sunset" or "Api-Deprecated-Versions"?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+Server MUST reject requests with an unsupported `Api-Version` with HTTP 400 and a Problem Details response, indicating the supported versions in the header.
```
dpostnikov, line 212: Should we split the table above into two?

One for a client to request and one for the server to respond?

Otherwise, it currently reads as the client needs to supply supported and deprecated versions.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+```
+
+1. Holder, in the Wallet Client Instance, selects a Credential to Issue where the required VerificationData is unknown.
+2. The Wallet Client generates a WSK and initiates a session via the Wallet Server.
+3. The Wallet Server calls `verification/initiate` on the Issuer Server.
+4. The Issuer validates the verificationNonce and returns a status of `ADDITIONAL_INFO_REQUIRED` along with a VerificationDataRequest.
+5. The Wallet Server forwards the `ADDITIONAL_INFO_REQUIRED` status to the Wallet Client.
+6. The Wallet Client requests the VerificationData from the Holder.
+7. The Holder provides the VerificationData.
+8. The Wallet Client sends the Signed/Encrypted VerificationData to the Wallet Server.
```
dpostnikov:
```suggestion
8. The Wallet Client sends the Signed and Encrypted VerificationData to the Wallet Server.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+| **Event**                   | Origin        | **Description**                                 |
+|-----------------------------|---------------|-------------------------------------------------|
+| PROVISIONING_SUCCESS        | Wallet Server | Credential successfully provisioned to device   |
+| PROVISIONING_FAILURE        | Wallet Server | Credential provisioning failed                  |
+| LIFECYCLE_SUCCESS           | Wallet Server | Lifecycle action completed                      |
+| LIFECYCLE_FAILURE           | Wallet Server | Lifecycle action failed                         |
+| UPDATE_SUCCESS              | Wallet Server | Credential update applied to device             |
+| UPDATE_FAILURE              | Wallet Server | Credential update failed                        |
+| UPDATE_REMINDER             | Wallet Server | Pending update awaiting delivery                |
+| CREDENTIAL_EXPIRING         | Wallet Server | Credential approaching expiry                   |
```
dpostnikov, line 766: What about CREDENTIAL_EXPIRED?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+### Event Types
+
+The following event types are defined by this specification:
+
+| **Event**                   | Origin        | **Description**                                 |
+|-----------------------------|---------------|-------------------------------------------------|
+| PROVISIONING_SUCCESS        | Wallet Server | Credential successfully provisioned to device   |
+| PROVISIONING_FAILURE        | Wallet Server | Credential provisioning failed                  |
+| LIFECYCLE_SUCCESS           | Wallet Server | Lifecycle action completed                      |
```
dpostnikov, line 761: how different is it from the events below?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+| -------------------- | ------------------ | --------------------------------------------------------------------------------- |
+| sessionId            | string             | **REQUIRED**: Identifier for the overarching Session.                             |
+| credentialInstanceId | string             | **REQUIRED**: Credential Identifier associated with the event.                    |
+| eventType            | string             | **REQUIRED**: The type of event being notified.                                   |
+| triggeringRequestId  | string (UUID)      | **OPTIONAL**: The x-request-id of the original request that triggered this event. |
+| eventTimestamp       | string (date-time) | **REQUIRED**: The timestamp when the event was created.                           |
+| eventDetails         | object             | **TYPE DEPENDENT**: Implementation-specific details dependent on the eventType.   |
+| extensions           | object             | **OPTIONAL**: Implementation-specific extension fields.                           |
+
+### Event Types
```
dpostnikov, line 753: Image

Should we include a credential lifecycle section to ensure we are on the same page?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+# Issuer Metadata {#issuer-metadata}
+
+**TODO**: work out what needs to be configurable. E.g. issuer_server_url, configurations, supported_signing_algs, supported_encryption_algs, issuer_encryption_jwks, supported_extensions, and what comes from VCI spec s.a supported_proofs etc.
+
+# Security Considerations {#security-considerations}
+
+## Wallet Client Origin {#wallet-client}
+
+This specification makes use of a Wallet Signing Key to authenticate payloads originating from the Wallet Client. As part of Verification, the Wallet Signing Key signs-and-encrypts Verification Data, demonstrating to the Issuer that the Verification Data was collected in the presence of the Wallet Signing Key. How strongly that Data types the Wallet Signing Key depends on the specifics of the issuance process and is outside the scope of this specification.
+
```
dpostnikov, line 787:
```suggestion

## Wallet Client and Wallet Client Instance Trust {#wallet-client-trust}

This specification delegates the decision about whether a Wallet Client, or a specific instance of it, can be trusted, and to what extent, to the Wallet Server. For example, the Wallet Server can securely verify the Wallet Client's integrity using platform-level hardware checks before executing any lifecycle actions. The exact mechanism for determining Wallet Client trust is out of scope for this specification, but must be considered by implementers.

```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
dpostnikov: INACTIVE or PENDING... Maybe there is another spec we can refer to...

Can draw something like this in ASCII, of course.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+# Privacy Considerations {#privacy-considerations}
```
dpostnikov, line 797:
```suggestion

## Pre-authorization Code security measures  {#preauth-security}

It is important to note that anyone with a valid Pre-Authorized Code, without additional security measures, can receive a Credential from the Credential Issuer. Implementers MUST implement the mitigations most suitable to the use case.

# Privacy Considerations {#privacy-considerations}
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+## Data Minimization {#data-minimization}
+
+By design, the Wallet Server is always aware of what types of Credentials and the Issuers that a Wallet Holder is interacting with. This is necessary for the Wallet Server to provide value to the Issuer and the Holder. The specification is however designed with the principle of Data Minimization, where application layer encryption combined with sender authentication (using the Wallet Signing Key) is used to ensure confidentiality of the sensitive portions of the payload that are not required by the Wallet Server.
+
+# Implementation Considerations {#implementation-considerations}
+
+## Asynchronous Endpoints {#asynchronous-endpoints}
+
+The API is designed for the Verification and Credential endpoints to operate asynchronously. Wallet Servers MUST support responses being returned both as immediate responses, as well as asynchronously in the future. As notifications are unreliable, it's RECOMMENDED for Wallet Servers to use both notifications and polling-with-backoff to receive the results from these endpoints as quickly and reliably as possible.
```
dpostnikov:
```suggestion
The APIs are designed so that the Verification and Credential endpoints operate asynchronously. Wallet Servers MUST support returning responses both immediately and asynchronously in the future. As notifications are unreliable, it's RECOMMENDED that Wallet Servers use both notifications and polling-with-backoff to receive results from these endpoints as quickly and reliably as possible.
```

dpostnikov (commented):
Some preliminary thoughts, questions, and editorial tweaks.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+# Terminology {#terminology}
+
+This specification uses the terms "Credential", "Credential Configuration", "Credential Dataset", "Credential Format", "Holder", "Presentation",  "Verifiable Credential" and "Verifiable Presentation" as defined in OpenID4VCI. 
+
+This specification defines the following terms. In the case where a term has a definition that differs, the definition below is authoritative for this specification.
+
+Credential Issuer (or Issuer):
+:  An entity that issues Verifiable Credentials. In the context of this specification, the Credential Issuer implements the Issuer Endpoints.
+
+CredentialInstanceId:
```
GarethCOliver:
```suggestion
Credential Instance Identifier:
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Added to definition. VCI doesn't define whether any components are servers or application. Given it is important on the Wallet side, seems good to clarify on the issuer side for consistence. 

either way, don't have strong opinions and can change it editorially after the initial commit.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: 'Auth on web' is coming from VCI/IAE, it's referring to a flow where you redirect to an issuer website, perform to the auth and then redirect back. 

Issue #757 is tracking defining our set of verification types, which we should also likely unify with IAE.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+  1. The Issuer can perform the evaluation asynchronously to support evaluations that are not immediate (such as human reviews). 
+  1. The Issuer Server can optionally repeat this process to trigger the collection of additional Verification.
+1. After reaching a verdict the Issuer Server updates the verification status and notifies the Wallet Server of the change.
+
+On successful completion the Issuer Server now has a WSK that can be used to authenticate payloads as originating on a particular Wallet Client Instance, and have established which Credentials can be issued. 
+
+The Wallet retrieves the credentials as follows:
+
+1. The Wallet Server retrieves the Credential Instance Identifiers from the Wallet Server using the Session Id and Verification Id. 
+1. The Wallet Client generates proofs for Presentation Keys and signs-then-encrypts them using the WSK. The Wallet Client also creates an encryption key and signs it with the WSK.
```
GarethCOliver:
```suggestion
1. The Wallet Client generates proofs for Presentation Keys and signs-then-encrypts them using the WSK and the Issuer encryption key. The Wallet Client also creates an encryption key and signs it with the WSK. 
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: We really need some term to cover the first phase where the WSK is established and the Issuer determines the eligibility of Issuance of the Credentials. Due to the broad range of types of credentials that can be issued, it's naturally a pretty inclusive step. 

Other names that have come up in small group discussion: 
  - User Authentication
  - Enrollment
  - Proofing 

This is functionally the same as what occurs in IAE. I guess we could call it 'Interactive Authorization' but that is also an awful name IMO so. 

Leaving comment open to have naming debate in call today, be good to start with a name that people are somewhat in agreement with, even if we can change later.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1349 @@
...
+
+Issuance begins with a Verification phase. This phase serves two purposes: 
+
+ - Establishing a Wallet Signing Key (WSK) with the Issuer that is under control of the Holder of the Wallet Client Instance.
+ - Authenticate the Holder of the Wallet Client Instance to the Issuer, and determine what Credential Datasets it is authorized to have issued to it.
+
+This is done as follows: 
+
+1. The Wallet Client Instance generates a WSK, and the Wallet Server initiates the Verification on the Issuer Server.
+  1. The Wallet provides a stable unique Session Id associated with the WSK and the Issuer provides a stable Verification Id to reference this session.
```
GarethCOliver, line 141:
```suggestion
  1. The Wallet provides Credential Configuration(s) to identify the type of credentials they wish issued.
  1. The Wallet provides a stable unique Session Id associated with the WSK and the Issuer provides a stable Verification Id to reference this session.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: It could be. The design trade-offs are: 

- There is a (non-zero) stand-up cost for endpoints
- polymorphic endpoints can be harder to reason about
- separating endpoints can make analytics and monitoring easier when you have distinct expected behavior for each endpoint. 

I (personally) think having non-polymorphic endpoints is better. It is harder/impossible for IAE to split their endpoint (because it is attempting to build on oauth, which has a single endpoint as you are meant to only call it once with the same payload), so at least (IMO) IAE is making a necessary compromise.

As part of harmonizing with canonical VCI we could consider merging endpoints for consistence, but there are likely to be sufficient other differences that we may be better merging messages on level down (this is why the VerificationDataRequest/Response Schemas exist). 

Leaving topic open for further discussion

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: It's not bi-directional because the Issuer can't tell the Wallet server to cancel an in-progress verification, it just tell the Wallet about the result. 

The Cancel step is mostly there for clean-up purposes. 

In the Credential Lifecycle case, we want to be able to initiate the state changes from both sides and still continue the management of the credential which is a good reason for it to have the bi-directional. 

At the end of the day they are all kinda notifications, so we could just make notify bi-directional, but we currently don't have enough different types to justify that. Worth thinking about though.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: provision returns what data sets (as identifier)  the Wallet Client Instance can have issued. Get returns the particular VDC instances for a given credential data set instance.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Good question: we could put it in scope. I left it as out of scope as it is out of scope of canonical VCI, but I guess we do have a channel...

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: A state change is occurring, so currently unknown (basically a try-again-later). It should only be relevant for particular complex credential sets.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Created #772 to track this. In practice when I've drawn these in the past they often end up every state able to transition to every other state and so are a bit infeasible (and so a representative example of primary flows is more useful).

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: SUSPEND is temporary vs REVOKED is permanent. I think there is room to iterate on the exact set of states that make sense, so lets treat this as a starting set and use #765 to track what the final set should be.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Note from group: kinda has to be something generic. Some preference for 'Enrollment'. Will open an issue to track name choice and leave as is for now.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1354 @@
...
+
+# Overview  {#overview}
+
+This specification defines endpoints on both the Wallet Server and Credential
+Issuer Server to perform credential issuance and lifecycle management. This
+specification is designed for use with an architecture involving 3 components:
+
+- An Issuer Server operated by the Issuer
+- A Wallet Server operated by the Wallet Provider
+- A Wallet Client provided by the Wallet Provider and under the control of the
```
GarethCOliver, line 108: Update this to be less normative of what it is, but be precise about capabilities and the mental model.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
Sakurann: having both terms Credential Issuer and Credential Issuer Server defined confuses me a bit more ... just use credential issuer server for consistency, no?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1354 @@
...
+| title     | string   | Short human-readable summary of the error type                                               |
+| status    | integer  | HTTP status code                                                                             |
+| detail    | string   | Human-readable explanation specific to this occurrence                                       |
+| instance  | URI      | URI identifying this specific error occurrence (for support/debugging)                       |
+
+Implementations MAY extend the Problem Details object with additional fields (e.g., errorCode for machine-readable application codes, retryable flag, validationErrors array for field-level issues). See Section 3.1 of [@!RFC9457] for more information.
+
+### Error Types
+
+This specification defines the following error types:
```
Sakurann, line 232: why do error types need to be URNs..?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
Sakurann: to be consistent with already defined client_public_key, keyType should probably be key_type.
attestations is very generic, key_attestation is what is meant?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1354 @@
...
+
+Implementations MAY define their own key types. It is RECOMMENDED that collision-resistant values are used, such as r-DNS or urns.
+
+**TODO**: add post quantum
+
+**TODO**: required curves
+
+## Content Reference {#content-reference}
+
+This section details how large content can be passed by-reference to a downloadable URL, rather than in-band.
```
Sakurann:
```suggestion
This section details how large content can be passed by reference to a downloadable URL, rather than by value.
```

by reference vs by value are the terms usually used i think

Sakurann: also, content of what? fields below are fields of what? please define clearly

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
c2bo: This reads a bit dangerous

```suggestion
SessionId:
:  A globally unique identifier associated with a particular Issuance process between a Wallet Client Instance and an Issuer, persistent throughout the issuance process.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1354 @@
...
+
+#### Session Establishment
+
+The Wallet Server provides a long-lived SessionId to associate with the WalletSigningKey. The Issuer provides the VerificationId in the response, which is used for follow-up calls to the Verification Endpoints.
+
+`CredentialInstanceIdentifiers` are retrieved after verification through the Credential Provision endpoint.
+
+#### Replay Protection
+
+This endpoint is protected against replay attacks through a signed nonce mechanism. The wallet produces a digital signature over the sessionId using the walletSigningKey and includes this signature as the verificationNonce field in the request.
```
c2bo, line 366: I might be wrong, but given the current definitions, how are we guaranteeing freshness / preventing replay - I don't see an issuer provided input currently (since sessionID is potentially long-lasting)?

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
```diff
@@ -0,0 +1,1354 @@
...
+| Get Credential Metadata   | /credential/metadata     | Issuer             | Retrieves metadata for a specific credential configuration.                                              |
+| Get Credential Status     | /credential/status       | Both               | Bidirectional endpoint to query the current status of a credential instance.                             |
+| Credential Management     | /credential/manage       | Both               | Handles ongoing management and lifecycle operations for a Credential Instance.                           |
+| Event Notification        | /event/notify            | Both               | Bidirectional `poke` channel for action-based notifications.                                             |
+
+The following requirements apply to all endpoints.
+
+## Server Authentication {#server-authentication}
+
+All endpoints MUST use mutual TLS (mTLS) as the authentication method between the Wallet and Issuer Servers, as described by [@!RFC8705]. Both servers present X.509 client certificates during the TLS handshake. Certificate management is outside of scope of this specification.
```
c2bo, line 191: We probably want to bind between `SessionId` and `server authentication?

```suggestion
All endpoints MUST use mutual TLS (mTLS) as the authentication method between the Wallet and Issuer Servers, as described by [@!RFC8705]. Both servers present X.509 client certificates during the TLS handshake. Certificate management is outside of scope of this specification.

Issuers MUST associate a SessionId with the mTLS client identity that created it and reject requests for that SessionId from any other identity.
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver:
```suggestion
This specification defines a set of APIs and a protocol for credential issuance and lifecycle management. It
is credential format agnostic so Credentials can be of any format. Credentials can be of any format including,
but not limited to, IETF SD-JWT VC [@I-D.ietf-oauth-sd-jwt-vc], ISO mdoc
[@ISO.18013-5], and W3C VCDM [@VC_DATA_2.0].
```

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: closed with Issue #785

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closing this with Issue #786 to tracking improving introduction, overview and terminology

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closed with Issue #786

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closed with Issue #784

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closed with Issue #784

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closed with Issue #784

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closed with Issue #784

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: closed with Issue #754

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closed with #787 to see if we can improve the symmetry of the APIs to improve ergonomics.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: closed with #788 to track

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closed with #789 to do this

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Created #790 to track this.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: #791 tracking property names
#792 tracking how to create tables to describe JSON types in an understandable way. 

Disagree on the naming, a property inside an object of type 'Key' is implicitly a key_attestation. double naming it is a bad pattern (the coding analogy here is that I'll have a ' key: KeyData, then a key.attestation property, having a key.key_attestation property repeats itself).

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Tracking feature in #793

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closed with #794 to discuss this and to update the wording to make the properties clear regardless.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: closing for now: added it to discussion in #765

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closing with #768 to discuss and add more details on the meaning of notification and what exact list we want.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closing and added to #768

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closing with Issue #784  to track improving overview.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closing with Issue #784 to improve the Overview.

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Closing with #784

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
GarethCOliver: Tracking in terminology issue #786

s2s/openid-4-verifiable-credential-issuance-server-2-server.md
Sakurann: https://github.com/openid/OpenID4VCI/issues/786#issuecomment-5182871749

Sakurann (approved):

martijnharing (approved):

davidz25 (approved):

bhjelm (approved):

PR #759, opened 2026-06-18
add `redirect_uri` to Credential Offer

paulbastian:
Replaces #723 

- [x] discuss whether to use SHOULD or MAY to open redirect_uri
- [x] make editorial changes to fix grants text in Credential Offer Parameters in 1.0?
- [x] discuss whether expected_redirect_origins should be origins or paths


1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -379,23 +391,21 @@ This specification defines the following parameters for the JSON-encoded Credent
...
+    * `authorization_server`: OPTIONAL string that the Wallet can use to identify the Authorization Server to use with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. It MUST NOT be used otherwise. The value of this parameter MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
+* `redirect_uri`: OPTIONAL. String that is a URL using the `https` scheme. If present, the Wallet SHOULD offer the End-User the option to navigate the User Agent to this URL after it has finished processing the Credential Offer, e.g. to return to the Credential Issuer's website. The Wallet SHOULD offer this option regardless of whether the issuance succeeded or failed. The Wallet SHOULD NOT navigate to this URL without explicit End-User consent. Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST validate this value against the `expected_redirect_origin` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters) and MUST ignore the `redirect_uri` if the validation fails.
```
```suggestion
* `redirect_uri`: OPTIONAL. String that is a URL using the `https` scheme. If present, the Wallet SHOULD navigate the User Agent to this URL after it has finished processing the Credential Offer, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this regardless of whether the issuance succeeded or failed. The Wallet MAY ask the End-User for consent. Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST validate this value against the `expected_redirect_origin` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters) and MUST ignore the `redirect_uri` if the validation fails.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -379,23 +391,21 @@ This specification defines the following parameters for the JSON-encoded Credent
...
+      * `length`: OPTIONAL. Integer specifying the length of the Transaction Code. This helps the Wallet to render the input screen and improve the user experience.
+      * `description`: OPTIONAL. String containing guidance for the Holder of the Wallet on how to obtain the Transaction Code, e.g., describing over which communication channel it is delivered. The Wallet is RECOMMENDED to display this description next to the Transaction Code input screen to improve the user experience. The length of the string MUST NOT exceed 300 characters. The `description` does not support internationalization, however the Issuer MAY detect the Holder's language by previous communication or an HTTP Accept-Language header within an HTTP GET request for a Credential Offer URI.
+    * `authorization_server`: OPTIONAL string that the Wallet can use to identify the Authorization Server to use with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. It MUST NOT be used otherwise. The value of this parameter MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
```
paulbastian, line 403: for reviews: this part is only moved upwards

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -285,6 +289,8 @@ Note: Steps (3) and (4) may happen in the front channel, by redirecting the End-
 
 If the Credential Issuer requires more time to issue a Credential, the Credential Issuer may return a Transaction ID and a time interval in the Credential Response. The Wallet may send a Deferred Credential Request with the Transaction ID to obtain a Credential after the specified time interval has passed, as defined in (#deferred-credential-issuance).
 
+(7) If the Credential Offer contained a `redirect_uri` parameter (#credential-offer-parameters), the Wallet should offer the End-User to navigate the User Agent back to the Credential Issuer, for example, to return to the Credential Issuer's website. This redirect is OPTIONAL, is performed only upon explicit End-User interaction, and happens regardless of whether the issuance succeeded or failed.
```
GarethCOliver:
```suggestion
(7) If the Credential Offer contained a `redirect_uri` parameter (#credential-offer-parameters), the Wallet SHOULD navigate the User Agent to this URL after it has finished processing the Credential Offer, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this regardless of whether the issuance succeeded, failed, was cancelled or entered a pending state. The Issuer MUST NOT rely on the redirect occurring for successful issuance. 
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -379,23 +391,21 @@ This specification defines the following parameters for the JSON-encoded Credent
...
+    * `authorization_server`: OPTIONAL string that the Wallet can use to identify the Authorization Server to use with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. It MUST NOT be used otherwise. The value of this parameter MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
+* `redirect_uri`: OPTIONAL. String that is a URL using the `https` scheme. If present, the Wallet SHOULD navigate the User Agent to this URL after it has finished processing the Credential Offer, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this regardless of whether the issuance succeeded or failed. The Wallet MAY ask the End-User for consent. Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST validate this value against the `expected_redirect_origin` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters) and MUST ignore the `redirect_uri` if the validation fails.
```
GarethCOliver:
```suggestion
* `redirect_uri`: OPTIONAL. String that is a URL using the `https` scheme. If present, the Wallet SHOULD navigate the User Agent to this URL after it has finished processing the Credential Offer, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this regardless of whether the issuance succeeded, failed, was cancelled or entered a pending state. The Issuer MUST NOT rely on the redirect occurring for successful issuance. The Wallet MAY ask the End-User for consent before navigating. Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST validate this value against the `expected_redirect_origin` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters) and MUST ignore the `redirect_uri` if the validation fails.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1918,6 +1929,14 @@ The Wallet MUST NOT accept Credentials just because this mechanism was used. All
 
 The Credential Issuer MUST ensure the release of any privacy-sensitive data in Credential Offer is legal.
 
+### Redirect to the Credential Issuer
+
+Because the Credential Offer is not signed and its origin is not authenticated, an attacker could inject or manipulate the `redirect_uri` Credential Offer parameter (#credential-offer-parameters) to direct the End-User to an attacker-controlled website, for example, to conduct a phishing attack while leveraging the End-User's trust that they are returning to the Credential Issuer.
```
GarethCOliver: Can probably just say 


```suggestion

Because the parameter values in the Credential Offer are not trustworthy, an attacker could inject or manipulate the `redirect_uri` Credential Offer parameter (#credential-offer-parameters) to direct the End-User to an attacker-controlled website, for example, to conduct a phishing attack while leveraging the End-User's trust that they are returning to the Credential Issuer.
```

to link it directly to the above section.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1918,6 +1929,14 @@ The Wallet MUST NOT accept Credentials just because this mechanism was used. All
...
+
+To mitigate this, the Wallet relies on the `expected_redirect_origin` Credential Issuer metadata parameter (#credential-issuer-parameters). Unlike the Credential Offer, the Credential Issuer metadata is retrieved by the Wallet directly from the Credential Issuer over a TLS-protected connection (#credential-issuer-wellknown) and is therefore authenticated. Before offering the End-User the option to follow the `redirect_uri`, the Wallet MUST verify that the origin of the `redirect_uri` matches one of the origins listed in `expected_redirect_origin`, and MUST discard the `redirect_uri` otherwise, including when `expected_redirect_origin` is absent.
+
+The Wallet MUST NOT navigate to the `redirect_uri` automatically; the redirect MUST be performed only upon explicit End-User interaction. The Wallet SHOULD NOT include any information about the issued Credentials or the outcome of the issuance in the request to the `redirect_uri` beyond what the Credential Issuer encoded in the URL itself.
```
GarethCOliver: Remove?

1.1/openid-4-verifiable-credential-issuance-1_1.md
fkj: To clarify what is being consented to:
```suggestion
* `redirect_uri`: OPTIONAL. String that is a URL using the `https` scheme. If present, the Wallet SHOULD navigate the User Agent to this URL after it has finished processing the Credential Offer, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this regardless of whether the issuance succeeded or failed. The Wallet MAY ask the End-User for consent before navigating. Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST validate this value against the `expected_redirect_origin` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters) and MUST ignore the `redirect_uri` if the validation fails.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1918,6 +1929,14 @@ The Wallet MUST NOT accept Credentials just because this mechanism was used. All
 
 The Credential Issuer MUST ensure the release of any privacy-sensitive data in Credential Offer is legal.
 
+### Redirect to the Credential Issuer
+
+Because the Credential Offer is not signed and its origin is not authenticated, an attacker could inject or manipulate the `redirect_uri` Credential Offer parameter (#credential-offer-parameters) to direct the End-User to an attacker-controlled website, for example, to conduct a phishing attack while leveraging the End-User's trust that they are returning to the Credential Issuer.
+
+To mitigate this, the Wallet relies on the `expected_redirect_origin` Credential Issuer metadata parameter (#credential-issuer-parameters). Unlike the Credential Offer, the Credential Issuer metadata is retrieved by the Wallet directly from the Credential Issuer over a TLS-protected connection (#credential-issuer-wellknown) and is therefore authenticated. Before offering the End-User the option to follow the `redirect_uri`, the Wallet MUST verify that the origin of the `redirect_uri` matches one of the origins listed in `expected_redirect_origin`, and MUST discard the `redirect_uri` otherwise, including when `expected_redirect_origin` is absent.
```
fkj: The guidance here is just repeating normative requirements above. Is it necessary?

1.1/openid-4-verifiable-credential-issuance-1_1.md
paulbastian:
```suggestion
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
fkj: My concern is mainly that readers will get confused about whether this is really a new requirement and think they have misunderstood something. Also, for certification it is annoying and confusing to have duplicate requirements.
My suggestion would be to try to merge the first sentence explaining the reason into the main text.

1.0/openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -383,23 +383,20 @@ This specification defines the following parameters for the JSON-encoded Credent
...
+      * `description`: OPTIONAL. String containing guidance for the Holder of the Wallet on how to obtain the Transaction Code, e.g., describing over which communication channel it is delivered. The Wallet is RECOMMENDED to display this description next to the Transaction Code input screen to improve the user experience. The length of the string MUST NOT exceed 300 characters. The `description` does not support internationalization, however the Issuer MAY detect the Holder's language by previous communication or an HTTP Accept-Language header within an HTTP GET request for a Credential Offer URI.
+    * `authorization_server`: OPTIONAL string that the Wallet can use to identify the Authorization Server to use with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. It MUST NOT be used otherwise. The value of this parameter MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
 
 Additional Credential Offer parameters MAY be defined and used.
 The Wallet MUST ignore any unrecognized parameters.
```
paulbastian, line 399: these editorial changes move these two lines further down, which currently split the grants parameter from the grants values

GarethCOliver (approved):

mickrau (approved):

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1754,6 +1764,7 @@ This specification defines the following Credential Issuer Metadata parameters:
...
 * `notification_endpoint`: OPTIONAL. URL of the Credential Issuer's Notification Endpoint, as defined in (#notification-endpoint). This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Credential Issuer does not support the Notification Endpoint.
+* `expected_redirect_origin`: REQUIRED if the Credential Issuer uses the `redirect_uri` Credential Offer parameter (#credential-offer-parameters), and otherwise omitted. A non-empty array of strings, where each string is an origin, as defined in [@RFC6454], that the Credential Issuer may use in the `redirect_uri` of the Credential Offer. If a Credential Offer contains a `redirect_uri`, the Wallet MUST NOT utilize it unless the origin of the `redirect_uri` matches one of the values in this array. If this parameter is omitted, the Wallet MUST ignore any `redirect_uri` present in a Credential Offer from this Credential Issuer.
```
jogu: Should probably be plural as it's an array? and this is a normative reference I think, so `!`

```suggestion
* `expected_redirect_origins`: REQUIRED if the Credential Issuer uses the `redirect_uri` Credential Offer parameter (#credential-offer-parameters), and otherwise omitted. A non-empty array of strings, where each string is an origin, as defined in [@!RFC6454], that the Credential Issuer may use in the `redirect_uri` of the Credential Offer. If a Credential Offer contains a `redirect_uri`, the Wallet MUST NOT utilize it unless the origin of the `redirect_uri` matches one of the values in this array. If this parameter is omitted, the Wallet MUST ignore any `redirect_uri` present in a Credential Offer from this Credential Issuer.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1876,6 +1887,13 @@ The Wallet MUST NOT accept Credentials just because this mechanism was used. All
 
 The Credential Issuer MUST ensure the release of any privacy-sensitive data in Credential Offer is legal.
 
+### Redirect to the Credential Issuer
+
+Because the parameter values in the Credential Offer are not trustworthy, an attacker could inject or manipulate the `redirect_uri` Credential Offer parameter (#credential-offer-parameters) to direct the End-User to an attacker-controlled website, for example, to conduct a phishing attack while leveraging the End-User's trust that they are returning to the Credential Issuer.
+
+To mitigate this, the Wallet relies on the `expected_redirect_origin` Credential Issuer metadata parameter (#credential-issuer-parameters). Unlike the Credential Offer, the Credential Issuer metadata is retrieved by the Wallet directly from the Credential Issuer over a TLS-protected connection (#credential-issuer-wellknown) and is therefore authenticated. Before offering the End-User the option to follow the `redirect_uri`, the Wallet MUST verify that the origin of the `redirect_uri` matches one of the origins listed in `expected_redirect_origin`, and MUST discard the `redirect_uri` otherwise, including when `expected_redirect_origin` is absent.
```
jogu: "Before offering the End-User the option" is inconsistent with the text elsewhere saying "MAY ask the End-User for consent before navigating" - the check needs to happen even if the wallet decides to proceed without offering the user the option.

1.1/openid-4-verifiable-credential-issuance-1_1.md
jogu: If we start from what I believe is the intended attacker model (that the attacker can modify the credential offer)  then only authenticating the origin feels very lenient. e.g. if `https://www.microsoft.com` was listed here I'd be pretty certain there's some url somewhere on the website that I can sent the user to as useful part of some wider phishing attack or similar.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -285,6 +289,8 @@ Note: Steps (3) and (4) may happen in the front channel, by redirecting the End-
 
 If the Credential Issuer requires more time to issue a Credential, the Credential Issuer may return a Transaction ID and a time interval in the Credential Response. The Wallet may send a Deferred Credential Request with the Transaction ID to obtain a Credential after the specified time interval has passed, as defined in (#deferred-credential-issuance).
 
+(7) If the Credential Offer contained a `redirect_uri` parameter (#credential-offer-parameters), the Wallet SHOULD navigate the User Agent to this URL after it has finished processing the Credential Offer, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this regardless of whether the issuance succeeded, failed, was cancelled or entered a pending state. The Issuer MUST NOT rely on the redirect occurring for successful issuance.
```
jogu: I don't think some of this text should actually be here - this section doesn't contain any other normative keywords. It should probably be removed from here as the same text is already in the definition of the redirect_uri parameter.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -379,23 +391,21 @@ This specification defines the following parameters for the JSON-encoded Credent
...
+* `redirect_uri`: OPTIONAL. String that is a URL using the `https` scheme. If present, the Wallet SHOULD navigate the User Agent to this URL after it has finished processing the Credential Offer, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this regardless of whether the issuance succeeded, failed, was cancelled or entered a pending state. The Issuer MUST NOT rely on the redirect occurring for successful issuance. The Wallet MAY ask the End-User for consent before navigating. Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST validate this value against the `expected_redirect_origin` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters) and MUST ignore the `redirect_uri` if the validation fails.
```
jogu: I think this is a more general sentence? The issuer shouldn't rely it on it all, for neither successful nor failed issuance. Also, Americans.


```suggestion
* `redirect_uri`: OPTIONAL. String that is a URL using the `https` scheme. If present, the Wallet SHOULD navigate the User Agent to this URL after it has finished processing the Credential Offer, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this regardless of whether the issuance succeeded, failed, was canceled or entered a pending state. The Issuer MUST NOT rely on the redirect occurring. The Wallet MAY ask the End-User for consent before navigating. Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST validate this value against the `expected_redirect_origin` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters) and MUST ignore the `redirect_uri` if the validation fails.
```

jogu: "the" User Agent probably isn't correct.  In the cross device case it definitely won't be the user agent that was used earlier in the flow, and even in the same device case that isn't guaranteed.

Not sure exactly what it should say. It's probably sensible to make some recommendation about launching in the system browser vs using an in-app browser tab, but I don't know what is best.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -379,23 +391,21 @@ This specification defines the following parameters for the JSON-encoded Credent
...
```
jogu: > after it has finished processing the Credential Offer

> The Wallet SHOULD do this regardless of whether the issuance succeeded, failed, was cancelled or entered a pending state.

These statement feels quite ambiguous. 'pending state' we don't use anywhere else in the spec and I think could be clearer (I presume it might be referring to deferred issuance, but not sure if it's intended to encompass other things too). 'finished processing the Credential Offer' feels like it doesn't really match with 'issuance succeeded'.

1.1/examples/credential_offer_authz_code.txt
```diff
@@ -10,5 +10,6 @@ Content-Type: application/json
     "authorization_code": {
       "issuer_state": "eyJhbGciOiJSU0Et...FYUaBy"
     }
-  }
+  },
+  "redirect_uri": "https://credential-issuer.example.com/return?issuer_state=eyJhbGciOiJSU0Et...FYUaBy"
```
jogu: This makes me feel uneasy. If issuer_state is likely to be something important here, it feels like we should guarantee that the issuer_state passed to the redirect_uri is the one that was used in this session? (i.e. that it hasn't been switched out by an attacker.)

jogu (changes requested):

1.0/openid-4-verifiable-credential-issuance-1_0.md
```diff
@@ -1565,7 +1562,7 @@ One such use case is low assurance Credentials, such as coupons or tickets.
...
 
-If Cryptographic Key Binding is not required, the Credential Issuer omits the `cryptographic_binding_methods_supported` parameter in the `credential_configurations_supported` object in the Credential Issuer Metadata as defined in (#credential-issuer-parameters). As a consequence, the `proof_types_supported` parameter is also absent, indicating to the Wallet to omit the `proofs` parameter in the Credential Request.
+If Cryptographic Key Binding is not required, the Credential Issuer omits the `cryptographic_binding_methods_supported` parameter in the `credential_configurations_supported` object in the Credential Issuer Metadata as defined in (#credential-issuer-parameters). As a consequence, the `proof_types_supported` parameter is also absent, indicating to the Wallet to omit the `proofs` parameter in the Credential Request.
```
c2bo, line 1565: What the heck changed here? I can't see any changes?

paulbastian: I have no clue either, I don't remember changing something here

1.1/openid-4-verifiable-credential-issuance-1_1.md
paulbastian: I would change them to lower-case words, as "may" is also used in (6)

jogu: That would be okay but I think I'd try to reduce the text for this bullet a reasonable amount as I think the exact text is found in the redirect_uri definition and stating the detail only once is better in my opinion (and reduces the chance that the two drift into saying contradicting things).

1.0/openid-4-verifiable-credential-issuance-1_0.md
jogu: Somehow commit 95c9ce36ec93376b4325b967b87f3b17da6e0940 added a windows line ending (\r\n) on this line (the only one in the whole file) - this change just removes it.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```suggestion
(1b) The Issuer-initiated flow begins as the Credential Issuer generates a Credential Offer for certain Credential(s) that it communicates to the Wallet, for example, as a QR code or as a URI. The Credential Offer contains the Credential Issuer's URL, information about the Credential(s) being offered and an optional Redirect URL. This step is defined in (#credential-offer).
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
paulbastian:
```suggestion
(7) The Wallet may navigate the User Agent to the Credential Issuer's Redirect URL if it was contained in the Credential Offer.
```

1.1/examples/credential_offer_authz_code.txt
paulbastian: Does it help to remove the issuer_state query parameter?

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -379,23 +391,21 @@ This specification defines the following parameters for the JSON-encoded Credent
...
+    * `authorization_server`: OPTIONAL string that the Wallet can use to identify the Authorization Server to use with this grant type when `authorization_servers` parameter in the Credential Issuer metadata has multiple entries. It MUST NOT be used otherwise. The value of this parameter MUST match with one of the values in the `authorization_servers` array obtained from the Credential Issuer metadata.
+* `redirect_uri`: OPTIONAL. String that is a URL using the `https` scheme. If present, the Wallet SHOULD navigate the User Agent to this URL after it has finished processing the Credential Offer, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this regardless of whether the issuance succeeded, failed, was canceled or entered a pending state. The Issuer MUST NOT rely on the redirect occurring. The Wallet MAY ask the End-User for consent before navigating. Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST validate this value against the `expected_redirect_origin` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters) and MUST ignore the `redirect_uri` if the validation fails.
```
```suggestion
* `redirect_uri`: OPTIONAL. String that is a URL using the `https` scheme. If present, the Wallet SHOULD navigate the User Agent to this URL after it has finished processing the Credential Offer, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this regardless of whether the issuance succeeded, failed, was canceled or entered deferred state. The Issuer MUST NOT rely on the redirect occurring. The Wallet MAY ask the End-User for consent before navigating. Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST validate this value against the `expected_redirect_origin` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters) and MUST ignore the `redirect_uri` if the validation fails.
```

1.0/openid-4-verifiable-credential-issuance-1_0.md
paulbastian: so okay to keep this?

1.1/openid-4-verifiable-credential-issuance-1_1.md
fkj: Makes sense to me.

1.1/openid-4-verifiable-credential-issuance-1_1.md
fkj: See my comment further up - I think we should just delete this text completely.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1876,6 +1887,13 @@ The Wallet MUST NOT accept Credentials just because this mechanism was used. All
 
 The Credential Issuer MUST ensure the release of any privacy-sensitive data in Credential Offer is legal.
 
+### Redirect to the Credential Issuer
+
+Because the parameter values in the Credential Offer are not trustworthy, an attacker could inject or manipulate the `redirect_uri` Credential Offer parameter (#credential-offer-parameters) to direct the End-User to an attacker-controlled website, for example, to conduct a phishing attack while leveraging the End-User's trust that they are returning to the Credential Issuer.
+
+To mitigate this, the Wallet relies on the `expected_redirect_origin` Credential Issuer metadata parameter (#credential-issuer-parameters). Unlike the Credential Offer, the Credential Issuer metadata is retrieved by the Wallet directly from the Credential Issuer over a TLS-protected connection (#credential-issuer-wellknown) and is therefore authenticated. Before offering the End-User the option to follow the `redirect_uri`, the Wallet MUST verify that the origin of the `redirect_uri` matches one of the origins listed in `expected_redirect_origin`, and MUST discard the `redirect_uri` otherwise, including when `expected_redirect_origin` is absent.
+
+
```
paulbastian: Removing the security consideration as it didn't add significant value
```suggestion
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
fkj: Would you propose registering the domain + path instead? I think that would probably be fine by me. Does someone have an issue with that?

1.1/openid-4-verifiable-credential-issuance-1_1.md
fkj: I think: 1) rewrite to just say "a User Agent". 2). Recommend system browser (so we can go back to that if we were there before.

In the cross-device case, I expect most people would not use this `redirect_uri` and instead just keep going in the original user agent.

1.1/openid-4-verifiable-credential-issuance-1_1.md
paulbastian: applied

1.1/openid-4-verifiable-credential-issuance-1_1.md
fkj: I agree on credential offer.

For the various states, is there a reason to not just say "in all cases" instead of listing them out?

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1754,6 +1764,7 @@ This specification defines the following Credential Issuer Metadata parameters:
...
 * `notification_endpoint`: OPTIONAL. URL of the Credential Issuer's Notification Endpoint, as defined in (#notification-endpoint). This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Credential Issuer does not support the Notification Endpoint.
+* `expected_redirect_origins`: REQUIRED if the Credential Issuer uses the `redirect_uri` Credential Offer parameter (#credential-offer-parameters), and otherwise omitted. A non-empty array of strings, where each string is an origin, as defined in [@RFC6454], that the Credential Issuer may use in the `redirect_uri` of the Credential Offer. If a Credential Offer contains a `redirect_uri`, the Wallet MUST NOT utilize it unless the origin of the `redirect_uri` matches one of the values in this array. If this parameter is omitted, the Wallet MUST ignore any `redirect_uri` present in a Credential Offer from this Credential Issuer.
```
```suggestion
* `expected_redirect_origins`: REQUIRED if the Credential Issuer uses the `redirect_uri` Credential Offer parameter (#credential-offer-parameters), and otherwise omitted. A non-empty array of strings, where each string is an origin, as defined in [!@RFC6454], that the Credential Issuer may use in the `redirect_uri` of the Credential Offer. If a Credential Offer contains a `redirect_uri`, the Wallet MUST NOT utilize it unless the origin of the `redirect_uri` matches one of the values in this array. If this parameter is omitted, the Wallet MUST ignore any `redirect_uri` present in a Credential Offer from this Credential Issuer.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
paulbastian: the entire security consideration got deleted

1.1/openid-4-verifiable-credential-issuance-1_1.md
paulbastian: I modified the text, exchanged "pending state" with deferred to align the terminology

1.1/openid-4-verifiable-credential-issuance-1_1.md
paulbastian: I understand that "the user agent" doesn't work for cross device. We could say "the Wallet's user agent", which should match to same and cross-device flow? I believe the redirect_uri makes only sense to system browser

1.1/openid-4-verifiable-credential-issuance-1_1.md
paulbastian: I had a discussion with @GarethCOliver and @c2bo at IETF on this topic and we believe it's better to use origins than URLs, I added a security consideration instead

1.1/examples/credential_offer_authz_code.txt
paulbastian: I remove the query parameter

1.1/openid-4-verifiable-credential-issuance-1_1.md
paulbastian: I updated with the following text: "If present, the Wallet SHOULD navigate **the User Agent of the Wallet's device** to this URL after it has finished processing the Credential Offer"

paulbastian:
@jogu could you review again please?
@fkj @c2bo can you review?

fkj (approved):

dpostnikov:
DCP WG call today: Awaiting @jogu and @c2bo review

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -379,23 +391,21 @@ This specification defines the following parameters for the JSON-encoded Credent
...
+* `redirect_uri`: OPTIONAL. String that is a URL using the `https` scheme. If present, the Wallet SHOULD navigate the User Agent of the Wallet's device to this URL after it has finished processing the Credential Offer, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this regardless of whether the issuance succeeded, failed, was canceled or entered deferred state. The Issuer MUST NOT rely on the redirect occurring. The Wallet MAY ask the End-User for consent before navigating. Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST validate this value against the `expected_redirect_origins` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters) and MUST ignore the `redirect_uri` if the validation fails.
```
c2bo, line 404:
```suggestion
* `redirect_uri`: OPTIONAL. String that is a URL using the `https` scheme. If present, the Wallet SHOULD navigate the User Agent of the Wallet's device to this URL after the issuance flow initiated by the Credential Offer has concluded, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this regardless of whether the issuance succeeded, failed, was canceled or entered deferred state. The Issuer MUST NOT rely on the redirect occurring. The Wallet MAY ask the End-User for consent before navigating. Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST validate this value against the `expected_redirect_origins` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters) and MUST ignore the `redirect_uri` if the validation fails.
```

c2bo (approved):
Small nit/proposal and I am wondering if calling it `redirect_uri` which directly clashes with OAuth naming is helpful or harmful.  Maybe we should call it something like `issuer_return_uri` instead?

Looks good to me otherwise.

jogu:
> Small nit/proposal and I am wondering if calling it `redirect_uri` which directly clashes with OAuth naming is helpful or harmful. Maybe we should call it something like `issuer_return_uri` instead?

I agree with Christian - I think using a different name would really help avoid confusing, having two redirect_uris (the oauth one and this new one) in the same flow sounds like something best avoided.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1876,6 +1887,10 @@ The Wallet MUST NOT accept Credentials just because this mechanism was used. All
 
 The Credential Issuer MUST ensure the release of any privacy-sensitive data in Credential Offer is legal.
 
+## Redirect to the Credential Issuer {#redirect-security}
+
+The `expected_redirect_origins` Credential Issuer metadata parameter allows the Wallet to authenticate the origin of the `redirect_uri`. An attacker that is able to modify a Credential Offer, as described in (#credential-offer-security), can therefore manipulate the path, query, and fragment components under a listed origin. Credential Issuers therefore SHOULD only list origins whose content they fully control. In multi-tenant environments, this is achieved by hosting each tenant on its own subdomain, so that every tenant has a distinct origin, and by listing only that tenant's origin in `expected_redirect_origins`. Credential Issuers that cannot separate origins in this way and that are concerned about phishing attacks SHOULD NOT use the `redirect_uri` Credential Offer parameter.
```
jogu, line 1892: This feels like we're still leaving a foot gun there - is there really a clear use case that means (for example) we couldn't like URLs and require everything except the url query to match?

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1876,6 +1887,10 @@ The Wallet MUST NOT accept Credentials just because this mechanism was used. All
 
 The Credential Issuer MUST ensure the release of any privacy-sensitive data in Credential Offer is legal.
 
+## Redirect to the Credential Issuer {#redirect-security}
```
jogu, line 1890: I think we need an additional section pointing out that the content of the redirect_uri (or issuer_return_uri) are basically untrusted and that the issuer might want to try and correlate them with a session or otherwise validate they're applicable to the user viewing the page before doing too much with any returned values.

perhaps something like:

> Credential Issuers MUST NOT rely on data conveyed in the `redirect_uri`, to identify or authenticate the session in which the Credential Offer was issued, as an attacker that modified the Credential Offer controls those values.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1876,6 +1887,10 @@ The Wallet MUST NOT accept Credentials just because this mechanism was used. All
...
+
+The `expected_redirect_origins` Credential Issuer metadata parameter allows the Wallet to authenticate the origin of the `redirect_uri`. An attacker that is able to modify a Credential Offer, as described in (#credential-offer-security), can therefore manipulate the path, query, and fragment components under a listed origin. Credential Issuers therefore SHOULD only list origins whose content they fully control. In multi-tenant environments, this is achieved by hosting each tenant on its own subdomain, so that every tenant has a distinct origin, and by listing only that tenant's origin in `expected_redirect_origins`. Credential Issuers that cannot separate origins in this way and that are concerned about phishing attacks SHOULD NOT use the `redirect_uri` Credential Offer parameter.
+
```
jogu, line 1893: I think we should add to the text in privacy considerations to, not disimmilar to the text we already have about the oauth redirect url.

1.1/openid-4-verifiable-credential-issuance-1_1.md
jogu: Agree with Christian's suggestion, I think the 'User Agent of the Wallet's device' is still ambiguous though - I would guess some implementations might still interpret that as allowing an in-app browser tab, and I'm not sure if that's a good thing.

Here's Christian's suggestion incorporating Frederik's previous suggestion as well:

```suggestion
* `redirect_uri`: OPTIONAL. String that is a URL using the `https` scheme. If present, the Wallet SHOULD navigate a User Agent to this URL once the issuance flow initiated by the Credential Offer has concluded, e.g. to return to the Credential Issuer's website. The Wallet SHOULD do this in all cases, including where issuance succeeded, failed, was canceled or entered a deferred state. The Credential Issuer MUST NOT rely on the redirect occurring. The Wallet MAY ask the End-User for consent before navigating. Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST NOT use this value unless it validates against the `expected_redirect_origins` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters); see (#redirect-security).
```

brentzundel:
Discussed today, no opposition in group to the change suggested here: https://github.com/openid/OpenID4VCI/pull/759#pullrequestreview-5064215741
@jogu will finish his review and @GarethCOliver will review

c2bo:
I thought a bit about @jogu's comments during today's WG call. I believe the usual OAuth risk of `redirec_uri` don't really apply here since the redirect is not part of the security critical parts, but simply a redirect for information/UX - so at most a phishing angle still protected by TLS? I guess it would make sense nonetheless to warn to not use an origin that also hosts an open redirector etc.

Alternative solution: Rename the expected_redirect_origins parameter to something else and allow it to carry both, origins and URIs -> deployments can choose to use either an origin or full URLs. That gives deployments the freedom to adjust this to their deployment reality and should be reasonably easy to implement on the Wallet side: if the string contains path/query values, treat it as a strict URL match, otherwise compare prefix (origin).

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1754,6 +1764,7 @@ This specification defines the following Credential Issuer Metadata parameters:
...
 * `notification_endpoint`: OPTIONAL. URL of the Credential Issuer's Notification Endpoint, as defined in (#notification-endpoint). This URL MUST use the `https` scheme and MAY contain port, path, and query parameter components. If omitted, the Credential Issuer does not support the Notification Endpoint.
+* `expected_redirect_origins`: REQUIRED if the Credential Issuer uses the `redirect_uri` Credential Offer parameter (#credential-offer-parameters), and otherwise omitted. A non-empty array of strings, where each string is an origin, as defined in [@!RFC6454], that the Credential Issuer may use in the `redirect_uri` of the Credential Offer. If a Credential Offer contains a `redirect_uri`, the Wallet MUST NOT utilize it unless the origin of the `redirect_uri` matches one of the values in this array. If this parameter is omitted, the Wallet MUST ignore any `redirect_uri` present in a Credential Offer from this Credential Issuer.
```
jogu, line 1767: The 'MUST NOT utilize' here duplicates the MUST in the parameter definition:

> Because the Credential Offer is not authenticated (see (#credential-offer-security)), the Wallet MUST validate this value against the `expected_redirect_origins` Credential Issuer metadata parameter as defined in (#credential-issuer-parameters) and MUST ignore the `redirect_uri` if the validation fails.

I'd suggest removing the MUST here and referring to the normative text in the parameter definition.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -263,13 +263,17 @@ Please note that the diagram does not illustrate all the optional features defin
...
 Figure: Issuance using Authorization Code Flow 
 
 (1a) The Wallet-initiated flow begins as the End-User requests a Credential via the Wallet from the Credential Issuer. The End-User either selects a Credential from a pre-configured list of Credentials ready to be issued, or alternatively, the Wallet gives guidance to the End-User to select a Credential from a Credential Issuer based on the information it received in the presentation request from a Verifier.
 
-(1b) The Issuer-initiated flow begins as the Credential Issuer generates a Credential Offer for certain Credential(s) that it communicates to the Wallet, for example, as a QR code or as a URI. The Credential Offer contains the Credential Issuer's URL and the information about the Credential(s) being offered. This step is defined in (#credential-offer).
+(1b) The Issuer-initiated flow begins as the Credential Issuer generates a Credential Offer for certain Credential(s) that it communicates to the Wallet, for example, as a QR code or as a URI. The Credential Offer contains the Credential Issuer's URL, information about the Credential(s) being offered and an optional Redirect URL. This step is defined in (#credential-offer).
```
jogu, line 276: If we're going to capitalise 'Redirect URL' it should be added to the defined terms section.

1.1/openid-4-verifiable-credential-issuance-1_1.md
jogu, line 1767: Somewhere we should probably say the matching is done as per https://www.rfc-editor.org/info/rfc6454/#section-5

1.1/openid-4-verifiable-credential-issuance-1_1.md
jogu: Paul and I discussed a bit:

I understand his reason for using origins is that it can be troublesome to get paths in the .well-known updated when the expected redirect uris need to change.

I still think that allowing a redirect to anywhere opens the possibility for some phishing attacks - e.g. as an attack I send a credential offer issues a credential to the user, then redirects to somewhere on a massive website that ends up suggesting the user presents their new credential to an attacker controlled site to "prove your credential works" or "see how the flow works".

It feels to me like the same set of people that have trouble updating origins in .well-known has some overlap with the set of people that will have huge websites where that potential for user generated content is there.

It's definitely a trade off. I think I'm still feeling like the better option is to force the security on people, meaning they end up having to either get the .well-known updated, or spin up a subdomain (the latter being what we're basically recommending anyway). My fear is without that people will cut corners ("it's only a should, we can ignore it as it's difficult for us").

brentzundel:
Discussed in meeting, @paulbastian will address open comments

PR #775, opened 2026-07-10
Remove any terminating `/` from the Credential Issuer Identifier when forming the Credential Issuer Metadata URL

### LINKED ISSUE #744: Trailing-slash removal inadvertently dropped from Credential Issuer Metadata path construction (regression from #527)
[Section 12.2.2 ](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-1_0-wg-draft.html#section-12.2.2)(Credential Issuer Metadata Retrieval) currently says:

> Credential Issuers publishing metadata MUST make a JSON document available at the path formed by inserting the string `/.well-known/openid-credential-issuer` into the Credential Issuer Identifier between the host component and the path component, if any.

This is silent on what to do when the Issuer Identifier ends in `/`. RFC 8414 §3.1, which #527 aimed to align with, explicitly requires:

> If the issuer identifier value contains a path component, any terminating "/" MUST be removed before inserting "/.well-known/" and the well-known URI suffix between the host component and the path component.

## History

Prior to PR #527, the spec did include this rule ("any terminating `/` MUST be removed before appending..."). #527 rewrote the sentence to fix the *insertion position* (concatenate-at-end → insert between host and path, matching RFC 8414/8615), but the terminating-`/` clause was dropped in the rephrase. The PR discussion only covered the worked example — the removal was not raised or I believe (as the PR author) was not intended.

## Impact

For an identifier like `https://issuer.example.com/tenant/`, the current wording yields `https://issuer.example.com/.well-known/openid-credential-issuer/tenant/`, whereas RFC 8414's rule yields `.../tenant`. Different implementations may diverge.

## Proposed fix

Restore the terminating-`/` removal rule to §12.2.2 (and the 1.1 draft) to match RFC 8414 §3.1, e.g. append:

> If the Credential Issuer Identifier contains a path component with a terminating `/`, the terminating `/` MUST be removed before inserting `/.well-known/openid-credential-issuer`.

Unfortunately this would technically be a breaking change I think, so requires some thought.

This got raised to the certification team under https://gitlab.com/openid/conformance-suite/-/work_items/1823

paulbastian:
Closes #744 

My observations while building this PR:
- the actual cleanest approach to this would be to forbid trailing slashes for the Credential Issuer Identifier in the first place
  - however, this is more of a breaking change and not aligned with RFC8414, which likely must have overseen this more elegant solution, as it prevents this edge case at the AS/RS side, instead of at client parsing
  - we still could add this as a recommendation, if there is appetite
- RFC8414 is also missing an example for the trailing slash, which in my opinion is important to highlight, so I added another example into the text that actually showcases the rule that is re-added in this PR

### DIFF
--- a/1.1/openid-4-verifiable-credential-issuance-1_1.md
+++ b/1.1/openid-4-verifiable-credential-issuance-1_1.md
@@ -1688,7 +1688,7 @@ The Credential Issuer Metadata contains information on the Credential Issuer's t
 
 ### Credential Issuer Identifier {#credential-issuer-identifier}
 
-A Credential Issuer is identified by a case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components. The Credential Issuer Identifier SHOULD NOT contain a terminating `/`.
+A Credential Issuer is identified by a case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components.
 
 ### Credential Issuer Metadata Retrieval {#credential-issuer-wellknown}
 
@@ -3722,6 +3722,5 @@ The technology described in this specification was made available from contribut
    * add iana registration for an openid foundation urn
    * add optional metadata to the credential response
    * use OAuth 2.0 for First-Party Applications as basis for Interactive Authorization
-   * add recommendation that Credential Issuer Identifier should not terminate with `/`
    * add back removal of any terminating `/` from the Credential Issuer Identifier when forming the Credential Issuer Metadata URL
    * update IA HTTP response codes for consistency with First-Party Application draft-4
--- a/1.1/openid-4-verifiable-credential-issuance-1_1.md
+++ b/1.1/openid-4-verifiable-credential-issuance-1_1.md
@@ -1688,7 +1688,7 @@ The Credential Issuer Metadata contains information on the Credential Issuer's t
 
 ### Credential Issuer Identifier {#credential-issuer-identifier}
 
-A Credential Issuer is identified by a case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components.
+A Credential Issuer is identified by a case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components. The Credential Issuer Identifier SHOULD not contain a terminating `/`.
 
 ### Credential Issuer Metadata Retrieval {#credential-issuer-wellknown}
 
@@ -3722,5 +3722,6 @@ The technology described in this specification was made available from contribut
    * add iana registration for an openid foundation urn
    * add optional metadata to the credential response
    * use OAuth 2.0 for First-Party Applications as basis for Interactive Authorization
+   * add recommendation that Credential Issuer Identifier should not terminate with `/`
    * add back removal of any terminating `/` from the Credential Issuer Identifier when forming the Credential Issuer Metadata URL
    * update IA HTTP response codes for consistency with First-Party Application draft-4
--- a/1.1/openid-4-verifiable-credential-issuance-1_1.md
+++ b/1.1/openid-4-verifiable-credential-issuance-1_1.md
@@ -1688,15 +1688,15 @@ The Credential Issuer Metadata contains information on the Credential Issuer's t
 
 ### Credential Issuer Identifier {#credential-issuer-identifier}
 
-A Credential Issuer is identified by a case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components. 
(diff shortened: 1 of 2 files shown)


Sakurann (approved):

Sakurann:
WG discussion:
- seems reasonable to add a text that recommends (SHOULD) to the issuers not to use this trailing /

c2bo (dismissed):
I believe we should apply the same rules the `sub` of the signed metadata? 

That is currently described as 

> sub: REQUIRED. String matching the Credential Issuer Identifier

With this PR we are relaxing the matching rules for metadata retrieval, but are not doing the same for `sub` -> will likely cause mismatches in the validation.

paulbastian:
> I believe must apply the same rules the `sub` of the signed metadata?
> 
> That is currently described as
> 
> > sub: REQUIRED. String matching the Credential Issuer Identifier
> 
> With this PR we are relaxing the matching rules for metadata retrieval, but are not doing the same for `sub` -> will likely cause mismatches in the validation.

This PR does not *relax* matching rules for, but fixes them by adding additional constraint.
As I see the `sub` value of signed metadata, it is a simple-string comparison with the Credential Issuer Identifier, which is not changed in this PR, so I don't see any issues.

paulbastian:
As discussed in the WG call, I added a recommendation that Credential Issuer Identifier SHOULD NOT terminate with `/`, I only added this for 1.1, as I wasn't sure this is justified for errata.

c2bo:
> This PR does not _relax_ matching rules for, but fixes them by adding additional constraint.
> As I see the `sub` value of signed metadata, it is a simple-string comparison with the Credential Issuer Identifier, which is not changed in this PR, so I don't see any issues.

The metadata is fetched for a Credential Issuer Identifier. With this PR, a provided Credential Issuer Identifier `https://issuer.example.com/` would be correctly retrieved from `https://tenant.issuer.example.com/.well-known/openid-credential-issuer`, but the `sub` in that signed metadata would likely be `https://issuer.example.com`, failing the matching rules of sub = Credential Issuer Identifier:

> sub: REQUIRED. String matching the Credential Issuer Identifier

We are basically just moving the problem 1 layer further down for signed metadata - the fix would be to mandate that Credential Issuer Identifier not having a trailing slash I guess, but I guess that would be a breaking change?

paulbastian:
> I believe must apply the same rules the `sub` of the signed metadata? 
> 
> That is currently described as 
> 
> > sub: REQUIRED. String matching the Credential Issuer Identifier
> 
> With this PR we are relaxing the matching rules for metadata retrieval, but are not doing the same for `sub` -> will likely cause mismatches in the validation.

If the Credential Issuer Identifier is https://issuer.example.com/ , why wouldn't  the sub value as well be https://issuer.example.com/ ?

I'm also in favor of fixing the Credential Issuer Identifier in the first place, but I don't understand the problem for the proposed text.

c2bo:
> If the Credential Issuer Identifier is https://issuer.example.com/ , why wouldn't the sub value as well be https://issuer.example.com/ ?

We get Credential Issuer Identifier from whatever source (e.,g., a  credential catalogue) and use it to resolve the issuer metadata to start an issuance process. With this PR we basically fix the metadata resolution to always work regardless of trailing slash, but for signed metadata it's imho somewhat likely that the sub == identifier match would still fail (for a trailing slash) --> I'd slightly modify that to not require strict string matching, but also allow removal of trailing slash for the matching.

Not a super critical thing imho, but this whole PR is basically trying to make the metadata fetching logic a bit more robust and adding the same for the sub = identifier matching logic is probably a good idea so we don't fail on the next step after fetching.

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -1688,15 +1688,15 @@ The Credential Issuer Metadata contains information on the Credential Issuer's t
 
 ### Credential Issuer Identifier {#credential-issuer-identifier}
 
-A Credential Issuer is identified by a case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components. 
+A Credential Issuer is identified by a case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components. The Credential Issuer Identifier SHOULD NOT contain a terminating `/`.
```
```suggestion
A Credential Issuer is identified by a case sensitive URL using the `https` scheme that contains scheme, host and, optionally, port number and path components, but no query or fragment components.
```

1.1/openid-4-verifiable-credential-issuance-1_1.md
```diff
@@ -3722,4 +3722,6 @@ The technology described in this specification was made available from contribut
    * add iana registration for an openid foundation urn
    * add optional metadata to the credential response
    * use OAuth 2.0 for First-Party Applications as basis for Interactive Authorization
+   * add recommendation that Credential Issuer Identifier should not terminate with `/`
```
```suggestion
```

fkj (approved):
I think this text is fine as is, and I don't think the comment about potentially also changing recommendations for `sub` should block this PR.

c2bo (approved):

PR #797, opened 2026-08-18
Add requirement to use state to bind concurrent sessions.

### LINKED ISSUE #702: Stuttgart 8: how does wallet determine which session if there are multiple?
> When a wallet is waiting for the user to "return" from a redirect_to_web, it expects to receive either a code or an auth_session in the AS's (redirect) response. If the wallet has multiple interactive authorization sessions in parallel, how can the wallet determine to which of these sessions an arriving response (at the wallet's redirect endpoint) belongs?

fkj:
This is intended to fix #702.


brentzundel (approved):

c2bo (approved):

GarethCOliver (approved):

paulbastian (approved):

brentzundel:
Multiple positive approvals, no requests for changes, sufficient review time. Merging.

PR #803, opened 2026-08-31
Add requirement to bind auth_session if DPoP is used.

### LINKED ISSUE #718: binding auth_session to dpop
https://www.ietf.org/archive/id/draft-ietf-oauth-first-party-apps-03.html#name-auth-session-dpop-binding recommends binding the auth_session to dpop - we should probably consider something similar, including potentially mandating it in HAIP.

fkj:
Resolves #718 according to the WG consensus deciding to mandate binding the `auth_session` if DPoP is used.
One potential review question: do we want to also require binding any "custom" session binding mechanisms for interaction types that don't use `auth_session`. And is it possible to write this in a generic way?


brentzundel (approved):

brentzundel:
Discussed today. @c2bo and @GarethCOliver  will review

GarethCOliver (approved):

